PR #65, opened 2020-04-17
[master] Update BitBucket links

chapulina:
Signed-off-by: Louise Poubel <louise@openrobotics.org>


iche033:
btw I think we can just remove `bitbucket-pipelines.yml`. We can do that in a separate PR if necessary

chapulina:
> I think we can just remove bitbucket-pipelines.yml

I think `release-tools`'s `auto` jobs may still be using it to figure out the Ubuntu version. CC @j-rivero

Issue #19, opened 2018-01-02
Optix: Support mouse picking


osrf-migration:
**Original report ([archived issue](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-rendering/issues/19)) by Ian Chen (Bitbucket: [Ian Chen](https://bitbucket.org/%7Beaa6fca5-6deb-43f6-907f-971c144735dd%7D/), GitHub: [iche033](https://github.com/iche033)).**

----------------------------------------
Add a function that lets user pass in x, y coordinates in screen and returns the name / ID of selected entity in the scene

Ogre implementation:

- port gazebo's [selection buffer](https://bitbucket.org/osrf/gazebo/src/eddba39b82cd7a8d9d2e3cff7c6a43174bdd972b/gazebo/rendering/selection_buffer/?at=default) and offer a function similar to gazebo's [Scene::VisualAt](https://bitbucket.org/osrf/gazebo/src/eddba39b82cd7a8d9d2e3cff7c6a43174bdd972b/gazebo/rendering/Scene.cc?at=default&fileviewer=file-view-default#Scene.cc-988)

Optix:

- TBD

PR #103, opened 2020-06-20
Added LidarVisual implementation for Ogre1, empty classes for Ogre2

mihirk284:
Please see issue [ignitionrobotics/ign-rendering#84](https://github.com/ignitionrobotics/ign-rendering/issues/84).

I have added lidar visualisation for Ogre1 and an example for the same.

The following image shows an example of the LidarVisual with 3 vertical sets of rays.
image

@ahcorde  Please do check the PR and let me know if there are any changes. ( This PR is different from the previous one so that it would be easier to review the code as it is separated cleanly per commit.)


examples/lidar_visual/Main.cc
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

src/LidarVisual.cc
```diff
@@ -0,0 +1,32 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
ahcorde, line 2:
```suggestion
 * Copyright (C) 2020 Open Source Robotics Foundation
```

src/LidarVisual_TEST.cc
```diff
@@ -0,0 +1,131 @@
+/* * Copyright (C) 2019 Open Source Robotics Foundation
```
ahcorde, line 1:
```suggestion
/* * Copyright (C) 2020 Open Source Robotics Foundation
```

include/ignition/rendering/LidarVisual.hh
```diff
@@ -0,0 +1,157 @@
...
+      public: virtual void SetOffset(const ignition::math::Pose3d _offset) = 0;
+
+      /// \brief Get offset of visual
+      /// \return The offset of the lidar rays from visual origin
+      public: virtual ignition::math::Pose3d  Offset() = 0;
+
+      /// \brief Get number of points in laser data
+      /// \return The number of points in the laser data
+      public: virtual unsigned int GetPointCount() = 0;
+
```
ahcorde, line 151: remove this line

include/ignition/rendering/base/BaseLidarVisual.hh
```diff
@@ -0,0 +1,383 @@
...
+
+      /// \brief Minimum Range
+      protected: double minRange = 0;
+
+      /// \brief Maximum Range
+      protected: double maxRange = 0;
+
+      /// \brief Offset of visual
+      protected: ignition::math::Pose3d offset = ignition::math::Pose3d::Zero;
+
```
ahcorde, line 160: remove this line

ogre2/src/Ogre2Scene.cc
```diff
@@ -26,6 +26,7 @@
 #include "ignition/rendering/ogre2/Ogre2Grid.hh"
```
ahcorde, line 26: You are adding the implementation for Ogre2 too, can you edit the PR message ?

include/ignition/rendering/base/BaseLidarVisual.hh
```diff
@@ -0,0 +1,383 @@
...
+      public: virtual void SetLidarMessage(std::vector<double> &msg);
+
+      // Documentation inherited
+      public: virtual void Update();
+
+      // Documentation inherited
+      public: virtual void Init();
+
+      // Documentation inherited
+      public: virtual void SetMinVerticalAngle(const double );
```
ahcorde, line 60: please add the argument names, with undescore `_` -> _minVerticalAngle. Same with the methods bellow

include/ignition/rendering/base/BaseLidarVisual.hh
```diff
@@ -0,0 +1,383 @@
...
+      public: virtual void PreRender() override;
+
+      // Documentation inherited
+      public: virtual void Destroy() override;
+
+      // Documentation inherited
+      public: virtual void ClearPoints() override;
+
+      // Documentation inherited
+      public: virtual void SetLidarMessage(std::vector<double> &msg);
```
ahcorde, line 51:
```suggestion
      public: virtual void SetLidarMessage(std::vector<double> & _msg);
```

ogre2/src/Ogre2Scene.cc
mihirk284: done :+1:

ogre2/src/Ogre2LidarVisual.cc
```diff
@@ -0,0 +1,76 @@
...
+#include <ignition/common/Console.hh>
+
+#include "ignition/rendering/ogre2/Ogre2DynamicRenderable.hh"
+#include "ignition/rendering/ogre2/Ogre2LidarVisual.hh"
+#include "ignition/rendering/ogre2/Ogre2Material.hh"
+#include "ignition/rendering/ogre2/Ogre2Mesh.hh"
+#include "ignition/rendering/ogre2/Ogre2Scene.hh"
+#include "ignition/rendering/ogre2/Ogre2Visual.hh"
+
+// class ignition::rendering::Ogre2LidarVisualPrivate
```
ahcorde, line 27: remove this code if you are not using it or uncomment them otherwise

ogre/src/media/materials/scripts/lidar_visual.material
```diff
@@ -0,0 +1,102 @@
...
+    {
+      ambient 0 0 1
+      diffuse 0 0 1
+      specular 0.1 0.1 0.1 1 1
+    }
+  }
+}
+
+
+
```
ahcorde, line 82: remove this line

ahcorde:
It will be great to add a test similar to these ones https://github.com/ignitionrobotics/ign-rendering/tree/master/test/integration

Sarath18:
Hey @mihirk284, I plan to integrate LidarVisual to ign-rviz and you might have checked out the ticket [#19](https://github.com/Sarath18/ign-rviz/issues/19) for the same. 

As far as I know, Rviz visualizes points to visualize LaserScan data but on the other gazebo uses triangles to visualize the data (your current implementation). 

Is it possible for you to add a method in the API to switch visualization type between triangles and points?

mihirk284:
> Hey @mihirk284, I plan to integrate LidarVisual to ign-rviz and you might have checked out the ticket [#19](https://github.com/Sarath18/ign-rviz/issues/19) for the same.
> 
> As far as I know, Rviz visualizes points to visualize LaserScan data but on the other gazebo uses triangles to visualize the data (your current implementation).
> 
> Is it possible for you to add a method in the API to switch visualization type between triangles and points?

Yeah... It would be possible to visualise points and to switch between them and triangles. In case you also need to vary the colours of points in a single message ( eg. like intensity, reflectivity, x, y, z, axis gradient) I'm not sure what would be the best way of implementing that here...

@iche033 @ahcorde  any suggestions for the same?

iche033:
about setting color of points, you can override the `SetMaterial` function in LidarVisual and just iterate through the dynamic lines and call `line->setMaterial(_material->Name())` where `_material` is the arg to the `SetMaterial` function`

include/ignition/rendering/LidarVisual.hh
```diff
@@ -0,0 +1,165 @@
...
+      /// \brief Clear the points of the lidar visual
+      public: virtual void ClearPoints() = 0;
+
+      /// \brief Update the Visual
+      public: virtual void Update() = 0;
+
+      /// \brief Callback when the senspr data is recieved
+      /// \param[in] _msg Vector of points representing distance of the ray
+      public: virtual void SetLidarMessage(std::vector<double> &_msg) = 0;
```
iche033, line 58: can we make this API name more generic like `SetPoints`?

include/ignition/rendering/LidarVisual.hh
```diff
@@ -0,0 +1,165 @@
...
+      public: virtual void SetOffset(
+                  const ignition::math::Pose3d _offset) = 0;
+
+      /// \brief Get offset of visual
+      /// \return The offset of the lidar rays from visual origin
+      public: virtual ignition::math::Pose3d  Offset() const = 0;
+
+      /// \brief Get number of points in laser data
+      /// \return The number of points in the laser data
+      public: virtual unsigned int GetPointCount() const = 0;
```
iche033, line 159: we follow google coding style guidelines which don't use `Get` prefix in getter functions. So `GetPointCount()` -> `PointCount()`

ogre/src/media/materials/scripts/lidar_visual.material
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+material Lidar/LightBlueLaser
```
iche033, line 18: The material script is specific to ogre 1.x which won't be used by other rendering engine implementations. So you may want to create materials manually by hand, [see examples here](https://github.com/ignitionrobotics/ign-rendering/blob/master/src/base/BaseScene.cc#L1150). You could create them in BaseLidarVisual so that when you work on ogre2 implementation, these materials will be available.

mihirk284:
Added example of a GPU rays sensor with data collected lidar data being visualised
image

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,332 @@
...
+  public: std::vector<std::shared_ptr<OgreDynamicLines>> rayStrips;
+
+  /// \brief Dead Zone Geometry DynamicLines Object to display
+  public: std::vector<std::shared_ptr<OgreDynamicLines>> deadZoneRayFans;
+
+  /// \brief Lidar Ray DynamicLines Object to display
+  public: std::vector<std::shared_ptr<OgreDynamicLines>> rayLines;
+
+  /// \brief The current lidar data message
+  public: std::vector<double> laserMsg;
```
iche033, line 41: can you rename this to something like `lidarPoints` since we are not using a message to store the data in ign-rendering

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,332 @@
...
+  double horizontalAngle = this->minHorizontalAngle;
+  double verticalAngle = this->minVerticalAngle;
+
+  if (this->horizontalCount > 1)
+  {
+    this->horizontalAngleStep =
+        (this->maxHorizontalAngle - this->minHorizontalAngle) /
+              (this->horizontalCount - 1);
+  }
+  if (this->verticalCount > 1)
```
iche033, line 152: can the vertical and horizontal checks be moved to `Init` or `Create`? We should check this early so we don't end up spamming the console on every update

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,332 @@
...
+      }
+
+      if (this->minHorizontalAngle - horizontalAngle > 1.0e-08 )
+      {
+        ignerr <<
+            "Horizontal angle less than minimum limits."
+          << std::endl;
+      }
+
+      if (horizontalAngle - this->maxHorizontalAngle > 1.0e-08)
```
iche033, line 248: would we ever get into a situation that we fail these `if` checks? 

Like mentioned before, we should make sure all the user specified params are validated early on so we don't run into these errors on update.

include/ignition/rendering/LidarVisual.hh
```diff
@@ -0,0 +1,151 @@
...
+      public: virtual void Update() = 0;
+
+      /// \brief Callback when the sensor data is recieved
+      /// \param[in] _msg Vector of points representing distance of the ray
+      public: virtual void SetPoints(std::vector<double> &_msg) = 0;
+
+      /// \brief Set minimum vertical angle
+      /// \param[in] _minVerticalAngle Minimum vertical angle
+      public: virtual void SetMinVerticalAngle(
+                  const double _minVerticalAngle) = 0;
```
iche033, line 63: we started removing `const` from primitive types a while back. Could you change the `const double` types in this class to just `double`? Same for the derived classes

include/ignition/rendering/LidarVisual.hh
```diff
@@ -0,0 +1,151 @@
...
+
+      /// \brief Clear the points of the lidar visual
+      public: virtual void ClearPoints() = 0;
+
+      /// \brief Update the Visual
+      public: virtual void Update() = 0;
+
+      /// \brief Callback when the sensor data is recieved
+      /// \param[in] _msg Vector of points representing distance of the ray
+      public: virtual void SetPoints(std::vector<double> &_msg) = 0;
```
iche033, line 58: can you pass by const ref? so we can guarantee to the caller of this API that we're not going to change the contents of the vector.

test/integration/lidar_visual.cc
```diff
@@ -0,0 +1,572 @@
...
+
+  // Test vertical measurements
+  public: void LaserVertical(const std::string &_renderEngine);
+};
+
+/////////////////////////////////////////////////
+/// \brief Test LidarVisual configuraions
+void LidarVisualTest::Configure(const std::string &_renderEngine)
+{
+  if (_renderEngine == "optix")
```
iche033, line 70: also add a check for `ogre2` for all the tests since users can still run tests using `ogre2` by doing:

```
 RENDER_ENGINE_VALUES=ogre2 make test
```

test/integration/lidar_visual.cc
```diff
@@ -0,0 +1,572 @@
...
+  double hMinAngle = -IGN_PI/2.0;
+  double hMaxAngle = IGN_PI/2.0;
+  double vMinAngle = -IGN_PI/4.0;
+  double vMaxAngle = IGN_PI/4.0;
+  double minRange = 0.1;
+  double maxRange = 5.0;
+  unsigned int hRayCount = 640;
+  unsigned int vRayCount = 4;
+
+  common::Time waitTime = common::Time(WAIT_TIME);
```
iche033, line 408: You probably got this from `gpu_rays` test but I think it is not used. We can remove this

examples/lidar_visual/GlutWindow.cc
```diff
@@ -0,0 +1,360 @@
...
+
+std::vector<ir::CameraPtr> g_cameras;
+ir::CameraPtr g_camera;
+ir::CameraPtr g_currCamera;
+unsigned int g_cameraIndex = 0;
+ir::ImagePtr g_image;
+bool g_updateAll = false;
+std::chrono::steady_clock::duration g_time{0};
+std::chrono::steady_clock::time_point g_startTime;
+std::chrono::steady_clock::time_point prevUpdateTime;
```
iche033, line 63: some of the variables above are not used. We should remove the unused ones

ogre/src/OgreLidarVisual.cc
mihirk284: I put them here as the parameters can be changed by the user after the lidarvisual is created, hence allowing the user to set the parameters irrespective of the order... for example, the lidarPoints may be set by the user before setting the Vertical or horizontal RayCount and it will only be evaluated after all the parameters have been fully set, i.e. checking performed within the update function. It would also allow the user to dynamically change the parameters if required.

As of now, the update function only checks for 1) if a new message has been received, and 2) if the number of points in the lidarPoints is consistent with vertical and horizontal raycount.

I think all other checks (from the comment below) can be removed as we changed the API that the user cannot set the angle step size, but can set only the max angle, min angle, and ray count for vertical and horizontal rays. I found that there is a significant chance of a user-side-error while setting step size. It is also a redundant parameter, which may cause behaviour that is undesirable to the user if any single parameter is set incorrectly.

ogre/src/OgreLidarVisual.cc
mihirk284: These will not fail with the current conditions, I'll remove them.

ogre/src/OgreLidarVisual.cc
mihirk284: Shall I make it a function like this instead? [link](https://github.com/ignitionrobotics/ign-rendering/blob/dfcf7c4a574b6b0e94419b7ba1ab3ebce64055f3/ogre/src/OgreGpuRays.cc#L245)

iche033: just noticed that you changed the `SetVerticalRayCount` and `SetHorizontalRayCount` functions to make sure these values are always greater than 1. So we shouldn't need the check now right?

+1 on removing the angle step size.

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,301 @@
...
+  this->dataPtr->lidarPoints = _msg;
+  this->dataPtr->receivedMsg = true;
+}
+
+//////////////////////////////////////////////////
+void OgreLidarVisual::Update()
+{
+  if (!this->dataPtr->receivedMsg || this->dataPtr->lidarPoints.size() == 0)
+  {
+    ignerr << "Message not received. Exiting update function"
```
iche033, line 134: can you change this to `ignwarn` and update the message to say something like: `Empty lidar points. Exiting update function`?

include/ignition/rendering/base/BaseLidarVisual.hh
```diff
@@ -0,0 +1,445 @@
...
+      return this->offset;
+    }
+
+    /////////////////////////////////////////////////
+    template <class T>
+    void BaseLidarVisual<T>::CreateMaterials()
+    {
+      MaterialPtr mtl;
+
+      if (!T::Scene()->MaterialRegistered("Lidar/BlueStrips"))
```
iche033, line 393: you can use `this->Scene()` instead of `T::Scene()`

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,301 @@
...
+
+      line = std::shared_ptr<OgreDynamicLines>(
+                  new OgreDynamicLines(MT_LINE_LIST));
+
+      line->setMaterial("Lidar/BlueRay");
+      mv = std::dynamic_pointer_cast<Ogre::MovableObject>(line);
+      this->Node()->attachObject(mv.get());
+      this->dataPtr->rayLines.push_back(line);
+
+      this->SetVisibilityFlags(0x0FFFFFFF);
```
iche033, line 212: let's remove this line for setting the visibility flags here. I think this should be done later by ign-gazebo.

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,301 @@
...
+  public: std::vector<std::shared_ptr<OgreDynamicLines>> deadZoneRayFans;
+
+  /// \brief Lidar Ray DynamicLines Object to display
+  public: std::vector<std::shared_ptr<OgreDynamicLines>> rayLines;
+
+  /// \brief The current lidar data message
+  public: std::vector<double> lidarPoints;
+
+  /// \brief True if new message is received
+  public: bool receivedMsg;
```
iche033, line 44: can you initialize this to false? It'll also help to use a word other than msg, e.g. `receivedData` or `receivedPoints`

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,301 @@
...
+{
+  this->dataPtr->noHitRayStrips.clear();
+  this->dataPtr->deadZoneRayFans.clear();
+  this->dataPtr->rayLines.clear();
+  this->dataPtr->rayStrips.clear();
+}
+
+
+//////////////////////////////////////////////////
+void OgreLidarVisual::SetPoints(const std::vector<double> &_msg)
```
iche033, line 123: `_msg` -> `_points`

ogre/src/OgreLidarVisual.cc
mihirk284: I edited the `SetVerticalCount` and `SetHorizontalCount` to ensure the values are non-zero. For the case where the value is 1, we do not need to update the step size. Otherwise, the above expression would give a runtime error. For this specific case, the step size does not need to be used as only one ray will be displayed at `minHorizontalAngle` and `minVerticalAngle`.

For all other cases, this expression would be required to calculate step size.

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,306 @@
...
+    horizontalAngle = this->minHorizontalAngle;
+
+    if (j+1 > this->dataPtr->rayLines.size())
+    {
+      // Ray Strips fill in between the line areas that intersect an object
+      std::shared_ptr<OgreDynamicLines> line =
+                  std::shared_ptr<OgreDynamicLines>(
+                              new OgreDynamicLines(MT_TRIANGLE_STRIP));
+
+      line->setMaterial("Lidar/BlueStrips");
```
iche033, line 185: windows build is failing due to the `setMaterial` calls in this function as windows is using a newer version of ogre. Looks like ogre deprecated this API in version 1.10.8 in favor of the overloaded `setMaterial(MaterialPtr &)` function, see [here](https://github.com/OGRECave/ogre/blob/v1.10.8/OgreMain/include/OgreSimpleRenderable.h#L77). Can you add a check for OGRE_VERSION and use the string version if version is <= 1.10.8? here's an example: https://github.com/ignitionrobotics/ign-rendering/blob/master/ogre/src/OgreRenderEngine.cc#L99

ogre/include/ignition/rendering/ogre/OgreScene.hh
```diff
@@ -142,6 +142,10 @@ namespace ignition
       protected: virtual MarkerPtr CreateMarkerImpl(unsigned int _id,
                      const std::string &_name);
 
+      // Documentation inherited
+      protected: virtual LidarVisualPtr CreateLidarVisualImpl(unsigned int _id,
+                     const std::string &_name) override;
```
iche033, line 147: the `override` keyword is correct and the right thing to do. It's however producing warnings on [homebrew build](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-homebrew-amd64/735/warnings10Result/new/). To resolve the warnings, we need to add the keyword to other functions, which would end up increasing the size of this PR. Let's remove `override` in this PR so we can get a clean homebrew build, and we can create a separate PR to fix these warnings later.

ogre/src/OgreLidarVisual.cc
```diff
@@ -0,0 +1,325 @@
...
+    {
+      // Ray Strips fill in between the line areas that intersect an object
+      std::shared_ptr<OgreDynamicLines> line =
+                  std::shared_ptr<OgreDynamicLines>(
+                              new OgreDynamicLines(MT_TRIANGLE_STRIP));
+      #if (OGRE_VERSION <= ((1 << 16) | (10 << 8) | 7))
+            line->setMaterial("Lidar/BlueStrips");
+      #else
+            line->setMaterial(
+                this->Scene()->Material("Lidar/BlueStrips")->Clone());
```
iche033, line 188: This returns an `ignition::rendering MateiralPtr` object but `setMaterial` expects an `Ogre::MateiralPtr` object. So windows build is still failing, see [console](https://build.osrfoundation.org/job/ign_rendering-pr-win/835/console) output

You'll need to get the material using ogre's material manager:

```
Ogre::MaterialPtr mat = Ogre::MaterialManager::getSingleton().getByName("Lidar/BlueStrips");
```

jacobperron:
It looks like this change introduced four new test failures [on Windows](https://build.osrfoundation.org/job/ign_rendering-ci-win/82/testReport/):

- [LidarVisual/LidarVisualTest.LidarVisual/ogre](https://build.osrfoundation.org/job/ign_rendering-ci-win/82/testReport/(root)/LidarVisual_LidarVisualTest/LidarVisual_ogre/)
- [LidarVisual/LidarVisualTest.RaysUnitBox/ogre](https://build.osrfoundation.org/job/ign_rendering-ci-win/82/testReport/(root)/LidarVisual_LidarVisualTest/RaysUnitBox_ogre/)
- [LidarVisual/LidarVisualTest.LaserVertical/ogre](https://build.osrfoundation.org/job/ign_rendering-ci-win/82/testReport/(root)/LidarVisual_LidarVisualTest/LaserVertical_ogre/)
- [LidarVisual/LidarVisualTest.Configure/ogre](https://build.osrfoundation.org/job/ign_rendering-ci-win/82/testReport/(root)/LidarVisual_LidarVisualTest/Configure_ogre/)

iche033:
> It looks like this change introduced four new test failures on Windows:

hmm unfortunately that's expected since windows build is not working properly. There is an attempt to fix windows CI in pull request #109

Issue #100, opened 2020-06-12
Support custom rendering engine plugins


chapulina:
Ignition Rendering provides a plugin interface so that different rendering engines can be loaded at runtime. All the APIs needed for a downstream project to implement their custom engine plugin should already be installed and visible from downstream projects.

However, at the moment, Ignition Rendering is only able to load engines that are provided within this package itself. These default engines are registered here:

https://github.com/ignitionrobotics/ign-rendering/blob/dfcf7c4a574b6b0e94419b7ba1ab3ebce64055f3/src/RenderEngineManager.cc#L337-L367

I'm not sure if there's any code path exposed to users which will allow them to load an engine other than these.

One suggestion for the TODO above is to install the default engines in a path where Ignition Rendering always knows to look for them, so that they don't need to be hardcoded.

---

The ultimate goal is to allow `ign-gazebo` to load custom rendering engines using the `--rendering-engine` option in combination with the `IGN_GAZEBO_RENDERING_ENGINE_PATH` environment variable, like how it's done for physics, see [this tutorial](https://github.com/ignitionrobotics/ign-gazebo/blob/master/tutorials/physics.md).

The [Finding Resources](https://github.com/ignitionrobotics/design/blob/2a2c22e6366ff3d6cf1feece03986d67199d645b/finding_resources.md) design document also offers some guidelines on how to let the users find custom engines.

----------------------------------------------------------------
[Edit - @JShep1]  Mainly adding a couple subtasks here mainly for me to keep track:

- [x] Add unversioned binaries (https://github.com/ignitionrobotics/ign-rendering/pull/141)
- [x] Add custom rendering engine support (https://github.com/ignitionrobotics/ign-rendering/pull/142)
- [x] Add plugin example (https://github.com/ignitionrobotics/ign-rendering/pull/154)
- [x] Add rendering plugin tutorial (https://github.com/ignitionrobotics/ign-rendering/pull/161)

Levi-Armstrong:
I believe this may be related to the current issue I am having with releasing a snap package levering Ignition GUI which fails to find the rendering engines even thought they are located in the snap package. I was having a similar issues with GUI plugins but was able to solve this by setting the `IGN_GUI_PLUGIN_PATH`.

Levi-Armstrong:
I created a PR #104 to support environment variable similar to ignition gui.

chapulina:
I looked a bit into this and got a proof of concept working:

<details><summary>See diff</summary>
[...]

PR #104, opened 2020-06-21
Add environment variable to locate plugins

Levi-Armstrong:
I just copied how this was done in Ignition GUI

### DIFF
--- a/ogre/src/OgreRenderEngine.cc
+++ b/ogre/src/OgreRenderEngine.cc
@@ -78,7 +78,11 @@ RenderEngine *OgreRenderEnginePlugin::Engine() const
 OgreRenderEngine::OgreRenderEngine() :
   dataPtr(new OgreRenderEnginePrivate)
 {
-  this->ogrePaths.push_back(std::string(OGRE_RESOURCE_PATH));
+  this->ogrePaths.push_back(OGRE_RESOURCE_PATH);
+
+  const char *env = std::getenv("OGRE_RESOURCE_PATH");
+  if(env)
+    this->ogrePaths.push_back(std::string(env));
 }
 
 //////////////////////////////////////////////////
--- a/ogre2/src/Ogre2RenderEngine.cc
+++ b/ogre2/src/Ogre2RenderEngine.cc
@@ -82,6 +82,10 @@ Ogre2RenderEngine::Ogre2RenderEngine() :
   std::string ogrePath = std::string(OGRE2_RESOURCE_PATH);
   this->ogrePaths.push_back(ogrePath);
 
+  const char *env = std::getenv("OGRE2_RESOURCE_PATH");
+  if(env)
+    this->ogrePaths.push_back(std::string(env));
+
 #ifdef __APPLE__
   // on OSX the plugins may be placed in the parent lib directory
   if (ogrePath.rfind("OGRE") == ogrePath.size()-4u)
--- a/src/RenderEngineManager.cc
+++ b/src/RenderEngineManager.cc
@@ -83,6 +83,9 @@ class ignition::rendering::RenderEngineManagerPrivate
 
   /// \brief Plugin loader for managing render engine plugin libraries
   public: ignition::plugin::Loader pluginLoader;
+  
+  /// \brief Environment variable which holds paths to look for plugins
+  public: std::string pluginPathEnv = "IGN_RENDERING_PLUGIN_PATH";
 
   /// \brief Mutex to protect the engines map.
   public: std::recursive_mutex enginesMutex;
@@ -373,7 +376,8 @@ bool RenderEngineManagerPrivate::LoadEnginePlugin(
   ignmsg << "Loading plugin [" << _filename << "]" << std::endl;
 
   ignition::common::SystemPaths systemPaths;
-
+  systemPaths.SetPluginPathEnv(this->pluginPathEnv);
+  
   // Add default install folder.
   systemPaths.AddPluginPaths(std::string(IGN_RENDERING_PLUGIN_PATH));
   // Add extra search path.


Levi-Armstrong:
I am not sure why the tests are failing on windows and I do not have a way to test this locally.

Levi-Armstrong:
This change solved the issue with locating the ignition rendering plugins within a snap package, but now at runtime Ogre does not have the OpenGL render plugin. I believe the Ignition Ogre Plugin is setting the path for ogre plugin but I am not certain. Does anyone have any thoughts on the issue?

chapulina:
> I am not sure why the tests are failing on windows and I do not have a way to test this locally.

They were failing before, not this PR's fault.

> now at runtime Ogre does not have the OpenGL render plugin

Interesting, I wouldn't expect the existing plugins to be affected by this change. How are you loading Ogre? Would one of the existing examples in this package fail the same way?

Levi-Armstrong:
It looks like the path it looks for ogre plugins is set at compile time [here](https://github.com/ignitionrobotics/ign-rendering/blob/master/ogre/src/CMakeLists.txt#L23). I am testing a [solution](https://github.com/Levi-Armstrong/ign-rendering/commit/153a5879a162e1b8f9d99d246bbcda668b707cba) at the moment to address this similar to how this is done in ogre. 

Another possible option would be to not provide the full path for the ogre plugins as done [here](https://github.com/ignitionrobotics/ign-rendering/blob/master/ogre/src/OgreRenderEngine.cc#L411-L469) but only include the names when loading ogre plugins. The only reason I can see to not make this change, is it would not support switching to the debug version. The reason behind only using the plugin name is that Ogre already supports bundling so it would detect that it inside a snap package and correctly locate its resources if the plugin name was only provided.

Levi-Armstrong:
So with these changes I have been able to successfully create a snap package leveraging ignition gui. Below is a app configuration setting leveraging the environment variables so resources are located within the snap package.

```
apps:
  tesseract-setup-wizard:
    extensions:
      - kde-neon
    environment:
      IGN_GUI_PLUGIN_PATH: $SNAP/opt/ros/melodic/lib/
      IGN_RENDERING_PLUGIN_PATH: $SNAP/opt/ros/melodic/lib/
      IGN_RENDERING_RESOURCE_PATH: $SNAP/opt/ros/melodic/share/ignition/ignition-rendering4/
      OGRE_RESOURCE_PATH: $SNAP/usr/lib/x86_64-linux-gnu/OGRE-1.9.0/
    command: opt/ros/melodic/bin/demo_window
```

Levi-Armstrong:
If you find these changes acceptable, I can go through and update OGRE2 plugin also.

chapulina:
The approach looks reasonable to me. The only thing I'm wondering is whether we can avoid exposing engine specific configuration to downstream users.

So I could see `OGRE_RESOURCE_PATH` somehow combined with `IGN_RENDERING_RESOURCE_PATH`, or maybe using a more generic variable name such as `IGN_RENDERING_ENGINE_RESOURCE_PATH` which is shared by all rendering engines.

I'm not 100% sure this is a good idea though. Not sure what @iche033 thinks of this?

iche033:
One small edge case issue I can think of for using a shared env variable like `IGN_RENDERING_ENGINE_RESOURCE_PATH` is when users load multiple render engines and need to specify a different resource path per engine. Ign-rendering is designed to support loading multiple rendering engines and switching between them at run time (at least that was the case when optix was working). So users may need to a way to specify different resource paths, and that may be easier to do with engine specific env variables.

Levi-Armstrong:
Another option is how ogre does this. The only reason I had to expose the OGRE_RESOURCE_PATH environment variable was because I am bundling it using snap. In this case ogre looks for the environment variable SNAP and if it exists it prepends the path as shown [here](https://github.com/OGRECave/ogre/blob/92800cbea94c5560d3d9cb0be1462f4e9bb25cf8/OgreMain/src/GLX/OgreFileSystemLayer.cpp#L82-L84).

```
        // With Ubuntu snaps absolute paths are relative to the snap package.
        char* env_SNAP = getenv("SNAP");
        if(env_SNAP && !path.empty() && path[0] == '/') // only adjust absolute dirs
            path = env_SNAP + path;

```

ogre/src/OgreRenderEngine.cc
```diff
@@ -78,7 +78,11 @@ RenderEngine *OgreRenderEnginePlugin::Engine() const
 OgreRenderEngine::OgreRenderEngine() :
   dataPtr(new OgreRenderEnginePrivate)
 {
-  this->ogrePaths.push_back(std::string(OGRE_RESOURCE_PATH));
+  const char *env = std::getenv("OGRE_RESOURCE_PATH");
+  std::string ogre_resource_path = (env) ? std::string(env) :
+      OGRE_RESOURCE_PATH;
```
chapulina, line 83: How about still adding the compile-time resource path even if there's an environment variable? This way the same application could use both pools of resources.

Also, a style nitpick, we use `camelCase` for variables.

Levi-Armstrong: Done.

Levi-Armstrong:
> Besides the inline comment, I'd like to clarify whether this is enough to close #100 (i.e. can we load rendering engine plugins other than Ogre and Optix?).
> 
> I suspect that the answer is no, which is fine and shouldn't block this PR.

I believe it should be enough to load plugins outside this package. Essential that is what is happening when I am bundling this into a snap. Everything is being loaded from the environment variables because the location of all resources are not in the location of the original install location.

Levi-Armstrong:
@chapulina and @iche033 I added one more commit which makes the same change in the ogre2 plugin.

chapulina:
> I believe it should be enough to load plugins outside this package.

That sounds great, thanks!

Unrelated to this PR, but I'd say we need some documentation and a demo on how to load a custom engine before closing #100. Similar to the [ign-physics examples](https://github.com/ignitionrobotics/ign-physics/tree/ign-physics2/examples). I haven't tried to do it myself, but looking at the code, I'm concerned about some assumptions made about the "default engines" in places like this:

https://github.com/ignitionrobotics/ign-rendering/blob/dfcf7c4a574b6b0e94419b7ba1ab3ebce64055f3/src/RenderEngineManager.cc#L287-L312

Levi-Armstrong:
Is this good to merge?

iche033:
merging!

PR #113, opened 2020-07-09
Particle emitter - part2

caguero:
This pull request focuses on the API of the particle emitter. Here are a few points for potential discussion:

**1. How does the `ParticleEmitter` class API set its type?**

Typically, there's a way to specify the geometry of the emitter, that dictates the origin from where the particles are created (point, box, cylinder, ellipsoid, hollow ellipsoid, ring, ...). Each shape can be specified with its own parameters. E.g.: {width, height, depth} for a box and {width, height, depth, internal_width, internal_height, internal_depth} for a hollow ellipsoid. 

1.1: Ogre decides to have a particle emitter for each type subclassing from the `ParticleEmitter` base class. Additionally, they provide a few instances of emitters for the regular types mentioned before. This design allows to create custom shapes for the emitters.

1.2: Another idea could be to create a `Geometry` class that can be subclassed with all the different geometries to support. The `SetType()` function should accept a `Geometry` parameter. Probably, each derived `Geometry` should have its own initialization parameters and all of them should have a way to sample a particle from that geometry.

1.3: The simplest approach that I can imagine is to offer a `SetType()` function where we specify the geometry as an `enum`, and then, to offer a `SetEmitterSize()` that accepts a `Vector3d`. The interpretation of the vector varies depending on the type, but we can at least represent a point, box, cylinder and ellipsoid.

I decided to go for (1.3) for its simplicity and for avoiding extra inheritance. Let me know if you think we should go for a more general approach even at the cost of adding some extra inheritance. I do think that (1.2) is a more elegant and could be a good balanced approach between complexity and flexibility.

**2. When does the emitter start emitting particles?**

2.1 One idea could be to start as soon as it can. This probably means when the emitter is instantiated. It will have default parameters, so it should be possible. As soon as the emitter instance is modified via the API, the behavior should change on the fly. 

2.2 The alternative approach is to use a function (e.g.: `SetEmitting()`) to start/stop emitting particles. Then, you could instantiate your object, set all the parameters, and then, call `SetEmitting(true)` when you're ready to emit particles. Even in this approach we could consider if modifying attributes while the emitter is enabled makes sense. 

This pull request uses (2.2) because the behavior looks a bit more controlled to me. I'm guessing that the expected behavior should be to support changes in the attributes even if we're emitting particles but I can convinced otherwise.

**3. What's the behavior of `SetColorRange()`?**

As far as I know Ogre samples a color from the `[start_color, end_color]`. In our API discussion, we mentioned interpolation between the two colors. Not sure if this means that the particle should have `start_color` when is spawn, should finish with `end_color` at the end of its lifetime, and in between we should interpolate the color smoothly. In this PR, I went for the Ogre approach based on simplicity but again, totally open to discussion.

**4. Getters?**

This is almost trivial but we didn't include it in our initial API. I didn't add them to this patch but I'm almost convinced that we should do it.


include/ignition/rendering/ParticleEmitter.hh
```diff
@@ -36,7 +55,86 @@ namespace ignition
...
+      /// \brief Set a velocity range and each particle is emitted with a
+      /// random velocity within this range (m/s).
+      /// Default value is 1 for both velocities.
+      /// \param[in] _minVel Minimum velocity.
+      /// \param[in] _maxVel Maximum velocity.
+      public: virtual void SetVelocityRange(double _minVel,
+                                            double _maxVel) = 0;
+
+      /// \brief Sets a colour for all particle emitted.
+      /// The actual colour will be randomly chosen between these two values.
```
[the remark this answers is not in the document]
caguero: Sure, changed in 0323230.

caguero:
> looks good to me! I made some minor style changes in [8a7b2ee](https://github.com/ignitionrobotics/ign-rendering/commit/8a7b2ee6e654e89a4e57c87f759ca6ddcd832999), mainly removing `const` from primitive types as we've been doing that recently.

Thanks!

> 
> 1.3: sounds good. I agree we should start with something simple. Using enum's is also consistent with how we are handling different [types of markers](https://github.com/ignitionrobotics/ign-rendering/blob/master/include/ignition/rendering/Marker.hh#L35).
> 
> 2.2: yep also agree with this. I think ogre equivalent is the `setEnabled` function.
> 
> 3: I made an inline comment about doxygen comment of the color range function as it currently says the colors are randomly chosen. I think doing interpolation is better.

Changed in 0323230.

> 
> 4: +1 yeah we should have getters. The accessor functions will also help with unit testing.

Accessors added in 0323230.

src/ParticleEmitter_TEST.cc
```diff
@@ -30,37 +31,139 @@
...
+  ignition::math::Vector3d expectedParticleSize = {100, 100, 100};
+  double                   expectedLifetime     = 5;
+  MaterialPtr              expectedMaterial     = nullptr;
+  double                   expectedMinVel       = 1;
+  double                   expectedMaxVel       = 1;
+  ignition::math::Color    expectedColorStart   = ignition::math::Color::White;
+  ignition::math::Color    expectedColorEnd     = ignition::math::Color::White;
+
+  // Check default expectations.
+  EXPECT_EQ(       expectedEmitterType,  particleEmitter->Type());
```
iche033, line 90: the nice formatting here produced a few code check errors for me :/

```
./src/ParticleEmitter_TEST.cc:90:  Extra space after ( in function call  [whitespace/parens] [4]
```

include/ignition/rendering/ParticleEmitter.hh
```diff
@@ -36,7 +55,173 @@ namespace ignition
...
+      /// \return Pose of the emitter.
+      /// \sa SetLocalPose
+      public: virtual ignition::math::Pose3d LocalPose() const = 0;
+
+      /// \brief Set the position of the emitter relative to the SceneNode the
+      /// particle system is attached to.
+      /// Default value is (0, 0, 0, 0, 0, 0).
+      /// \param[in] _pose Pose of the emitter.
+      /// \sa LocalPose
+      public: virtual void SetLocalPose(
```
iche033, line 119: just realized that you shouldn't need to declare `SetLocalPose` and `LocalPose` here because `ParticleEmitter` should inherit [these two functions](https://github.com/ignitionrobotics/ign-rendering/blob/master/include/ignition/rendering/Node.hh#L61) from `Node` already.

caguero: Good to know. 1843624.

src/ParticleEmitter_TEST.cc
caguero: I don't see any errors on cppcheck `1.82` but I tweaked it a bit. If you still see errors feel free to edit it directly until it's clean. 1843624

iche033: cool errors are gone now. I think it was found by [cppylint.py](https://github.com/ignitionrobotics/ign-rendering/blob/master/tools/cpplint.py#L2529) in `tools` dir when I ran `sh tools/code_check.sh` locally.

PR #116, opened 2020-07-22
Ogre2 Implementation for Lidar Visual

mihirk284:
Added the Ogre2 implementation for Lidar Visual
Pls see attached screenshots

image
image

@iche033 
@ahcorde


examples/lidar_visual/Main.cc
```diff
@@ -219,7 +219,7 @@ LidarVisualPtr createLidar(ScenePtr _scene)
   // LVT_RAY_LINES -> Lines along the lidar sensor to the obstacle
   // LVT_TRIANGLE_STRIPS -> Coloured triangle strips denoting hitting and
   // non-hitting parts of the scan
-  lidar->SetType(LidarVisualType::LVT_RAY_LINES);
+  lidar->SetType(LidarVisualType::LVT_TRIANGLE_STRIPS);
```
ahcorde, line 222: Is there any reason to change this?

ahcorde:
Some tests are failing 

[LidarVisual/LidarVisualTest.RaysUnitBox/ogre (from LidarVisual_LidarVisualTest](https://build.osrfoundation.org/job/ign_rendering-pr-win/860/testReport/junit/(root)/LidarVisual_LidarVisualTest/RaysUnitBox_ogre/)

[LidarVisual/LidarVisualTest.LaserVertical/ogre (from LidarVisual_LidarVisualTest)](https://build.osrfoundation.org/job/ign_rendering-pr-win/860/testReport/junit/(root)/LidarVisual_LidarVisualTest/LaserVertical_ogre/)

ahcorde:
when I try to launch to example with ogre2 I get this error:

```
./lidar_visual ogre2
Starting engine [ogre2]
[Msg] Loading plugin [ignition-rendering4-ogre2]
[Wrn] [Ogre2GpuRays.cc:280] Only one vertical ray but vertical min. and max. angle are not equal. Min. angle is used.
Starting engine [ogre2]
[Err] [BaseRenderEngine.cc:191] Scene already exists with id: 4294967295
```

ahcorde:
I tried to add another render engine here

https://github.com/ignitionrobotics/ign-rendering/blob/e22fd80e25aa89454a2be1d69891c83c7900f4c0/examples/lidar_visual/Main.cc#L279

```bash
  engineNames.push_back("ogre2");
```

To be able to change the render engine with Tab, but I'm getting the following error

```bash
lidar_visual: /build/ogre-1.9-kiU5_5/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreSingleton.h:80: Ogre::Singleton<T>::Singleton() [with T = Ogre::LogManager]: Assertion `!msSingleton' failed.
```

I know this issue is unrelated but do you mind to have a look ?

examples/lidar_visual/Main.cc
mihirk284: I did this as LVT_TRIANGLE_STRIPS was the standard visual before ( before we added types, and the only visual in Gazebo Classic), in order to retain the same in the example.

mihirk284:
@ahcorde  Regarding the tests, I think this is a known test failure which is happening on Windows with other tests as well. I'm not sure why this is failing on Windows

ogre2/include/ignition/rendering/ogre2/Ogre2LidarVisual.hh
```diff
@@ -29,9 +30,13 @@ namespace ignition
   namespace rendering
   {
     inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
-    /// \brief Ogre 2.x implementation of a marker geometry.
+    //
+    // Forward declaration
+    class Ogre2LidarVisualPrivate;
+
+    /// \brief Ogre implementation of a Lidar Visual.
```
iche033, line 37: `Ogre` -> `Ogre 2.x`

iche033:
> lidar_visual: /build/ogre-1.9-kiU5_5/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreSingleton.h:80: Ogre::Singleton<T>::Singleton() [with T = Ogre::LogManager]: Assertion `!msSingleton' failed.

Regarding this error when loading `ogre2` along side of `ogre`, I looked into this some time ago and found that it's likely due to symbol collision between the ogre and ogre2 libraries. When ogre1.x is loaded, a few singletons were created, and when switching to ogre 2.x, it tries to create more singletons that collide with the ones that are already in memory. I think the would either need to 1) namespace the ogre singletons (not sure how to do this without editing ogre itself or 2) completely unload ogre 1.x plugin before switching to ogre 2.x at runtime). 

I haven't figured out a solution to this issue yet so any help / suggestions welcome.

Sarath18:
Hey @mihirk284. I was integrating LidarVisual to ign-rviz and found that `ClearPoints` API doesn't seem to clear the visual that is already been visualized, rather it just clears the internal state and there is no direct API to clear the visual. I went through the ogre LidarVisual code and found a method `ClearVisualData` but it is not exposed to the user. It will be great if we have a method to clear the visuals.

examples/lidar_visual/GlutWindow.cc
```diff
@@ -249,12 +249,19 @@ void updateLidarVisual()
   // change detected due to key press
   if (g_lidarVisualUpdateDirty)
   {
-    g_lidar->SetDisplayNonHitting(g_showNonHitting);
-    g_lidar->SetPoints(g_lidarData);
-    g_lidar->SetType(g_lidarVisType);
-    g_lidar->Update();
+    if(g_clear == true)
```
ahcorde, line 252: https://github.com/ignitionrobotics/ign-rendering/pull/116/checks?check_run_id=922956902

```suggestion
    if (g_clear == true)
```

iche033:
can you fix DCO? instructions in [Details](https://github.com/ignitionrobotics/ign-rendering/pull/116/checks?check_run_id=923268756) next to the DCO check.

iche033:
Regarding the `ClearVisual` and `ClearPoints` functions, I think it could be confusing for the user of the API of what the differences are. As a user, when I clear or update points, I expect the visual to be updated as well otherwise we end up with inconsistent state between the data and the visualization. How about for the public API of the LidarVisual interface, we offer just the function`ClearPoints` and internally we clear both the points and the visualization?

PR #122, opened 2020-08-06
Add support for Gaussian noise render pass in Ogre2DepthCamera 

iche033:
Depth cameras in ign-rendering generates both depth and color data using custom shader scripts so we can not apply regular image based Gaussian noise pass to its output. Instead, I updated Ogre2DepthCamera class to support its own render pass chain, and wrote a separate fragment shader that works with the depth camera data format. 

The Gaussian noise is applied individually to each component in the point cloud data, i.e. x, y, z, and r, g, b. 

Here's a [depth_noise.sdf](https://gist.github.com/iche033/898d7888462b2cdf264a0808942660b0) for testing in ign-gazebo (requires ignitionrobotics/ign-sensors#35). To launch:

```
ign gazebo -v 4 depth_noise.sdf
```

Select the `/camera/image` topic in the ImageDisplay plugin to see the following image output:

depth_color_noise

As for the depth image, the effect is not very noticeable in the `/camera/depth_image` topic in ImageDisplay because I think the conversion of depth to color image normalizes the data and that reduces the effect of noise. Here's a better visualization in rviz that shows noise in the point cloud data generated by the depth camera:

rgbd_noise

### DIFF
--- a/ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
+++ b/ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
@@ -119,6 +119,9 @@ namespace ignition
       /// ogre camera has not been created.
       public: double FarClipPlane() const override;
 
+      // Documentation inherited.
+      public: void AddRenderPass(const RenderPassPtr &_pass) override;
+
       /// \brief Get a pointer to the render target.
       /// \return Pointer to the render target
       protected: virtual RenderTargetPtr RenderTarget() const override;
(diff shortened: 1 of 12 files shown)


chapulina:
This PR broke ABI on Blueprint, then was ported to Citadel and broke ABI there too. Now it's being ported to Dome in #167 and we only caught the ABI breakage at that point.

The ABI breakage happened from:

* 2.4.0 to 2.5.0
* 3.1.0 to 3.2.0

The ABI checker failed on the last commit of this PR, so in theory we should have caught it and at least discussed it:

https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/254/API_5fABI_20report/

We should think about how we can avoid this in the future. One thing I did was making the ABI checker required for all release branches. It can still be overridden by admins, but that should at least be an extra checkpoint.

PR #124, opened 2020-08-11
Add LidarVisual point colors for Ogre1

mihirk284:
@iche033  Please refer to this PR for point color material and implementation, Thanks

EDIT:- adding image
image

To use this you can do the following :-
`lidarPointer->SetPoints(points_vector, colors_vector);`

### DIFF
--- a/ogre/include/ignition/rendering/ogre/OgreLidarVisual.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreLidarVisual.hh
@@ -61,6 +61,7 @@ namespace ignition
               const std::vector<double> &_points) override;
 
       // Documentation inherited
+      // This only affects lidar visuals with type LVT_POINTS
       public: virtual void SetPoints(const std::vector<double> &_points,
                         const std::vector<ignition::math::Color> &_colors)
                                     override;
--- a/ogre/include/ignition/rendering/ogre/OgreLidarVisual.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreLidarVisual.hh
@@ -61,6 +61,11 @@ namespace ignition
               const std::vector<double> &_points) override;
 
       // Documentation inherited
+      public: virtual void SetPoints(const std::vector<double> &_points,
+                        const std::vector<ignition::math::Color> &_colors)
+                                    override;
+
+      // Documentation inherited
       public: virtual void ClearPoints() override;
 
       // Documentation inherited
--- a/ogre/src/OgreLidarVisual.cc
+++ b/ogre/src/OgreLidarVisual.cc
@@ -50,6 +50,9 @@ class ignition::rendering::OgreLidarVisualPrivate
   /// \brief The current lidar points data
   public: std::vector<double> lidarPoints;
 
+  /// \brief The colour of rendered points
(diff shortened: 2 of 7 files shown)


ogre/src/OgreLidarVisual.cc
```diff
@@ -149,6 +152,26 @@ void OgreLidarVisual::ClearVisualData()
 void OgreLidarVisual::SetPoints(const std::vector<double> &_points)
 {
   this->dataPtr->lidarPoints = _points;
+  this->dataPtr->pointColors.clear();
+  for (u_int64_t i = 0; i < this->dataPtr->lidarPoints.size(); i++)
+  {
+    this->dataPtr->pointColors.push_back(ignition::math::Color::White);
```
iche033, line 158: we should keep the color consistent and set it to the same color as `Lidar/BlueRay`.

ogre/include/ignition/rendering/ogre/OgreLidarVisual.hh
```diff
@@ -60,6 +60,11 @@ namespace ignition
       public: virtual void SetPoints(
               const std::vector<double> &_points) override;
 
+      // Documentation inherited
```
iche033, line 63: can you add a note here that it currently only affects lidar visuals with type `LVT_POINTS`?

ogre/src/OgreLidarVisual.cc
```diff
@@ -305,11 +328,14 @@ void OgreLidarVisual::Update()
...
-              line->setMaterial("Lidar/BlueRay");
+              Ogre::MaterialPtr mat =
+                  Ogre::MaterialManager::getSingleton().getByName(
+                                                    "PointCloudPoint");
+            line->setMaterial(mat->getName());
         #else
             Ogre::MaterialPtr mat =
                   Ogre::MaterialManager::getSingleton().getByName(
-                                                    "Lidar/BlueRay");
+                                                    "PointCloudPoint");
```
iche033, line 338: optimization suggestion: can you move the `Ogre::MaterialManager::getSingleton().getByName` calls outside the for loop so we don't have to search for the material every iteration of the loop.

same for the other materials like `Lidar/BlueStrips`, `Lidar/TransBlack`, `Lidar/LightBlueStrips`

iche033:
there are some [windows build errors](https://build.osrfoundation.org/job/ign_rendering-pr-win/899/console), can you take a look?

ogre/src/OgreLidarVisual.cc
```diff
@@ -149,6 +152,26 @@ void OgreLidarVisual::ClearVisualData()
 void OgreLidarVisual::SetPoints(const std::vector<double> &_points)
 {
   this->dataPtr->lidarPoints = _points;
+  this->dataPtr->pointColors.clear();
+  for (auto i = 0ul; i < this->dataPtr->lidarPoints.size(); i++)
```
iche033, line 156: do you need an unsigned long? vector `size()` returns `size_t` which is an unsigned integer

this should work

```
 for (unsigned int i = 0u; i < this->dataPtr->lidarPoints.size(); ++i)
```

PR #132, opened 2020-08-26
minor change in LidarVisual for the case when sensor data is more than sensor range

mihirk284:
This PR attempts to fix the case when the data given to the sensor happens to contain a range value which is greater than, or exactly equal to the sensor range.
For rays when the sensor does not hit anything in the range, it returns `inf` for the ray. However, when we add adding noise to the measurements defaults the `inf` reading exactly equal to  `maxRange` of the sensor. This causes the lidar visual to display the rays, although they are not hitting any object.

Additionally, if a point is a finite value (even if it is greater than the maxRange), the LidarVisual showed it since the only criteria of determining if the sensor ray was hitting or not was if the value returned by the sensor was `inf`.


ahcorde (approved):
LGTM

PR #133, opened 2020-08-28
Adding option for visibility

mihirk284:
@iche033  Please see this PR to solve the issue in [ign-gazebo/#301](https://github.com/ignitionrobotics/ign-gazebo/pull/301)

I have added to the API to set and get the visible state of the visual. I could not access the existing visibility status of the ogre node as they have not provided API's for the same. [[link]](https://forums.ogre3d.org/viewtopic.php?t=23694)

### DIFF
--- a/ogre/src/OgreLidarVisual.cc
+++ b/ogre/src/OgreLidarVisual.cc
@@ -55,6 +55,9 @@ class ignition::rendering::OgreLidarVisualPrivate
 
   /// \brief True if new points data is received
   public: bool receivedData = false;
+
+  /// \brief The visibility of the visual
+  public: bool visible = true;
 };
 
 using namespace ignition;
@@ -482,6 +485,11 @@ void OgreLidarVisual::Update()
     }
     verticalAngle += this->verticalAngleStep;
   }
+
+  // The newly created dynamic lines are having default visibility as true.
+  // The visibility needs to be set as per the current value after the new
+  // renderables are created.
+  this->SetVisible(this->dataPtr->visible);
 }
 
 //////////////////////////////////////////////////
@@ -494,4 +502,11 @@ unsigned int OgreLidarVisual::PointCount() const
 std::vector<double> OgreLidarVisual::Points() const
 {
   return this->dataPtr->lidarPoints;
+}
+
+//////////////////////////////////////////////////
+void OgreLidarVisual::SetVisible(bool _visible)
+{
+  this->dataPtr->visible = _visible;
+  this->ogreNode->setVisible(this->dataPtr->visible);
 }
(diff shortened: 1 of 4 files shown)


ogre/src/OgreLidarVisual.cc
```diff
@@ -349,6 +349,11 @@ void OgreLidarVisual::Update()
       }
     }
 
+    // The newly created dynamic lines are having default visibility as true.
+    // The visibility needs to be set as per the current value after the new
+    // renderables are created.
+    this->SetVisible(this->Visible());
```
ahcorde, line 355: This is a little bit weird. Are you setting the variable with the letter? I think it's because this line, right?
```
      T::SetVisible(_visible);
```

ahcorde (changes requested):
I made a comment

ogre/src/OgreLidarVisual.cc
mihirk284: I'm sorry I don't understand the question. 

I did the above as the visibility changes once the newly created  `DynamicRenderables`  are added to the LidarVisual, the default value of the visibility of the newly created renderable is `true`. Hence I am setting the current value of the visibility again to make sure that the newly added DyanmicRenderables do not have a separate value from the intended current value. On deleting the old ones and creating the new ones (which happens either during changing of visual type or changing the sensor topic or setting DisplayNonHitting false) it used to be visible even though the checkbox was off.

include/ignition/rendering/LidarVisual.hh
```diff
@@ -179,6 +179,14 @@ namespace ignition
       /// \brief Get if non-hitting rays will be displayed
       /// \return Boolean value if non-hitting rays will be displayed
       public: virtual bool DisplayNonHitting() const = 0;
+
+      /// \brief Set visibility of the visual
+      /// \param[in] _visible Visibility of the visual
+      public: virtual void SetVisible(bool _visible) = 0;
```
iche033, line 185: LidarVisual should inherit the `SetVisible` function from Visual class already so you should not have to add a new one here

ogre2/src/Ogre2LidarVisual.cc
```diff
@@ -297,6 +297,11 @@ void Ogre2LidarVisual::Update()
       }
     }
 
+    // The newly created dynamic lines are having default visibility as true.
+    // The visibility needs to be set as per the current value after the new
+    // renderables are created.
+    this->SetVisible(this->Visible());
```
iche033, line 303: I think you can move this call out side of the for loop, e.g. at the end of the function? Same for `OgreLidarVisual` class

ahcorde (approved):

PR #135, opened 2020-09-02
Replace common::Time with std::chrono

ahcorde:
This PR is part of this issue https://github.com/ignitionrobotics/ign-common/issues/61. The idea it's to deprecate the class `common::Time` in favor of `std::chrono`.

Depends on:
 - https://github.com/ignitionrobotics/ign-math/pull/150

### DIFF
--- a/include/ignition/rendering/Scene.hh
+++ b/include/ignition/rendering/Scene.hh
@@ -77,12 +77,21 @@ namespace ignition
 
       /// \brief Get the last simulation update time
       /// \return The last simulation update time
+      public: virtual common::Time IGN_DEPRECATED(4) SimTime() const = 0;
+
+      /// \brief Get the last simulation update time
+      /// \return The last simulation update time
       public: virtual std::chrono::steady_clock::time_point
-        SimTime() const = 0;
+        Time() const = 0;
 
       /// \brief Set the last simulation update time
       /// \param[in] _time Latest simulation update time
-      public: virtual void SetSimTime(
+      public: virtual void IGN_DEPRECATED(4)
+        SetSimTime(const common::Time &_time) = 0;
+
+      /// \brief Set the last simulation update time
+      /// \param[in] _time Latest simulation update time
+      public: virtual void SetTime(
         const std::chrono::steady_clock::time_point &_time) = 0;
 
       /// \brief Get root Visual node. All nodes that are desired to be
--- a/include/ignition/rendering/Scene.hh
+++ b/include/ignition/rendering/Scene.hh
@@ -81,7 +81,7 @@ namespace ignition
 
       /// \brief Get the last simulation update time
       /// \return The last simulation update time
-      public: virtual std::chrono::steady_clock::time_point
+      public: virtual std::chrono::steady_clock::duration
         Time() const = 0;
 
       /// \brief Set the last simulation update time
@@ -92,7 +92,7 @@ namespace ignition
       /// \brief Set the last simulation update time
       /// \param[in] _time Latest simulation update time
       public: virtual void SetTime(
-        const std::chrono::steady_clock::time_point &_time) = 0;
+        const std::chrono::steady_clock::duration &_time) = 0;
 
       /// \brief Get root Visual node. All nodes that are desired to be
       /// rendered in a scene should be added to this Visual or one of its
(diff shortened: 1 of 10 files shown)


include/ignition/rendering/base/BaseScene.hh
```diff
@@ -55,9 +55,11 @@ namespace ignition
 
       public: virtual std::string Name() const override;
 
-      public: virtual common::Time SimTime() const override;
+      public: virtual std::chrono::steady_clock::time_point SimTime()
+        const override;
 
-      public: virtual void SetSimTime(const common::Time &_time) override;
+      public: virtual void SetSimTime(
+        const std::chrono::steady_clock::time_point &_time) override;
```
chapulina, line 62: All functions in the public API should go through a deprecation cycle (deprecate on `ign-rendering4` and remove on `ign-rendering5`).

We can't overload the `SimTime()` function, so this is tricky. I don't know what @iche033 would prefer - no tick-tock, or a change in the function name.

ahcorde: friendly ping @iche033

chapulina: How about changing the functions to `SetTime` and `Time`? Technically `ign-rendering` doesn't care about whether it's displaying a simulation or something else.

iche033: I think the `SetSimTime` and `SimTime` APIs were added a while back mainly for the [gazebo_scene_viewer](https://github.com/ignitionrobotics/ign-rendering/blob/master/examples/gazebo_scene_viewer/SceneManager.cc#L862) example. It's not really used internally. Changing them to `SetTime` and `Time` sounds good.

iche033: as for the tricky situation with changing return type, I remember in the past we would overload the `operator=` function in return type class, i.e. in this case, we overload the `Time::operator=` to take in a `std::chrono` object. Now that we're just changing the functions to `SetTime` and `Time`, we don't have to do this but I thought to bring it up if that's needed somewhere else.

include/ignition/rendering/Scene.hh
```diff
@@ -77,11 +77,22 @@ namespace ignition
 
       /// \brief Get the last simulation update time
       /// \return The last simulation update time
-      public: virtual common::Time SimTime() const = 0;
+      public: virtual common::Time IGN_DEPRECATED(4) SimTime() const = 0;
+
+      /// \brief Get the last simulation update time
+      /// \return The last simulation update time
+      public: virtual std::chrono::steady_clock::time_point
```
chapulina, line 84: Can we use duration to match the rest of the libraries?

chapulina (commented):
I know the deprecated functions don't have tests, but this could be a good opportunity to add simple unit tests for the new functions, if it doesn't take too much time.

include/ignition/rendering/base/BaseScene.hh
```diff
@@ -610,7 +617,8 @@ namespace ignition
 
       protected: std::string name;
 
-      protected: common::Time simTime;
+      protected: std::chrono::steady_clock::duration simTime =
```
chapulina, line 620: We can call the new variable `time` and deprecate `simTime`.

src/base/BaseScene.cc
```diff
@@ -118,17 +118,38 @@ std::string BaseScene::Name() const
...
 
+//////////////////////////////////////////////////
+common::Time BaseScene::SimTime() const
+{
+  std::pair<int64_t, int64_t> secsAndNsecs =
+    math::durationToSecNsec(this->simTime);
+#ifndef _WIN32
+# pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
chapulina, line 140: Technically these aren't necessary because `ign-rendering4` is using `ign-common3`, which still supports `common::Time`.

chapulina (commented):
We should add a note to the migration guide.

Also bump the `ign-math` version on `CMakeLists` to 6.6.0.

chapulina:
`ign-math 6.6.0~pre1` debs are out, testing this PR against that here:

[Build Status](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/800/)

Using this branch: https://github.com/ignition-tooling/release-tools/tree/chapulina/rendering/pr/135

include/ignition/rendering/base/BaseScene.hh
chapulina: f063b087c8f789ccb761d77d5acc6967ee1a5370

chapulina (approved):

PR #141, opened 2020-09-23
Add unversioned symlinks for engine plugins

JShep1:
Another step towards getting full support for custom render engines from command line.  This PR symlinks binaries that don't contain the version number (same as how physics is doing for `dartsim` and `tpe`) so the user will not have to specify the version of ignition-rendering they are currently using when specifying an engine binary name.  More progress towards getting https://github.com/ignitionrobotics/ign-rendering/issues/100 resolved

Signed-off-by: John Shepherd <john@openrobotics.org>

### DIFF
--- a/ogre/src/CMakeLists.txt
+++ b/ogre/src/CMakeLists.txt
@@ -51,10 +51,14 @@ set (versioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_LOWER}-${engine_name
 set (unversioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_NO_VERSION_LOWER}-${engine_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
 
 if (WIN32)
-  # create_symlink requires cmake 3.13 on windows
-  cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+  # disable MSVC inherit via dominance warning
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4250")
+  INSTALL(CODE "EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E copy
+  ${IGNITION_RENDERING_ENGINE_INSTALL_DIR}\/${versioned} 
+  ${IGNITION_RENDERING_ENGINE_INSTALL_DIR}\/${unversioned})")
+else()
+  EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E create_symlink ${versioned} ${unversioned})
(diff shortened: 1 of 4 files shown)


ogre/src/CMakeLists.txt
```diff
@@ -35,12 +37,24 @@ target_link_libraries(${ogre_target}
...
 
+set (versioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_LOWER}-${engine_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
+set (unversioned ${CMAKE_SHARED_LIBRARY_PREFIX}${PROJECT_NAME_NO_VERSION_LOWER}-${engine_name}${CMAKE_SHARED_LIBRARY_SUFFIX})
+
+if (WIN32)
+  # create_symlink requires cmake 3.13 on windows
+  cmake_minimum_required(VERSION 3.13 FATAL_ERROR)
+endif()
+EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E create_symlink ${versioned} ${unversioned})
+INSTALL(FILES ${PROJECT_BINARY_DIR}/${unversioned} DESTINATION ${IGNITION_RENDERING_ENGINE_INSTALL_DIR})
```
iche033, line 58: [windows build](https://build.osrfoundation.org/job/ign_rendering-pr-win/1015/) failed as it has problem with installing sym links. I think the dartsim plugin is not built on windows so jenkins does not catch the problem but we ran into this with TPE and had to do a copy instead of a symlink on windows:
https://github.com/ignitionrobotics/ign-physics/blob/master/tpe/plugin/CMakeLists.txt#L42-L51

chapulina: As a reference, we decided to use the `cmake_minimum_required(VERSION 3.13 FATAL_ERROR)` approach on [this PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/222/page/1#comment-94939048). I guess it doesn't work afterall. CC @scpeters

scpeters: I thought it requires special privileges on windows to create symlinks:

* https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-gazebo/pull-requests/222/page/1#comment-94939827
* https://www.howtogeek.com/howto/16226/complete-guide-to-symbolic-links-symlinks-on-windows-or-linux/

JShep1: I added the build rule for windows that was in TPE, I guess as we work on Windows compatibility going forward in Q4, I can do more native Windows build testing in the next few months?

iche033: sounds good. We need a bit of help on the Windows side..!

iche033 (approved):
new windows build looks good

scpeters:
we need to update the debian metadata so that the `engine-plugins` folder will be included in our debian packages: https://github.com/ignition-release/ign-rendering4-release/issues/2

PR #143, opened 2020-09-24
Updates to tutorials

chapulina:
* Instruct user to clone this repository: users landing in https://ignitionrobotics.org/api/rendering/4.0/tutorials.html don't know where to start from otherwise
* Split command from command results, so it's clearer to the user what they need to do
* Fix the custom shaders tutorial - it was telling the user to compile the camera tracking

I updated many tutorials, but I only walked through these ones:

* custom shaders
* particles
* text geometry
* render pass

There may be more things to fix on the other ones.


iche033 (approved):
looks good to me

PR #147, opened 2020-09-26
Update gazebo scene viewer tutorial

iche033:
* Fix compilation against gazebo9. When building the examples, it always includes the `Helper.hh` header from math4 (gz9 dependency) instead of math6, causing compilation error due to missing `math::secNsecToDuration` function. The changes make sure that the examples are able to build with either math4 or math6 version.
* moved the Code section to the bottom so that the tutorial layout is consistent with other tutorials


tutorials/11_gazebo_scene_viewer_tutorial.md
```diff
@@ -70,3 +50,23 @@ Launch the example to visualize the objects:
 You can use the `Tab` button to change the rendering engine.
 
 ![](img/gazebo_scene_viewer2_demo.gif)
```
chapulina, line 52: One thing I noticed is that these images are being rendered too large, and the page is not horizontally scrollable. I think we should reduce their size.

chapulina (approved):
Works for me, I tested against Gazebo 11. 

I just have a question and a comment about the image sizes, which is not introduced by this PR.

tutorials/11_gazebo_scene_viewer_tutorial.md
iche033: reized gifs to 50%. 8c6fb3c

examples/gazebo_scene_viewer/SceneManager.cc
```diff
@@ -1812,8 +1815,8 @@ CurrentSceneManager::~CurrentSceneManager()
 void CurrentSceneManager::OnPoseUpdate(::ConstPosesStampedPtr &_posesMsg)
 {
   // record pose timestamp
-  this->timePosesReceived = math::secNsecToDuration(
-      _posesMsg->time().sec(),  _posesMsg->time().nsec());
+  this->timePosesReceived = std::chrono::seconds(_posesMsg->time().sec()) +
+      std::chrono::nanoseconds(_posesMsg->time().nsec());
```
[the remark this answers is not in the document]
iche033: had an issue with compiling against gazebo9 as it always find the math4 headers instead of math6. There could be a way to tweak cmake include directory search order but I figured it's easier to change the code so that the example works with either ign-math versions.

PR #148, opened 2020-09-26
Update custom shaders tutorial

iche033:
* copied the custom shaders over to the example's own media directory so that demo is self-contained.
* tweaked the depth camera shaders to produce a nicer grayscale appearance for objects in the scene (it used to be all white which does not really show that the shaders are doing anything)
* updated tutorial to explain better the two uses of shaders


examples/custom_shaders/CMakeLists.txt
```diff
@@ -36,3 +36,8 @@ target_link_libraries(custom_shaders
   ${GLEW_LIBRARIES}
   ${IGNITION-RENDERING_LIBRARIES}
 )
+
+add_custom_command(TARGET custom_shaders POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_directory
+  ${CMAKE_SOURCE_DIR}/media
```
chapulina, line 42: Maybe you forgot to commit this directory?

```
Error copying directory from "/home/chapulina/dome_ws/src/ign-rendering/examples/custom_shaders/media" to "/home/chapulina/dome_ws/src/ign-rendering/examples/custom_shaders/build/media".
```

iche033: oops, added in f36a6134cd0d01a63b8d819b783434eb8c8c8f0a

chapulina (approved):

PR #153, opened 2020-09-29
Fire breathing duck

JShep1:
Just a small suggestion, I had to do it. :smile: :duck: :fire: 

Signed-off-by: John Shepherd <john@openrobotics.org>


chapulina (approved):
:rofl:

iche033:
It's a crime that a fire breathing duck PR does not come with a gif

fire-breathing-ducky

PR #155, opened 2020-10-01
Limit number of shadow casting lights in ogre2

iche033:
It was found that when the number of shadow casting lights exceed a certain number, the dynamically generated shaders produce a compile error that suggests it's reached the max number of uniform variables allowed. This causes an rendering issue where parts of the scene becomes dark. So the workaround here is to put a hard limit on the max number of lights allowed in ogre2. 

Here's a [max_shadows.sdf](https://gist.github.com/iche033/62d2e226fd726c9f0241e2f84894bea5) world file with lights for testing. Before the fix, the model in the world will be dark and you'll see the compiler error in `~/.ignition/rendering/ogre2.log`. After the fix, the model should render fine.


ogre2/src/Ogre2RenderTarget.cc
```diff
@@ -346,6 +346,17 @@ void Ogre2RenderTarget::UpdateShadowNode()
...
+  // limit number of shadow maps
+  // shaders dynamically generated by ogre produce compile error at runtime if
+  // the number of shadow maps exceeds certain number. The error seems to
+  // suggest that the number of uniform variables has exceeded the max number
+  // allowed
+  unsigned int maxShadowMaps = 25u;
+  dirLightCount = std::min(static_cast<unsigned int>(maxShadowMaps / 3),
+      dirLightCount);
+  spotPointLightCount = std::min(std::max(maxShadowMaps - dirLightCount*3u, 0u),
+      spotPointLightCount);
```
chapulina, line 358: This looks like something that may confuse users down the line. How about printing a warning when the maximum number is reached? I think the warning should only be printed once.

chapulina (approved):
LGTM, I just think we should warn the user about what's happening.

ogre2/src/Ogre2RenderTarget.cc
iche033: yep added warning msg in 7b84199

chapulina (approved):

peci1:
Is this actually somehow discussed with OGRE people? It seems that 25 light sources is pretty easy to get with the numbers of robots allowed...

iche033:
Not yet. I should check to see if it's still an issue in the latest version of OGRE 2.x. It's probably rare for games to have that many shadows.

PR #158, opened 2020-10-07
Fix unloading render engines

iche033:
The tests are still unloading the render engines using the engine name (`ogre2`) as opposed to the library name (`ign-rendering-ogre2`). The changes add an additional logic to translate the engine name to lib name before unloading the engines.


JShep1 (approved):
Good catch, thanks for making these fixes!

PR #167, opened 2020-10-23
Merge 3 to 4

iche033:
checking CI


examples/hello_world_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+
+set(IGN_PLUGIN_VER 1)
+set(IGN_COMMON_VER 3)
+
+find_package(ignition-rendering3 REQUIRED)
```
chapulina, line 6:
```suggestion
find_package(ignition-rendering4 REQUIRED)
```

ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
```diff
@@ -119,6 +119,9 @@ namespace ignition
       /// ogre camera has not been created.
       public: double FarClipPlane() const override;
 
+      // Documentation inherited.
+      public: void AddRenderPass(const RenderPassPtr &_pass) override;
```
chapulina, line 123: The ABI checker is complaining about this. Was it ok to add it to `ign-rendering3`?

tutorials/03_rendering_plugins.md
```diff
@@ -0,0 +1,35 @@
...
+## Ignition Rendering
+
+The \ref ignition::rendering "Ignition Rendering" library integrates external rendering engines into the Ignition Simulation eco-system.
+It allows users to select from multiple supported rendering engines based on their simulation needs.
+Its plugin interface loads rendering engines at runtime.
+It is also possible to integrate your own selected rendering engine by writing a compatible plugin interface.
+
+#### How to Write Your Own Rendering Engine Plugin
+
+A mocked example of a custom rendering engine plugin can be found [here](https://github.com/ignitionrobotics/ign-rendering/tree/ign-rendering3/examples/hello_world_plugin).  In order
```
chapulina, line 14:
```suggestion
A mocked example of a custom rendering engine plugin can be found [here](https://github.com/ignitionrobotics/ign-rendering/tree/ign-rendering4/examples/hello_world_plugin).  In order
```

There are more instances in this file

iche033: fixed. 20d6471

examples/hello_world_plugin/CMakeLists.txt
iche033: fixed. 20d6471

ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
iche033: I don't remember if the issue was there when merging rendering2 to rendering3 but it could have slipped through. I started trying to workaround this but realized that this would be removing a public function that was previously added in rendering2 and rendering3.  The ABI report shows that the difference is in 2 entries in the vtable related to `AddRenderPass()`, which is that now the derived `Ogre2DepthCamera::AddRenderPass` function will be called instead of the base class. It's the intended behavior change but I don't know if that'll cause undefined behavior for downstream libraries. We can either

1. Leave the function here as it is with the ABI issue

1. Remove the function and find a workaround in order keep the depth noise feature (that relies on the AddRenderPass) and update the migration guide to explain reason behind the function removal. 

I'm tempted to go with 1. but either way works.

chapulina: I left a comment on the offending PR here: https://github.com/ignitionrobotics/ign-rendering/pull/122#issuecomment-717560255

At this point, I guess we may as well go with option 1. I think we should add a note to the migration guide about the breakage for all versions from Blueprint to Citadel.

chapulina (approved):

PR #168, opened 2020-10-26
Add ogre2 skybox support 

iche033:
Add new functions in Scene for setting background material (`SetBackgroundMaterial`) and also enabling skybox (`SetSkyEnabled`). The `SetSkyEnabled` function is just a convenient function to set the background material to the default skybox cubemap texture supplied by ign-rendering.

Currently only working in ogre2 with regular RGB `Ogre2Camera`. There has been some changes in `Ogre2DepthCamera` compositor setup in ign-rendering2 that has not merged into `main` yet so I'm planning to merge the changes forward first before updating `Ogre2DepthCamera` to support skybox, e.g. so that it's also visible in an ign-sensors' `RgbdCameraSensor`

see screenshot from `ogre2_demo`

ogre2_demo_sky


ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -283,6 +283,32 @@ void Ogre2DepthCamera::CreateDepthTexture()
...
 
+  // create background material is specified
+  MaterialPtr backgroundMaterial = this->Scene()->BackgroundMaterial();
+  bool validBackground = backgroundMaterial &&
+      !backgroundMaterial->EnvironmentMap().empty();
+
+  if (validBackground)
+  {
+    Ogre::MaterialManager &matManager = Ogre::MaterialManager::getSingleton();
+    std::string skyMatName = "SkyBox_" + this->Name();
```
chapulina, line 294: The `SkyBox` string is repeated in a few places. How about storing it somewhere as a constant?

ogre2/src/Ogre2RenderTarget.cc
```diff
@@ -110,12 +110,144 @@ Ogre2RenderTarget::~Ogre2RenderTarget()
...
+  // sky is enabled.
+  // todo(anyone) Note the definition programmatically created here
+  // replaces the one defined in the script so it maybe safe to remove the
+  // PbsMaterials.compositor file
+  std::string wsDefName = "PbsMaterialWorkspace_" + this->Name();
+  this->ogreCompositorWorkspaceDefName = wsDefName;
+  if (!ogreCompMgr->hasWorkspaceDefinition(wsDefName))
+  {
+    // PbsMaterialsRenderingNode
+    std::string nodeDefName = wsDefName + "/PbsMaterialsRenderingNode";
```
chapulina, line 134: There are a few strings being reused here too, we should consider storing them in a common place.

ogre2/src/Ogre2Scene.cc
```diff
@@ -567,3 +569,39 @@ bool Ogre2Scene::ShadowsDirty() const
...
+    {
+      camera->SetBackgroundMaterial(skyboxMat);
+    }
+  }
+}
+
+//////////////////////////////////////////////////
+bool Ogre2Scene::SkyEnabled() const
+{
+  return this->BackgroundMaterial() != nullptr;
```
chapulina, line 606: Is it possible that the user sets a different background material using `SetBackgroundMaterial`, and this returns true even though the sky is not enabled?

ogre2/src/media/materials/programs/skybox_fs.glsl
```diff
@@ -0,0 +1,16 @@
+#version 330
```
chapulina, line 1: I wonder if we should be putting copyright headers in these files.

ogre2/src/media/materials/programs/skybox_vs.glsl
```diff
@@ -0,0 +1,22 @@
...
+uniform mat4 worldViewProj;
+
+out gl_PerVertex
+{
+    vec4 gl_Position;
+};
+
+out block
+{
+    vec3 cameraDir;
```
chapulina, line 14: Nit: the indentation is not consistent, sometimes 2, sometimes 4 spaces.

test/integration/sky.cc
```diff
@@ -0,0 +1,155 @@
...
+
+/////////////////////////////////////////////////
+void SkyTest::Sky(const std::string &_renderEngine)
+{
+  // create and populate scene
+  RenderEngine *engine = rendering::engine(_renderEngine);
+  if (!engine)
+  {
+    igndbg << "Engine '" << _renderEngine
+              << "' is not supported" << std::endl;
```
chapulina, line 50: I think these debug messages won't show. Adding this above should do it:

```
  // Documentation inherited
  public: void SetUp() override
  {
    ignition::common::Console::SetVerbosity(4);
  }
```

include/ignition/rendering/Scene.hh
```diff
@@ -954,6 +964,14 @@ namespace ignition
       public: virtual ParticleEmitterPtr CreateParticleEmitter(
                   unsigned int _id, const std::string &_name) = 0;
 
+      /// \brief Enable sky in the scene.
+      /// \param[in] _enabled True to enable sky
+      public: virtual void SetSkyEnabled(bool _enabled) = 0;
+
+      /// \brief Get whether the sky is enabled in the scene.
+      /// \return true to sky is enabled, false otherwise
+      public: virtual bool SkyEnabled() const = 0;
```
chapulina, line 973: The current sky API doesn't allow setting sky properties like cloud speed, humidity, ambient color... Are these supposed to be set through a custom background material?

chapulina (commented):
Works for me! I just have some smaller comments and questions.

ogre2/src/Ogre2DepthCamera.cc
iche033: done in 0823885 and fbae46078e563bbda29f6dbfd713f4cd4e976919

ogre2/src/Ogre2RenderTarget.cc
iche033: done in 0823885 and fbae460

ogre2/src/Ogre2Scene.cc
iche033: added a variable to keep track of sky state 0823885

ogre2/src/media/materials/programs/skybox_fs.glsl
iche033: these were adapted from ogre samples so added their license header and a note 0823885

ogre2/src/media/materials/programs/skybox_vs.glsl
iche033: fixed. 0823885

test/integration/sky.cc
iche033: thanks, added. 0823885

include/ignition/rendering/Scene.hh
iche033: those are not supported yet so I held off on adding new APIs for setting these properties until we implement the features.

chapulina: Got it. The reason I asked is because it's not clear to me where those would go. Would they all be functions under `Scene`, like `Scene::SetSkyHumidity`, or do we need a `Sky` class to handle that, and the `Scene` only provides accessors to it (i.e. `Scene::CreateSky` / `Scene::Sky`)?

iche033: Since there should only be one global sky, I am leaning towards something like `Scene::SetSkyProperties(const SkyConfiguration &_config)` function. Open to suggestions though

chapulina: Got it, that sounds good to me and I think it fits well with the functions added in this PR :+1:

ogre2/src/Ogre2DepthCamera.cc
chapulina: Thanks. Does the `SkyBox` string in `RenderTarget` need to match the one in `DepthCamera`? Keeping both in the same place could help future readers understanding how everything fits together, and prevent errors in case one is changed but not the other.

I'll leave it up to you if you think it's worth it.

chapulina (approved):
:sun_behind_large_cloud:

chapulina:
The `ign-gazebo` PR works well with this. I'll go ahead and merge this!

PR #171, opened 2020-11-03
Updated to work with ign-plugin

ahcorde:
This PR depends and needs a release from this other PR https://github.com/ignitionrobotics/ign-plugin/pull/30

All the symbols are loaded in the executable because of the `RTLD_GLOBAL`. This logic will filter the symbols

### DIFF
--- a/src/RenderEngineManager.cc
+++ b/src/RenderEngineManager.cc
@@ -401,7 +401,12 @@ bool RenderEngineManagerPrivate::LoadEnginePlugin(
     return false;
   }
 
-  std::string pluginName = *pluginNames.begin();
+  std::string pluginName;
+  for (auto plugin: pluginNames) {
+    if (plugin.find("rendering") != std::string::npos) {
+      pluginName = plugin;
+    }
+  }
   if (pluginName.empty())
   {
     ignerr << "Failed to load plugin [" << _filename <<


chapulina:
Let me just check if I understood why this PR is needed. If https://github.com/ignitionrobotics/ign-plugin/pull/30 gets in, `ign-rendering` won't work properly anymore unless this filtering is done?

This sounds like a breaking change on `ign-plugin`, that will required a major version bump on that library. 

I'm also curious about what other `pluginNames` are showing up here that don't contain `rendering`. Do you remember?

ahcorde:
Yes, this may require some more mayor version bumps: ign-gazebo, ign-gui, ign-rendering, sensors.

> I'm also curious about what other pluginNames are showing up here that don't contain rendering. Do you remember?

I will check it.

chapulina:
With the proposal on https://github.com/ignitionrobotics/ign-plugin/pull/35, I think this PR isn't necessary.

chapulina:
Since https://github.com/ignitionrobotics/ign-rendering/pull/236/, we're looking explicitly for plugins that implement the `ignition::rendering::RenderEnginePlugin` interface, and in case we find several, we load the first one and print a warning to the user.

So I'm closing this one because I think it is no longer necessary. Please speak up if you think otherwise, @ahcorde .

PR #180, opened 2020-11-25
Heightmap for Ogre 1

chapulina:
This is a port of Gazebo classic's heightmap implementation.

* New generic APIs so any render engine can implement heightmaps
* Implementation for Ogre 1 with tests
* Example
* Tutorial

I'm testing this locally with Ogre 1.9.1.

image

*Update*: Also got it to work with 1.9.0 (debs or source with [this patch](https://gist.github.com/chapulina/37ffbc3f54139c2ca849094cfb07c3ae))

### Geometry vs Object

~~My initial implementation had heightmaps as a new geometry type that can be attached to visuals. That's different from Gazebo classic, but it matches the SDF spec.~~

~~However, Ogre terrains don't seem to have an [Ogre::MovableObject](https://github.com/ignitionrobotics/ign-rendering/blob/f8ec77282d02a0b92de1ee5b24c5a3bd4948775e/ogre/include/ignition/rendering/ogre/OgreGeometry.hh#L45), which is required for `OgreGeometry`s. Looking online, it doesn't seem like terrains are meant to be attached to nodes and moved around. So I stopped pursuing that direction, but left it on branch `chapulina/5/heightmap`. I don't think we should make decisions about `ign-rendering`'s API based solely on Ogre though, so I'm open to bringing that back if we think that it will fit well with other engines.~~

~~For now, heightmaps are extending `Object`, which seemed to be the correct class for entities that won't be moving too much.~~

Heightmaps are now a Geometry, see conversation below.

### Multiple heightmaps

~~The current implementation leaves room for loading multiple heightmaps at the same time. It currently loads the geometry well, but the materials can't be loaded yet. I'm planning to keep looking into fixing that.~~

Multiple heightmaps are now working, see example.

### Scene API

~~The only function added to Scene so far was `CreateHeightmap` and it's storing the heightmaps in a simple vector. I didn't want to go too far until we ruled out the geometry approach. If we keep heightmaps as objects that are managed by the scene, I think we should:~~

* ~~Create a `HeightmapStorage` class~~
* ~~Add `HasHeightmap*` / `HeightmapBy*` / `HeightmapCount` / `DestroyHeightmap` APIs to `Scene`~~

~~I can do that as part of this PR if we want to take this direction.~~

Punting on this now that heightmaps are geometries.

### Descriptor

I took some inspiration from `MeshDescriptor` and stored all the information for the heightmap into `HeightmapDescriptor`. Some key differences:

* `HeightmapDescriptor` is a first-class element of `Heightmap`, while `Mesh` doesn't know about `MeshDescriptor`.
* `Heightmap` stores a copy of `HeightmapDescriptor` and exposes that through the `Descriptor()` function. That's different to storing and accessing each property separately. If this is an acceptable pattern, I can look into PIMPLizing it so it can be extended in the future.

### Material

Admittedly, I blindly copied the terrain material code from Gazebo classic and haven't looked closely into it yet. There may be potential for using `ignition::rendering::Material` inside `Heightmap`, but judging by all the custom `TerrainMaterial` code, that may not be possible or desirable. I just wanted to make sure I brought this up.

### Functionality

- [x] Multiple textures with blending
- [x] Paging
- [x] Caching - not working yet. Saving seems to work, but loading isn't working yet.
- [x] Loading multiple heightmaps - material is broken


iche033:
**Geometry vs Object**: looking at threejs as another example, it treats heightmap as a geometry. I think semantically it makes sense. The OgreObject() will then just return null. We may have to update `OgreVisual::AttachGeometry` function to work with this special case of a null obj. The other option is to derive the Heightmap class from Visual, so when the user calls CreateHeightmap, it's creating the node and the geometry at the same time. I'm leaning towards the Geometry approach but not strongly attached to it.

**Multiple heightmaps**: sounds good

**Scene API**: If it's derived from Geometry, we could create a general GeometryStorage class. It looks like currently other geometries (e.g. text and markers, etc) are not stored in a storage. So we could start storing them and offer APIs for retrieving and destroying geometries. This can be ticketed as an issue and addressed separately. If Heightmap is derived from Visual, then we can just call Scene::RegisterVisual to store it in the visual store

**Descriptor**: MeshDescriptor class is a good idea

**Material** ogre heightmap material is tricky. It may not be trivial to make it work with a generic `rendering::Material` yet. I'm leaning towards not offering the API for users to set heightmap material for now.

chapulina:
Thanks for the input, @iche033 . I'll revert to inheriting from `Geometry` and fix whatever is needed from there. While I'm working on the `ign-gazebo` integration, I'm also seeing that approach should make things easier there.

chapulina:
All outstanding issues have been addressed, this is ready for review! The tutorial example loads 2 heightmaps, which isn't possible in Gazebo classic. Caching and paging are also supported.

heightmap_ign

iche033:
I'm getting a couple of compile errors, which are also happening on jenkins:

```
/home/osrf/code/ign_d_ws2/src/ign-rendering/ogre/src/OgreHeightmap.cc:1503:1: error: ‘class IgnTerrainMatGen::SM2Profile::ShaderHelperGLSL’ is protected within this context
 IgnTerrainMatGen::SM2Profile::ShaderHelperGLSL::
```

chapulina:
> compile errors

Ahh sorry, it should be fixed now: b3efe1c

ogre/src/OgreHeightmap.cc
```diff
@@ -0,0 +1,3304 @@
...
+      this->dataPtr->ignMatGen = new IgnTerrainMatGen();
+
+    auto ptr = Ogre::TerrainMaterialGeneratorPtr();
+    ptr.bind(this->dataPtr->ignMatGen);
+
+    this->dataPtr->terrainGlobals->setDefaultMaterialGenerator(ptr);
+#else
+    // init custom material generator
+    Ogre::TerrainMaterialGeneratorPtr terrainMaterialGenerator;
+    auto terrainMaterial = OGRE_NEW TerrainMaterial("Ignition/Grey");
```
iche033, line 1054: `Ignition/Grey` does not exist. You can use [Default/White](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/src/base/BaseScene.cc#L1275) since this should just be a dummy material.

src/HeightmapDescriptor.cc
```diff
@@ -0,0 +1,359 @@
...
+  public: math::Vector3d size{1.0, 1.0, 1.0};
+
+  /// \brief Heightmap XYZ origin in meters.
+  public: math::Vector3d position{0.0, 0.0, 0.0};
+
+  /// \brief Flag that enables/disables the terrain paging
+  public: bool useTerrainPaging{false};
+
+  /// \brief Number of samples per heightmap datum.
+  public: unsigned int sampling{2u};
```
iche033, line 66: 2 was a bad default to use actually. This does super sampling and hence doubled the number of vertices in gazebo-classic, which caused a performance hit. We should set this to 1. I believe 2 was used a long time ago when we wanted to be able to have finer control over the heightmap using the (disabled) [terrain editor](https://github.com/osrf/gazebo/tree/gazebo11/gazebo/gui/terrain).

ogre/src/OgreHeightmap.cc
```diff
@@ -0,0 +1,3304 @@
...
+  public: std::vector<std::vector<float>> subTerrains;
+
+  /// \brief When the terrain paging is enabled, the terrain information
+  /// for every piece of terrain is stored in disk. This is the path of
+  /// the top level directory where these files are located.
+  public: std::string pagingDir;
+
+  /// \brief Name of the top level directory where all the paging info is
+  /// stored
+  public: const std::string pagingDirname{"paging"};
```
iche033, line 409: maybe rename to `ogre-paging` (to leave room for ogre2-paging later)?

ogre/include/ignition/rendering/ogre/OgreHeightmap.hh
```diff
@@ -0,0 +1,126 @@
...
+      /// \param[in] _heightmap Source vector of floats with the heights.
+      /// \param[in] _n Number of subterrains.
+      /// \param[out] _v Destination vector with the subterrains.
+      private: void SplitHeights(const std::vector<float> &_heightmap,
+                  const int _n, std::vector<std::vector<float>> &_v);
+
+      /// \brief Define a section of the terrain.
+      /// \param[in] _x X coordinate of the terrain.
+      /// \param[in] _y Y coordinate of the terrain.
+      private: void DefineTerrain(const int _x, const int _y);
```
iche033, line 101: I think we can leave out `const` for primitive types. I've started doing that to be consistent with other libraries.

ogre/src/OgreHeightmap.cc
```diff
@@ -0,0 +1,3304 @@
...
+    };
+
+    // Needed to allow access from ShaderHelperGLSL to protected members
+    // of SM2Profile.
+    friend ShaderHelperGLSL;
+
+    /// Keeping the CG shader for reference.
+    /// \brief Utility class to help with generating shaders for Cg / HLSL.
+    /// Original implementation from Ogre that generates Cg shaders
+    protected: class ShaderHelperCg :
```
iche033, line 208: we can actually remove all `ShaderHelperCg` code and references to ShaderHelperCg. It was disabled and not being maintained in gazebo classic, and ign-rendering is not loading the Cg plugin.

iche033 (commented):
works for me! Did a first pass and left some minor comments

src/HeightmapDescriptor.cc
chapulina: This is matching the [SDFormat default](https://github.com/osrf/sdformat/blob/ed198a063cfabd5be3b0d8de4b3787eea7a1f68a/sdf/1.8/heightmap_shape.sdf#L40). I'll open a PR against `sdf11` changing it there too.

ogre/src/OgreHeightmap.cc
chapulina: Removing code feels good! 915fdd495d141b01a44c8259947b1f2b2cf95d5f

src/HeightmapDescriptor.cc
chapulina: https://github.com/osrf/sdformat/pull/459

tutorials/21_heightmap.md
```diff
@@ -0,0 +1,56 @@
...
+```{.sh}
+[Msg] Loading plugin [ignition-rendering5-ogre]
+[Msg] Loading heightmap: scene::Heightmap(65528)
+[Msg] Heightmap loaded. Process took 217 ms.
+===============================
+  TAB - Switch render engines
+  ESC - Exit
+===============================
+```
+@image html img/heightmap.png
```
iche033, line 36: `heightmap**s**.png`

iche033:
@osrf-jenkins run tests please

include/ignition/rendering/HeightmapDescriptor.hh
```diff
@@ -0,0 +1,255 @@
...
+
+    /// \brief Set whether the heightmap uses terrain paging. Defaults to false.
+    /// \param[in] _use True to use.
+    public: void SetUseTerrainPaging(bool _use);
+
+    /// \brief Get the heightmap's sampling per datum.
+    /// \return The heightmap's sampling.
+    public: unsigned int Sampling() const;
+
+    /// \brief Set the heightmap's sampling. Defaults to 2.
```
iche033, line 212: `Defaults to 1`

iche033 (commented):
looks good! waiting for one more round of CI builds

iche033 (approved):
I see lots of greens!

PR #182, opened 2020-12-02
Add support for lightmaps in ogre2

iche033:
depends on ignitionrobotics/ign-common#132

main changes:
* updated mesh factory to create mesh with multiple texture coordinate sets
* added lightmap APIs to Material - users can choose which texture coordinate set to use for the lightmap
* To keep things simple, lightmap is currently blended with the base albedo map using the Overlay bend mode (here's a [description](https://helpx.adobe.com/photoshop/using/blending-modes.html) of the different blend modes).
* fixed a transparency issue that happens when a grayscale albedo map is used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,7 +33,9 @@ set(IGN_MATH_VER ${ignition-math6_VERSION_MAJOR})
 
 #--------------------------------------
 # Find ignition-common
-ign_find_package(ignition-common3 REQUIRED COMPONENTS graphics events)
+ign_find_package(ignition-common3 REQUIRED
+  COMPONENTS graphics events
+  VERSION 3.9)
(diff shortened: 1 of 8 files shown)


chapulina (approved):

chapulina:
@osrf-jenkins run tests please :pray:

chapulina:
Ah no, triggered too soon :woman_facepalming:

chapulina:
I fixed code check and a failing test on 1d69e73.

@iche033 , I removed a big chunk of code from the test which didn't seem to be used. Please check if that was indeed unnecessary and if it was ok to remove it.

iche033:
> I removed a big chunk of code from the test which didn't seem to be used. Please check if that was indeed unnecessary and if it was ok to remove it.

actually the test was supposed to test and verify copying from this `common::Material` but it was doing a `CopyFrom` from the wrong material variable. I've updated the test in 0d62019.

PR #186, opened 2020-12-16
Fix transparency issue for textures without alpha channel

iche033:
When `SetAlphaFromTexture(true)` is invoked on a material with texture that does not have an alpha channel, it causes the visual with the assigned material to be completely transparent. The fix is to do a check to make sure the texture does have an alpha channel, and if not we disable using alpha from the texture.

These changes were included in pull request #182 for ign-rendering5, which are necessary for displaying the `Indoor Lighting` model. 

Another test: launch ign-gazebo and insert the [DronePlatformX1](https://app.ignitionrobotics.org/OpenRobotics/fuel/models/DronePlatformX1) model. Without these changes, the model is transparent.


nkoenig (approved):
Works great.

iche033:
@osrf-jenkins run tests please

PR #193, opened 2021-01-01
ensure the proper library name is used to load the plugins

acxz:
Resolves https://github.com/ignitionrobotics/ign-rendering/issues/90#issuecomment-753255337

It may also resolve the issue as well, not sure though.

### DIFF
--- a/src/RenderEngineManager.cc
+++ b/src/RenderEngineManager.cc
@@ -409,7 +409,7 @@ void RenderEngineManagerPrivate::RegisterDefaultEngines()
   // TODO(anyone): Find a cleaner way to get the default engine library name
 
   // cppcheck-suppress unreadVariable
-  std::string libName = "ignition-rendering-";
+  std::string libName = "ignition-rendering4-";
 
   // cppcheck-suppress unusedVariable
   std::string engineName;


src/RenderEngineManager.cc
```diff
@@ -409,7 +409,7 @@ void RenderEngineManagerPrivate::RegisterDefaultEngines()
   // TODO(anyone): Find a cleaner way to get the default engine library name
 
   // cppcheck-suppress unreadVariable
-  std::string libName = "ignition-rendering-";
+  std::string libName = "ignition-rendering4-";
```
chapulina, line 412: This shouldn't be needed, because we're installing version-less symlinks to the versioned libraries. Maybe you're having trouble creating the symlinks or installing them?

chapulina (changes requested):

src/RenderEngineManager.cc
acxz: Ah yeah, upon closer inspection seems like there is a missing symlink. Adding the symlink does resolve the issue.

The installed libraries are as follows:
```
ignition-rendering /usr/lib/libignition-rendering4-ogre.so
ignition-rendering /usr/lib/libignition-rendering4-ogre.so.4
ignition-rendering /usr/lib/libignition-rendering4-ogre.so.4.1.0
ignition-rendering /usr/lib/libignition-rendering4.so
ignition-rendering /usr/lib/libignition-rendering4.so.4
ignition-rendering /usr/lib/libignition-rendering4.so.4.1.0
```
None of which are `/usr/lib/libignition-rendering-ogre.so`.
Running `simple_demo` after the following command:
```
sudo ln -s libignition-rendering4-ogre.so libignition-rendering-ogre.so
```
does work.

I'll look into this later, but it seems that I need to see how the symlinks are done in cmake.

iche033: there was a fix to install unversioned libs: ignition-release/ign-rendering4-release#3 in Nov. Maybe you just have to apt update and reinstall ignition-rendering?

iche033: just noticed in issue #90 you mentioned that you're on Arch linux. Not sure how ign-rendering is installed there but if it helps the relevant cmake line is here: https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/ogre2/src/CMakeLists.txt#L41

acxz: I see, i'm using the ogre1 backend and it also has the unversioned library being installed in cmake.

For some reason I can't seem to reproduce the issue now, so I'll close this PR. If I do come across this again, I'll know where to look. Thx for the tips!

PR #202, opened 2021-01-20
Added LightVisual

ahcorde:
This PR adds LightVisual. It's related with this issue https://github.com/ignitionrobotics/ign-gazebo/issues/193

Ogre2 is working fine:

Selección_090

But in Ogre I'm only able to see the LightVisual attached to an object.

Selección_091

When I run the simple_demo which includes some lights I'm able to see it in both cases:

- Ogre

Selección_092

- Ogre2
Selección_093

@iche033, any idea why in Gazebo the LightVisual is not being rendered ?


iche033:
> @iche033, any idea why in Gazebo the LightVisual is not being rendered ?

turns out to be an issue on the ign-gazebo side. We are not attaching the light to the scene. By default ogre2 attaches a light without parent to the root but ogre1.x does not and we have to manually attach it to the scene tree ourselves. Here's a patch to make lights appear in ign-gazebo:

```
diff --git a/src/rendering/SceneManager.cc b/src/rendering/SceneManager.cc
index 6aa73f78..1377def0 100644
--- a/src/rendering/SceneManager.cc
+++ b/src/rendering/SceneManager.cc
@@ -926,6 +926,8 @@ rendering::LightPtr SceneManager::CreateLight(Entity _id,
 
   if (parent)
     parent->AddChild(light);
+  else
+    this->dataPtr->scene->RootVisual()->AddChild(light);
 
   return light;
 }
```

iche033:
Can we move the code for creating the light visual to ign-gazebo gui plugin? So instead of always creating the `LightVisual` when a `Light` is created, we create them in ign-gazebo only when the users requests to visualize them. Similar to the idea of lidar visualization (and we can then also toggle their visibility on/off). These visualizations are gui only so we don't need to always create them.

include/ignition/rendering/LightVisual.hh
```diff
@@ -0,0 +1,82 @@
...
+namespace ignition
+{
+  namespace rendering
+  {
+    inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
+
+      /// \brief Enum for LightVisual types
+      enum IGNITION_RENDERING_VISIBLE LightVisualType
+      {
+        LightVisual_NONE                = 0,
```
iche033, line 34: add `/// \brief`

iche033: can you also change the naming convention of these enums to match others in ign-rendering,
 e.g.
https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/include/ignition/rendering/Marker.hh#L38

 so this would be something like `LVT_NONE`

include/ignition/rendering/LightVisual.hh
```diff
@@ -0,0 +1,82 @@
...
+        /// \brief Spot light
+        LightVisual_SPOT                = 3
+      };
+
+    /// \class LightVisual LightVisual.hh ignition/rendering/LightVisual.hh
+    /// \brief Represents a light visual
+    class IGNITION_RENDERING_VISIBLE LightVisual :
+      public virtual Visual
+    {
+      /// \brief Deconstructor
```
iche033, line 51: you probably got this from other files but: `Desconstructor` -> `Descructor`

ogre/src/OgreLightVisual.cc
```diff
@@ -0,0 +1,125 @@
...
+//////////////////////////////////////////////////
+OgreLightVisual::~OgreLightVisual()
+{
+}
+
+//////////////////////////////////////////////////
+void OgreLightVisual::PreRender()
+{
+  // if (this->dirtyLightVisual)
+  // {
```
iche033, line 49: remove these comments?

ogre2/src/Ogre2LightVisual.cc
```diff
@@ -0,0 +1,130 @@
...
+  this->dataPtr->lightVisual->SetOperationType(MarkerType::MT_LINE_LIST);
+  if (this->dataPtr->material == nullptr)
+  {
+    MaterialPtr defaultMat = this->Scene()->Material("Default/TransGreen");
+    this->SetMaterial(defaultMat, false);
+  }
+
+  std::vector<ignition::math::Vector3d> positions = this->CreateVisualLines();
+
+  for (auto p : positions)
```
iche033, line 90: `const auto &p`?

ogre/include/ignition/rendering/ogre/OgreLightVisual.hh
```diff
@@ -0,0 +1,79 @@
...
+
+    // Forward declaration
+    class OgreLightVisualPrivate;
+
+    class IGNITION_RENDERING_OGRE_VISIBLE OgreLightVisual :
+      public BaseLightVisual<OgreVisual>
+    {
+      protected: OgreLightVisual();
+
+      public: virtual ~OgreLightVisual();
```
iche033, line 47: add `/// \brief` for the constructor and descructor

include/ignition/rendering/LightVisual.hh
```diff
@@ -0,0 +1,82 @@
...
+      public: virtual double InnerAngle() = 0;
+
+      /// \brief set inner angle for spot lights
+      /// \param[in] _type inner angle
+      public: virtual void SetOuterAngle(double _outerAngle) = 0;
+
+      /// \brief Get inner angle
+      /// \return The light inner angle
+      public: virtual double OuterAngle() = 0;
+
```
iche033, line 77: remove space

src/LightVisual_TEST.cc
```diff
@@ -0,0 +1,82 @@
...
+
+  // create visual
+  LightVisualPtr lightVisual = scene->CreateLightVisual();
+  ASSERT_NE(nullptr, lightVisual);
+
+  // check initial values
+  EXPECT_EQ(LightVisualType::LightVisual_NONE, lightVisual->Type());
+
+  lightVisual->SetType(LightVisualType::LightVisual_POINT);
+  EXPECT_EQ(LightVisualType::LightVisual_POINT, lightVisual->Type());
```
iche033, line 61: can you also test the inner and outer angle functions?

ogre2/src/Ogre2LightVisual.cc
```diff
@@ -0,0 +1,130 @@
...
+
+using namespace ignition;
+using namespace rendering;
+
+class ignition::rendering::Ogre2LightVisualPrivate
+{
+  /// \brief Grid materal
+  public: Ogre2MaterialPtr material = nullptr;
+
+  /// \brief Ogre renderable used to render the grid.
```
iche033, line 29: update comment here and above

ogre/src/OgreLightVisual.cc
```diff
@@ -0,0 +1,125 @@
...
+
+//////////////////////////////////////////////////
+void OgreLightVisual::CreateVisual()
+{
+  if (!this->dataPtr->line)
+  {
+    this->dataPtr->line = std::shared_ptr<OgreDynamicLines>(
+      new OgreDynamicLines(MT_LINE_LIST));
+    this->ogreNode->attachObject(this->OgreObject());
+    this->dataPtr->line->setMaterial("Default/TransRed");
```
iche033, line 78: keep this the same color as ogre2 and gazebo classic? ie `Default/TransGreen`

ogre/src/OgreLightVisual.cc
```diff
@@ -0,0 +1,125 @@
...
+  }
+  std::vector<ignition::math::Vector3d> positions = this->CreateVisualLines();
+
+  for (auto p : positions)
```
iche033, line 82: `const auto &p`?

include/ignition/rendering/LightVisual.hh
ahcorde: There is a conflict with LidarVisualType. Any recomendations ?

include/ignition/rendering/LightVisual.hh
ahcorde: Fix to the other files https://github.com/ignitionrobotics/ign-rendering/pull/220

include/ignition/rendering/LightVisual.hh
iche033: ah that's unfortunate, how about `LVT_EMPTY`?

ahcorde: @iche033 Resolved! https://github.com/ignitionrobotics/ign-rendering/pull/202/commits/69232c03650aa1784a35b6b75a134706610b11ad

iche033 (approved):
I added some doc in a4f25d0285778f3a77813da13614f1479b61fc34. Changes look good. I think windows build failure is an issue with the CI machine

peci1:
@ahcorde can you check spotlights do not suffer from the same bug as Gazebo Classic doubling their FOV? https://github.com/osrf/gazebo/pull/2947 . Looking at the code it seems to me the same bug was copied...

ahcorde:
@peci1 FYI: PR https://github.com/ignitionrobotics/ign-rendering/pull/308

PR #203, opened 2021-01-20
Added Ellipsoid Geometry

ahcorde:
Added Ellipsoid geometry. Related with this issue https://github.com/ignitionrobotics/ign-rendering/issues/198

Ogre 2

Selección_095

Ogre:

Selección_094


ahcorde:
Depends on https://github.com/ignitionrobotics/ign-common/pull/154

ogre2/src/Ogre2Scene.cc
```diff
@@ -361,6 +362,14 @@ MeshPtr Ogre2Scene::CreateMeshImpl(unsigned int _id, const std::string &_name,
   return this->CreateMeshImpl(_id, _name, descriptor);
 }
 
+//////////////////////////////////////////////////
+EllipsoidPtr Ogre2Scene::CreateEllipsoidImpl(unsigned int _id,
+    const std::string &_name)
+{
+  Ogre2EllipsoidPtr ellipsoid(new Ogre2Ellipsoid);
```
chapulina, line 369: I'm curious, why did you create an `Ogre2Ellipsoid` class instead of following the same pattern as sphere, cylinder, box, plane, and using `CreateMeshImpl`?

ahcorde: I created an implementation without having a look how the basic shapes were done. I can remove the code in both shapes (capsule and ellipsoid) probably is going to be easy to maintain.

ahcorde: @iche033 any thoughts?

iche033: +1 on using `CreateMeshImpl`. We could add a `unit_ellipsoid` in ign-common and make use of that here in ign-rendering.

ahcorde: Update https://github.com/ignitionrobotics/ign-rendering/pull/203/commits/61644e4163a87341bdc308d6b8288e13ac1fbff2

ogre/src/OgreScene.cc
```diff
@@ -502,6 +502,13 @@ MeshPtr OgreScene::CreateMeshImpl(unsigned int _id, const std::string &_name,
   return this->CreateMeshImpl(_id, _name, descriptor);
 }
 
+//////////////////////////////////////////////////
+MeshPtr OgreScene::CreateEllipsoidImpl(
+  unsigned int _id, const std::string &_name)
+{
+  return this->CreateMeshImpl(_id, _name, "unit_ellipsoid");
```
chapulina, line 509: As discussed in https://github.com/ignitionrobotics/ign-common/pull/165, this could just be `unit_sphere`

examples/simple_demo/Main.cc
```diff
@@ -118,6 +118,15 @@ void buildScene(ScenePtr _scene)
   box->SetMaterial(blue);
   root->AddChild(box);
 
+  // create ellipsoid visual
+  VisualPtr ellipsoidVisual = _scene->CreateVisual();
+  auto ellipsoid = _scene->CreateEllipsoid();
+  ellipsoid->SetRadii(ignition::math::Vector3d(1.2, 0.7, 0.5));
```
chapulina, line 124: This example needs to be updated now that there's no ellipsoid class. The ellipsoid's size will be defined using `SetLocalScale`, the same way as the sphere above.

Which brings the question of whether special API is needed for the ellipsoid at all. Despite its name, `ign-rendering`'s "sphere" already behaves like an "ellipsoid", because it can be scaled in 3 dimensions.

At this point, my suggestion would be to:
* remove all the ellipsoid logic added in this PR 
* update `CreateSphere`'s documentation to say that it can be used to create ellipsoids
* update this example to create an "ellipsoid" through the `CreateSphere` function

chapulina (changes requested):

examples/simple_demo/Main.cc
ahcorde: Thank you for the bullet points :construction_worker:

chapulina (approved):
Thanks for iterating! The demo is very suspenseful :laughing: I'm staring asking "where's the green ellipsoid :eyes: :sweat_smile: 

ellipsoid

PR #204, opened 2021-01-20
Call XCloseDisplay in screenScalingFactor

brawner:
I clicked too much in gazebo, and the gui crashed :(

This function is called a bunch when Transform Control is used. Because the display was not cleaned up, eventually `XOpenDisplay` will throw an exception about too many connections. 

Signed-off-by: Stephen Brawner <brawner@gmail.com>


iche033 (approved):
looks good to me.

brawner:
Failures in ign_rendering-pr-win and ignition_rendering-ci-pr seem unrelated.

PR #206, opened 2021-01-22
👩‍🌾 Clear all Windows warnings

chapulina:
It was about time! 8k warnings gone!

The warnings follow these general categories:

* Warnings from Ogre 1 / 2 source code
* C4251: the good old "missing DLL interfaces". I started suppressing them one by one but eventually gave in to the blanket suppression in `CMakeLists`. This alone corresponds to **95%** of all warnings!
* Type conversion warnings
* A couple of _discarding return value of function with 'nodiscard' attribute_ - one of which was actually a bug (`this->ogreMaterial == nullptr;)`
* C4275: :warning: I'm not sure that suppressing this one is a final solution, I think that `Ogre::MaterialManager::Listener` should be in the dll interface like all other classes we're inheriting from.
* `inconsistent dll linkage` for `BaseDepthCamera` - removed the visibility since it's a header-only class and shouldn't need it
* :warning: : the `OgreSelectionBufferPrivate` class was declared as `struct` but defined as `class`. I'm not sure if fixing this doesn't break ABI.
* There were a few GCC pragmas that Windows could still see and didn't like them
* Windows wasn't liking some of our preprocessor directives (_unexpected tokens following preprocessor directive - expected a newline_), so I changed them until it was satisfied

Unfortunately, CI isn't blue yet because of #201 :cry: I added some nullptr checks to see if that would be improved with no luck...

Test build before PR with no warnings: https://build.osrfoundation.org/job/ign_rendering-pr-win/1175/

---

https://github.com/osrf/buildfarmer/issues/145

### DIFF
--- a/ogre/src/CMakeLists.txt
+++ b/ogre/src/CMakeLists.txt
@@ -2,6 +2,15 @@
 # "gtest_sources" variable.
 ign_get_libsources_and_unittests(sources gtest_sources)
 
+if (MSVC)
+  # Warning #4251 is the "dll-interface" warning that tells you when types used
+  # by a class are not being exported. These generated source files have private
+  # members that don't get exported, so they trigger this warning. However, the
+  # warning is not important since those members do not need to be interfaced
+  # with.
+  set_source_files_properties(${sources} ${gtest_sources} COMPILE_FLAGS "/wd4251")
+endif()
+
 set(engine_name "ogre")
 
 ign_add_component(${engine_name} SOURCES ${sources} GET_TARGET_NAME ogre_target)
@@ -54,7 +63,7 @@ if (WIN32)
   # disable MSVC inherit via dominance warning
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4250")
   INSTALL(CODE "EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E copy
-  ${IGNITION_RENDERING_ENGINE_INSTALL_DIR}\/${versioned} 
+  ${IGNITION_RENDERING_ENGINE_INSTALL_DIR}\/${versioned}
   ${IGNITION_RENDERING_ENGINE_INSTALL_DIR}\/${unversioned})")
 else()
   EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E create_symlink ${versioned} ${unversioned})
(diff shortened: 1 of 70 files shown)


chapulina:
Oh look at that, a :heavy_check_mark: on Windows :scream:

ogre/src/OgreText.cc
```diff
@@ -380,7 +380,7 @@ void OgreMovableText::SetFontNameImpl(const std::string &_newFontName)
 #if OGRE_VERSION_LT_1_10_1
       this->ogreMaterial.setNull();
 #else
-      this->ogreMaterial == nullptr;
+      this->ogreMaterial = nullptr;
```
j-rivero, line 383: uh .. this one was important

j-rivero (approved):
Holy cow @chapulina , I take my hat off :tophat:

chapulina:
> At least mentioning it in the weekly meeting could be a good idea.

Cool, I added an item for us not to forget :tada:

ogre/src/CMakeLists.txt
```diff
@@ -28,10 +37,10 @@ set_property(
 target_link_libraries(${ogre_target}
   PUBLIC
     ${ignition-common${IGN_COMMON_VER}_LIBRARIES}
+    IgnOGRE::IgnOGRE
```
chapulina, line 40: I think this broke downstream libraries like `ign-gui`:

```
CMake Error at src/plugins/CMakeLists.txt:29 (add_library):
  Target "Grid3D" links to target "IgnOGRE::IgnOGRE" but the target was not               
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?       
Call Stack (most recent call first): 
  src/plugins/CMakeLists.txt:77 (ign_gui_add_library)
  src/plugins/grid_3d/CMakeLists.txt:1 (ign_gui_add_plugin)
                                    
                                     
CMake Error at src/plugins/CMakeLists.txt:29 (add_library):
  Target "Scene3D" links to target "IgnOGRE::IgnOGRE" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?            
Call Stack (most recent call first):  
  src/plugins/CMakeLists.txt:77 (ign_gui_add_library)
  src/plugins/scene3d/CMakeLists.txt:1 (ign_gui_add_plugin)  
```

PR #211, opened 2021-01-23
Add temperature variations to thermal camera readings based on color

iche033:
The thermal camera class has functions to set heat source and ambient temperature range. This was simulated by varying an object's temperature based on its depth value (distance from the camera). This PR provides an alternative way to vary an object's temperature based on its color value.

Non-heat source objects, i.e visuals without a uniform temperature or heat signature, will now have variations in their thermal readings as a function of color - the darker the object the warmer the object. See https://github.com/ignitionrobotics/ign-gazebo/pull/578 for an example output

Note: This is not how a physical thermal camera works. The goal is to provide automated way of simulating enough temperature differences so that thermal data processing algorithms are able to segment out different objects in the scene.

### DIFF
--- a/ogre2/include/ignition/rendering/ogre2/Ogre2Material.hh
+++ b/ogre2/include/ignition/rendering/ogre2/Ogre2Material.hh
@@ -199,6 +199,12 @@ namespace ignition
       // Documentation inherited
       public: virtual void SetDepthWriteEnabled(bool _enabled) override;
 
+      /// \brief Helper function to convert an ogre Pbs datablck to Unlit datablock
+      /// \param[in] _in Input pbs datablock
+      /// \param[out] _out Output unlit datablock
+      public: static void PbsToUnlitDatablock(Ogre::HlmsPbsDatablock *_in,
+          Ogre::HlmsUnlitDatablock *_out);
+
       /// \brief Set the texture map for this material
       /// \param[in] _texture Name of the texture.
       /// \param[in] _type Type of texture, i.e. diffuse, normal, roughness,
--- a/ogre2/src/Ogre2ThermalCamera.cc
+++ b/ogre2/src/Ogre2ThermalCamera.cc
@@ -28,6 +28,9 @@
 #include <unordered_map>
 #include <variant>
 
+#include <Hlms/Pbs/OgreHlmsPbsDatablock.h>
+#include <Hlms/Unlit/OgreHlmsUnlitDatablock.h>
+
 #include <ignition/common/Console.hh>
 #include <ignition/common/Filesystem.hh>
 #include <ignition/math/Helpers.hh>
@@ -35,6 +38,7 @@
 #include "ignition/rendering/RenderTypes.hh"
 #include "ignition/rendering/ogre2/Ogre2Conversions.hh"
 #include "ignition/rendering/ogre2/Ogre2Includes.hh"
+#include "ignition/rendering/ogre2/Ogre2Material.hh"
 #include "ignition/rendering/ogre2/Ogre2RenderEngine.hh"
 #include "ignition/rendering/ogre2/Ogre2RenderTarget.hh"
 #include "ignition/rendering/ogre2/Ogre2RenderTypes.hh"
@@ -95,6 +99,9 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \brief The name of the thermal camera sensor
   private: const std::string name;
 
+  /// \brief The thermal camera
+  private: const Ogre::Camera* ogreCamera{nullptr};
+
   /// \brief Custom parameter index of temperature data in an ogre subitem.
   /// This has to match the custom index specifed in ThermalHeatSource material
   /// script in media/materials/scripts/thermal_camera.material
@@ -102,6 +109,9 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
 
   /// \brief A map of ogre sub item pointer to their original hlms material
   private: std::map<Ogre::SubItem *, Ogre::HlmsDatablock *> datablockMap;
+
+  /// \brief A map of ogre sub item pointer to their unlit hlms material
+  private: std::map<Ogre::SubItem *, Ogre::HlmsDatablock *> unlitDatablockMap;
 };
 }
 }
@@ -171,6 +181,8 @@ Ogre2ThermalCameraMaterialSwitcher::Ogre2ThermalCameraMaterialSwitcher(
 
   this->baseHeatSigMaterial = Ogre::MaterialManager::getSingleton().
     getByName("ThermalHeatSignature");
+
+  this->ogreCamera = this->scene->OgreSceneManager()->findCamera(this->name);
 }
 
 //////////////////////////////////////////////////
@@ -238,7 +250,7 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
         if (temp >= 0.0)
         {
           // set visibility flag so thermal camera can see it
-          item->addVisibilityFlags(0x10000000);
+          // item->addVisibilityFlags(0x10000000);
           for (unsigned int i = 0; i < item->getNumSubItems(); ++i)
           {
             Ogre::SubItem *subItem = item->getSubItem(i);
@@ -248,7 +260,7 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
               float color = temp * 100.0 /
                   static_cast<float>(std::numeric_limits<uint16_t>::max());
               subItem->setCustomParameter(this->customParamIdx,
-                  Ogre::Vector4(color, color, color, 1.0));
+                  Ogre::Vector4(color, 0, 0, 0.0));
             }
             Ogre::HlmsDatablock *datablock = subItem->getDatablock();
             this->datablockMap[subItem] = datablock;
@@ -340,9 +352,6 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
           this->heatSignatureMaterials[item->getId()] = heatSignatureMaterial;
         }
 
-        // set visibility flag so thermal camera can see it
-        item->addVisibilityFlags(0x10000000);
-
         for (unsigned int i = 0; i < item->getNumSubItems(); ++i)
         {
           Ogre::SubItem *subItem = item->getSubItem(i);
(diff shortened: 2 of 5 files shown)


ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -353,6 +368,43 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
+          Ogre2MaterialPtr ogreMat = std::dynamic_pointer_cast<Ogre2Material>(mat);
+          Ogre::HlmsUnlitDatablock *unlit = ogreMat->UnlitDatablock();
+
+          for (unsigned int i = 0; i < item->getNumSubItems(); ++i)
+          {
+            Ogre::SubItem *subItem = item->getSubItem(i);
+            auto it = this->unlitDatablockMap.find(subItem);
+            if (it == this->unlitDatablockMap.end())
+            {
+              this->unlitDatablockMap[subItem] = unlit;
```
adlarkin, line 394: Can you help me understand the purpose of this assignment? I don't see `unlitDatablockMap` being accessed anywhere else. Is `unlitDatablockMap` by chance a testing/outdated variable that can be removed?

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -353,6 +368,43 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
+            Ogre::HlmsDatablock *datablock = subItem->getDatablock();
+            this->datablockMap[subItem] = datablock;
+            subItem->setDatablock(unlit);
+          }
+
+          // make sure background objects are not visible in depth texture
+          // we are using this info to determine what pixels are background
+          // background objects vs heat source
+          // auto flags = item->getVisibilityFlags();
+          // item->setVisibilityFlags(flags & ~0x01000000);
```
adlarkin, line 405: Can these lines be deleted?

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -362,8 +414,8 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
 void Ogre2ThermalCameraMaterialSwitcher::postRenderTargetUpdate(
     const Ogre::RenderTargetEvent & /*_evt*/)
 {
-  // restore item to use hlms material
-  for (auto it : this->datablockMap)
+  // restore item to use pbs hlms material
+  for (auto &it : this->datablockMap)
```
adlarkin, line 418: Is there a reason why we are now taking a reference to the iterator? Based on my understanding, `datablockMap` is a map of pointers, so I don't think there'd be any issues related to using copies vs references.

ogre2/src/media/materials/programs/heat_signature_fs.glsl
```diff
@@ -46,5 +46,8 @@ float mapNormalized(float num)
 void main()
 {
   float heat = texture(RT, inPs.uv0.xy).x;
-  fragColor = vec4(mapNormalized(heat), 0, 0, 1.0);
+
+  // set g, b, a to 0. This will be used by thermal_camera_fs.glsl to determine
+  // if a particular fragment is a heat source or not
+  fragColor = vec4(mapNormalized(heat), 0, 0, 0.0);
```
adlarkin, line 52: Based on my understanding, objects with a heat signature use `heat_signature_fs.glsl` instead of `thermal_camera_fs.glsl`: https://github.com/ignitionrobotics/ign-rendering/blob/c6ee8b9a772db37a27edab361a176ecee80c417e/ogre2/src/Ogre2ThermalCamera.cc#L172-L190

So, if my understanding is correct, can you help me understand how `thermal_camera_fs.glsl` will determine if the output from `heat_signature_fs.glsl` is a heat source or not? I'm having a hard time understanding how these two shader files share information with one another (they seem independent to me).

ogre2/src/media/materials/programs/thermal_camera_fs.glsl
```diff
@@ -51,28 +52,62 @@ void main()
...
 
-    // set temperature variation for heat source
-    heatRange = heatSourceTempRange;
+  // add temperature variation, either as a function of color or depth
+  if (rgbToTemp == 1)
+  {
+    if (heat > 0.0 && !isHeatSource)
+    {
+      // convert to grayscale: darker = warmer
+      float gray = rgba.r * 0.3 + rgba.g * 0.59 + rgba.b*0.11;
```
adlarkin, line 95: Would it be worth adding a comment about why these particular weights were chosen? I know that there are various RGB -> grayscale conversion formulas out there that use different weights.

ogre2/src/media/materials/programs/heat_signature_fs.glsl
adlarkin: It was a misunderstanding on my end (I didn't realize that one shader is called earlier in the pipeline than the other, but it makes sense when you see how one is called in pre render target update, and the other in post), this looks good to me :+1:

ogre2/include/ignition/rendering/ogre2/Ogre2Material.hh
```diff
@@ -199,6 +199,12 @@ namespace ignition
       // Documentation inherited
       public: virtual void SetDepthWriteEnabled(bool _enabled) override;
 
+      /// \brief Helper function to convert an ogre Pbs datablck to Unlit datablock
+      /// \param[in] _in Input pbs datablock
+      /// \param[out] _out Output unlit datablock
+      public: static void PbsToUnlitDatablock(Ogre::HlmsPbsDatablock *_in,
+          Ogre::HlmsUnlitDatablock *_out);
```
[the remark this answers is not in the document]
iche033: Adding a new static function should be fine, see a list of things allowed [here](https://www.ros.org/reps/rep-0009.html)

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -150,6 +160,10 @@ class ignition::rendering::Ogre2ThermalCameraPrivate
   /// \brief Pointer to material switcher
   public: std::unique_ptr<Ogre2ThermalCameraMaterialSwitcher>
       thermalMaterialSwitcher = nullptr;
+
+  /// \brief Add variation to temperature values based on object rgb values
+  /// This only affects objects that are not heat sources
+  public: bool rgbToTemp = true;
```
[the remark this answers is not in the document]
iche033: I was going to add a API for setting this value but realized that could break ign-rendering ABI so I didn't do it. It can remain const for now but we should consider making this configurable in edifice.

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -353,6 +368,43 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
+        Ogre::Aabb aabb = item->getWorldAabbUpdated();
+        Ogre::AxisAlignedBox box = Ogre::AxisAlignedBox(aabb.getMinimum(), aabb.getMaximum());
+
+        // we will be converting rgb values to tempearture values in shaders
+        // but we want to make sure the object rgb values are not affected by lighting
+        // so disable lighting
+        // Also check if objects are within camera view
+        if (ogreVisual->GeometryCount() > 0u && this->ogreCamera->isVisible(box))
+        {
+          auto geom = ogreVisual->GeometryByIndex(0);
```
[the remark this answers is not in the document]
iche033: attached geometries always start from index 0. Since there is a check above for `ogreVisual->GeometryCount() > 0`, there should be a valid geometry here. I can also add an extra check for null ptr to be sure.

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -353,6 +368,43 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
+          MaterialPtr mat = geom->Material();
+          Ogre2MaterialPtr ogreMat = std::dynamic_pointer_cast<Ogre2Material>(mat);
+          Ogre::HlmsUnlitDatablock *unlit = ogreMat->UnlitDatablock();
+
+          for (unsigned int i = 0; i < item->getNumSubItems(); ++i)
+          {
+            Ogre::SubItem *subItem = item->getSubItem(i);
+            auto it = this->unlitDatablockMap.find(subItem);
+            if (it == this->unlitDatablockMap.end())
```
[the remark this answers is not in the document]
iche033: yes I can change them to unordered map

ogre2/src/Ogre2ThermalCamera.cc
iche033: you're right, we don't need to store it. I'll remove it.

ogre2/src/Ogre2ThermalCamera.cc
iche033: I think I did that out of habit when I'm not sure what the type is. Since it's a pointer, we shouldn't need to do it. I'll revert this change.

ogre2/src/media/materials/programs/thermal_camera_fs.glsl
iche033: I rounded the coefficients from here: https://docs.opencv.org/3.4/de/d25/imgproc_color_conversions.html

I noticed there are other ones for gamma sRGB image space here:
https://en.wikipedia.org/wiki/Grayscale

ogre2/src/Ogre2ThermalCamera.cc
iche033: 397a443

ogre2/src/Ogre2ThermalCamera.cc
iche033: removed. 397a443

ogre2/src/Ogre2ThermalCamera.cc
iche033: removed. 397a443

ogre2/src/Ogre2ThermalCamera.cc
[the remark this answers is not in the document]
iche033: done. 397a443

ogre2/src/Ogre2ThermalCamera.cc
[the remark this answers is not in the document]
iche033: added todo note 397a443

ogre2/src/Ogre2ThermalCamera.cc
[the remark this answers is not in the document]
iche033: added check 397a443

Issue #170, opened 2020-10-28
Camera/CameraTest.VisualAt/ogre2 flaky test on Bionic/amd64


j-rivero:
I've seen the `Camera/CameraTest.VisualAt/ogre2` failing in [some](https://build.osrfoundation.org/job/ignition_rendering-ci-ign-rendering3-bionic-amd64/21/testReport/) [builds](https://build.osrfoundation.org/job/ignition_rendering-ci-ign-rendering3-bionic-amd64/25/) for Bionic/amd64.
[Changes listed in build 20](https://build.osrfoundation.org/job/ignition_rendering-ci-ign-rendering3-bionic-amd64/20/)seems to be the starting point of the failures.

```
54: [ RUN      ] Camera/CameraTest.VisualAt/ogre2
54: [1;32m[Msg] [0m[1;32mLoading plugin [[0m[1;32mignition-rendering-ogre2[0m[1;32m][0m[1;32m[0m
54: /var/lib/jenkins/workspace/ignition_rendering-ci-ign-rendering3-bionic-amd64/ign-rendering/test/integration/camera.cc:243: Failure
54: Expected: (nullptr) != (vis), actual: (nullptr) vs 16-byte object <00-00 00-00 00-00 00-00 00-00 00-00 00-00 00-00>
54: [  FAILED  ] Camera/CameraTest.VisualAt/ogre2, where GetParam() = "ogre2" (224 ms)
```

chapulina:
The test is flaky on Citadel:

https://build.osrfoundation.org/job/ignition_rendering-ci-ign-rendering3-bionic-amd64/27/testReport/(root)/Camera_CameraTest/VisualAt_ogre2/history/

But it hasn't failed in Dome yet:

https://build.osrfoundation.org/job/ignition_rendering-ci-ign-rendering4-bionic-amd64/1/testReport/(root)/Camera_CameraTest/VisualAt_ogre2/history/

chapulina:
On #174, I'm adding more information to the test so we can understand better how it's failing.

chapulina:
> But it hasn't failed in Dome yet:

Now that there are more builds, it's possible to see that the test is flaky on Dome too. And on Edifice:

https://build.osrfoundation.org/job/ignition_rendering-ci-main-bionic-amd64/lastStableBuild/testReport/(root)/Camera_CameraTest/VisualAt_ogre2/history/

#255 added a bit more debug information

chapulina:
[...]

PR #221, opened 2021-01-29
Fixed transform control example

ahcorde:
I found an issue in the rendering examples, in particular in the `transform_control` example. When I try to click in the arrows or in one of the other Gizmo visuals is not getting the visual. I need to click in other part of the screen. See the gif below.

transform_bad

With this fix I'm able to get the right visual:

transform_ok

This is also happening in ign-gazebo.

### DIFF
--- a/include/ignition/rendering/Utils.hh
+++ b/include/ignition/rendering/Utils.hh
@@ -37,7 +37,6 @@ namespace ignition
     /// \brief Get the screen scaling factor.
     /// \param[out] _xScale The X screen scaling factor.
     /// \param[out] _yScale The Y screen scaling factor.
-    /// \return The screen scaling factor.
     IGNITION_RENDERING_VISIBLE
     void screenScalingFactor(float &_xScale, float &_yScale);
     }
--- a/include/ignition/rendering/Utils.hh
+++ b/include/ignition/rendering/Utils.hh
@@ -34,6 +34,12 @@ namespace ignition
     /// \return The screen scaling factor.
     IGNITION_RENDERING_VISIBLE
     float screenScalingFactor();
+    /// \brief Get the screen scaling factor.
+    /// \param[out] _xScale The X screen scaling factor.
+    /// \param[out] _yScale The Y screen scaling factor.
+    /// \return The screen scaling factor.
+    IGNITION_RENDERING_VISIBLE
+    void screenScalingFactor(float &_xScale, float &_yScale);
     }
   }
 }
--- a/ogre/src/OgreCamera.cc
+++ b/ogre/src/OgreCamera.cc
@@ -216,10 +216,11 @@ VisualPtr OgreCamera::VisualAt(const ignition::math::Vector2i
     }
   }
 
-  float ratio = screenScalingFactor();
+  float xScale, YScale;
+  screenScalingFactor(xScale, YScale);
   ignition::math::Vector2i mousePos(
-      static_cast<int>(std::rint(ratio * _mousePos.X())),
-      static_cast<int>(std::rint(ratio * _mousePos.Y())));
+      static_cast<int>(std::rint(xScale * _mousePos.X())),
+      static_cast<int>(std::rint(YScale * _mousePos.Y())));
 
   Ogre::Entity *entity = this->selectionBuffer->OnSelectionClick(
       mousePos.X(), mousePos.Y());
--- a/ogre/src/OgreCamera.cc
+++ b/ogre/src/OgreCamera.cc
@@ -216,11 +216,11 @@ VisualPtr OgreCamera::VisualAt(const ignition::math::Vector2i
     }
   }
 
-  float xScale, YScale;
-  screenScalingFactor(xScale, YScale);
+  float xScale, yScale;
+  screenScalingFactor(xScale, yScale);
   ignition::math::Vector2i mousePos(
       static_cast<int>(std::rint(xScale * _mousePos.X())),
-      static_cast<int>(std::rint(YScale * _mousePos.Y())));
+      static_cast<int>(std::rint(yScale * _mousePos.Y())));
 
   Ogre::Entity *entity = this->selectionBuffer->OnSelectionClick(
       mousePos.X(), mousePos.Y());
--- a/src/Utils.cc
+++ b/src/Utils.cc
@@ -31,61 +31,20 @@ inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
 /////////////////////////////////////////////////
 float screenScalingFactor()
 {
-  // todo(anyone) set device pixel ratio for high dpi displays on Windows
-  float ratio = 1.0;
-#ifdef __linux__
-  Display *disp = XOpenDisplay(nullptr);
-  char *resourceString = XResourceManagerString(disp);
-
-  if (resourceString)
-  {
-    char *type = nullptr;
-    float dpiDesktop = 0.0;
-
-    // Need to initialize the DB before calling Xrm* functions
-    XrmInitialize();
-
-    XrmValue value;
-    XrmDatabase db = XrmGetStringDatabase(resourceString);
-
-    // Debug:
-    // printf("Entire DB:\n%s\n", resourceString);
-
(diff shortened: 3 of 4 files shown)


iche033:
could be related to https://github.com/ignitionrobotics/ign-gazebo/issues/147

ahcorde:
@iche033 I completely changed the PR, do you mind to have another look ?

iche033:
interesting, so you have different xScale and yScale values?

iche033:
The changes look good to me. I'm not able to reproduce this issue on my laptop though.

@jennuine do you still have this issue? If so, do you mind testing to see if this fixes the problem?

ahcorde:
> interesting, so you have different xScale and yScale values?

Yes, I have xScale = 80 and yScale 81,5.

But I think the main issue was that if the scale is lower than the DPI then no scale is needed

jennuine:
I still have this issue and unfortunately the changes did not fix the problem for me

ahcorde:
Thank you for testing it @jennuine, can you remove this comment ?

https://github.com/ignitionrobotics/ign-rendering/blob/53e850f0f71e9521fcdc59bec492baa437710c9f/src/Utils.cc#L151-L153

And post the output here.

jennuine:
```
DPI Desktop: 96.000000, DPI XY: [162.019928, 160.891495], Scale XY: [1.687708, 1.675953]
```

peci1:
I think the code you're searching for Windows is:

    #include <windows.h>
    #include <GL/gl.h>
    // windows.h defines these macros which break other parts of the library
    #undef max
    #undef min

     float xDpiDesktop = GetDeviceCaps(wglGetCurrentDC(), LOGPIXELSX);
     float xPixels = GetDeviceCaps(wglGetCurrentDC(), HORZRES);
     float xMM = GetDeviceCaps(wglGetCurrentDC(), HORZSIZE);
     float yDpiDesktop = GetDeviceCaps(wglGetCurrentDC(), LOGPIXELSY);
     float yPixels = GetDeviceCaps(wglGetCurrentDC(), VERTRES);
     float yMM = GetDeviceCaps(wglGetCurrentDC(), VERTSIZE);

I did a test on my computer with screen scaling set to 125%, with a FullHD 21,5" display:

    xDpiDesktop = 120
    xPixels = 1920
    xMM = 477

I think it corresponds to the values you're trying to get in X11.

However, the transform control example works for me even without this patch on Windows.

Docs here: https://docs.microsoft.com/en-us/windows/win32/api/wingdi/nf-wingdi-getdevicecaps . The GetDeviceCaps method returns some "generic" DPI common to all screens, but I think it might be better than nothing.

You can also use `GetDpiForWindow(hwnd)` which gets you DPI for a specific window, which might be better. But I haven't found a good way to access the window handle of the OGRE window. There is [`Ogre::RenderTarget::getCustomAttribute("WINDOW", &hwnd)`](https://ogrecave.github.io/ogre/api/1.11/class_ogre_1_1_render_window.html#a6c32524efacf6da599d5daccf3d9e166), but it would need some more logic for different render engines and OSes. However, it might be useful even in case of X11 because that would return you a handle to the actual window and its screen, and not always screen 0.

Also, the current implementation of the added `screenScalingFactor(xScale, yScale)` function should default to returning ones for unsupported operating systems, not "nothing" as it does now.

include/ignition/rendering/Utils.hh
```diff
@@ -40,6 +40,13 @@ namespace ignition
     IGNITION_RENDERING_VISIBLE
     float screenScalingFactor();
 
+    /// \brief Get the screen scaling factor.
+    /// \param[out] _xScale The X screen scaling factor.
+    /// \param[out] _yScale The Y screen scaling factor.
+    /// \return The screen scaling factor.
```
chapulina, line 46: Return is `void` :wink:

src/Utils.cc
```diff
@@ -89,6 +89,72 @@ float screenScalingFactor()
   return ratio;
 }
 
+/////////////////////////////////////////////////
+void screenScalingFactor(float &_xScale, float &_yScale)
+{
+  // todo(anyone) set device pixel ratio for high dpi displays on Windows
+#ifdef __linux__
```
chapulina, line 96: Should we set the ratio to 1.0 for non-linux like the other `screenScalingFactor` function does?

ogre/src/OgreCamera.cc
```diff
@@ -216,10 +216,11 @@ VisualPtr OgreCamera::VisualAt(const ignition::math::Vector2i
     }
   }
 
-  float ratio = screenScalingFactor();
+  float xScale, YScale;
+  screenScalingFactor(xScale, YScale);
```
chapulina, line 220: Can we cache this value instead of computing it every time the user clicks the screen? I assume this wouldn't change unless the user changes some display settings, right?

chapulina (commented):
Another reference point:

* Ubuntu Bionic Docker container (host is Ubuntu Focal)
* Display resolution: 2560x1600, scale: 100%
* Output: `DPI Desktop: 96.000000, DPI XY: [100.968941, 101.094528], Scale XY: [1.051760, 1.053068]`

transform

ogre/src/OgreCamera.cc
peci1: The user could also move the window to a different screen... I think it is choosing between a little runtime overhead or very complex event-reacting code `ifdef`d for each OS or display system... But maybe I'm wrong :)

chapulina:
I believe this issue has been manifesting on CI, see #170. The `VisualAt` test has been "flaky", but actually it consistently passes / fails according to which machine it runs on.

chapulina:
I pushed some tweaks in 58a0857:

* Remove duplicate code
* Break down the math to make it easier to debug

---

I also tracked down the history of the scaling factor.

* It was introduced in [this BitBucket PR](https://osrf-migration.github.io/ignition-gh-pages/#!/ignitionrobotics/ign-rendering/pull-requests/166/page/1) to support high DPI displays. Before that, the scaling was always 1.0.
* Gazebo classic has [APIs to set the scaling factor](https://github.com/osrf/gazebo/blob/1c58a0a9f040eedb85ae6d9418a36b2df10776e6/gazebo/rendering/UserCamera.cc#L636-L646), but that's never called internally, so it just uses the default 1.0.

With that in mind, I tried setting the scaling to 1.0 and voilà, it worked like a charm.

I don't really understand why the resolution we get from the calculation is different from the one reported by `Xrm` and how this scaling changes for different hardware, so it's hard for me to make a recommendation here. The fact that Gazebo classic has been using 1.0 so far is a strong indication for me that reverting to 1.0 may be a good option.

---

So I see 3 different setups in this thread:

* @ahcorde 's is fixed with the current PR
* @jennuine 's doesn't work with or without the PR
* @chapulina 's doesn't work with or without the PR, but works with `1.0`

@jennuine , @ahcorde : can you try 1.0 / 1.0 scaling just for fun?

I'd also be interested in seeing if @nkoenig could test the laptop that gave origin to the scaling in the first place.

---

https://github.com/osrf/buildfarmer/issues/207

ahcorde:
it's working for me, 

I retarget this to `ign-rendering3`, because `ign-rendering4` it's no Citadel or Edifice.

jennuine:
It's working for me now with or without the PR now... With the PR, I tested both as is and with scaling 1.0 and it works either way. Without the PR, I've tested this on Citadel, Dome, and Edifice and all work as expected now.

:thinking: I don't know what could have changed between now and then, possibly a system update? I'm using Ubuntu Focal. Maybe @mabelzhang can give it a try, I believe she had come across this issue awhile ago too.

chapulina:
> it's working for me

@ahcorde , you mean with this PR or with 1.0 or both?

ahcorde:
I only tried 1.0

chapulina:
I have a suggestion in https://github.com/ignitionrobotics/ign-rendering/issues/381 for a way to let each user calibrate their screens independently.

chapulina:
I'm closing this in favor of #446, which was merged into Fortress. We can backport that if needed.

PR #224, opened 2021-02-02
3 to 4

adlarkin:
Forward-port `ign-rendering3` to `ign-rendering4`.

Comparison: https://github.com/ignitionrobotics/ign-rendering/compare/ign-rendering4...ign-rendering3

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -18,9 +18,68 @@
 #include <ignition/math/Vector2.hh>
 #include <ignition/math/Vector3.hh>
 
+#include <ignition/common/Console.hh>
+#include <ignition/math/Helpers.hh>
+
 #include "ignition/rendering/ogre2/Ogre2Camera.hh"
 #include "ignition/rendering/ogre2/Ogre2GpuRays.hh"
 #include "ignition/rendering/ogre2/Ogre2RenderEngine.hh"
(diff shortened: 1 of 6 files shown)


chapulina (approved):

ogre2/src/Ogre2GpuRays.cc
```diff
@@ -589,6 +794,24 @@ void Ogre2GpuRays::Setup1stPass()
     this->dataPtr->ogreCompositorWorkspace1st[i] =
         ogreCompMgr->addWorkspace(this->scene->OgreSceneManager(),
         rt, this->dataPtr->cubeCam[i], wsDefName, false);
+
+    // add laser retro material switcher to render target listener
+    // so we can switch to use laser retro material when the camera is being udpated
```
mjcarroll, line 799:
```suggestion
    // so we can switch to use laser retro material when the camera is being updated
```

I think there is also a line length problem here.

chapulina (changes requested):
Interesting, there are a couple of codechecker issues:

```
  /github/workspace/ogre2/src/Ogre2GpuRays.cc:799:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
  /github/workspace/ogre2/src/Ogre2GpuRays.cc:810:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
```

Let's fix them here

chapulina (approved):

PR #230, opened 2021-02-03
Fix converting Pbs to Unlit material conversion

iche033:
Thermal camera uses unlit material for background objects. When visualizing thermal camera output in a more complex environment, I noticed that the textures are not correct copied over from Pbs to Unlit datablock. The changes here make sure we set the correct texture file to the unlit material.


adlarkin (approved):
We'll need to make an `ign-rendering` patch release for this to be used in SubT (`4.3.1`, right?).

iche033:
> We'll need to make an ign-rendering patch release for this to be used in SubT (4.3.1, right?).

yeah 4.3.1

PR #235, opened 2021-02-05
Add support for 8 bit thermal camera image format

iche033:
## Summary

The existing thermal camera implementation produces image output in 16 bit format and that's being hard coded in various places including the shaders. This PR makes the implementation slightly more generic to support generating 8 bit data.

There are a few workarounds that had to be done so that this change does not break ABI. For example, the new thermal image event's function callback is hard coded to `uint16_t` and that can not be changed so unfortunately I had to workaround this by packing 8 bit data into 16 bit array. 

## Checklist
- [x] Added tests

---

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,11 @@
 
 ### Ignition Rendering 4.X
 
+### Ignition Rendering 4.x.x (202x-xx-xx)
+
+1. Add support for 8 bit thermal camera image format
+    * [Pull Request #235](https://github.com/ignitionrobotics/ign-rendering/pull/235)
+
 ### Ignition Rendering 4.3.1 (2021-02-03)
 
 1. Fix converting Pbs to Unlit material conversion (#230)
--- a/ogre2/src/Ogre2ThermalCamera.cc
+++ b/ogre2/src/Ogre2ThermalCamera.cc
@@ -79,7 +79,8 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \param[in] _format Inage format
   public: void SetFormat(PixelFormat _format);
 
-  /// \brieft Set linear resolution
+  /// \brief Set temperature linear resolution
+  /// \param[in] _resolution Temperature linear resolution
   public: void SetLinearResolution(double _resolution);
 
   /// \brief Callback when a render target is about to be rendered
@@ -133,7 +134,7 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \brief thermal camera image format
   private: PixelFormat format = PF_L16;
 
-  /// \brief thermal camera image format
+  /// \brief thermal camera image bit depth
   private: int bitDepth = 16u;
 };
 }
--- a/ogre2/src/Ogre2ThermalCamera.cc
+++ b/ogre2/src/Ogre2ThermalCamera.cc
@@ -75,6 +75,13 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \brief destructor
   public: ~Ogre2ThermalCameraMaterialSwitcher() = default;
 
+  /// \brief Set image format
+  /// \param[in] _format Inage format
+  public: void SetFormat(PixelFormat _format);
+
+  /// \brieft Set linear resolution
+  public: void SetLinearResolution(double _resolution);
+
   /// \brief Callback when a render target is about to be rendered
   /// \param[in] _evt Ogre render target event containing information about
   /// the source render target.
@@ -119,6 +126,15 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \brief A map of ogre sub item pointer to their original hlms material
   private: std::unordered_map<Ogre::SubItem *, Ogre::HlmsDatablock *>
       datablockMap;
+
+  /// \brief linear temperature resolution. Defaults to 10mK
+  private: double resolution = 0.01;
+
+  /// \brief thermal camera image format
+  private: PixelFormat format = PF_L16;
+
+  /// \brief thermal camera image format
+  private: int bitDepth = 16u;
 };
 }
 }
@@ -130,7 +146,7 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
 class ignition::rendering::Ogre2ThermalCameraPrivate
 {
   /// \brief The thermal buffer
-  public: uint16_t *thermalBuffer = nullptr;
+  public: unsigned char *thermalBuffer = nullptr;
 
   /// \brief Outgoing thermal data, used by newThermalFrame event.
   public: uint16_t *thermalImage = nullptr;
@@ -172,6 +188,9 @@ class ignition::rendering::Ogre2ThermalCameraPrivate
   /// This only affects objects that are not heat sources
   /// TODO(anyone) add API for setting this value?
   public: bool rgbToTemp = true;
+
+  /// \brief bit depth of each pixel
+  public: unsigned int bitDepth = 16u;
 };
 
 using namespace ignition;
@@ -196,6 +215,18 @@ Ogre2ThermalCameraMaterialSwitcher::Ogre2ThermalCameraMaterialSwitcher(
   this->ogreCamera = this->scene->OgreSceneManager()->findCamera(this->name);
 }
 
+//////////////////////////////////////////////////
+void Ogre2ThermalCameraMaterialSwitcher::SetFormat(PixelFormat _format)
+{
+  this->format = _format;
+  this->bitDepth = 8u * PixelUtil::BytesPerChannel(format);
+}
+
+//////////////////////////////////////////////////
+void Ogre2ThermalCameraMaterialSwitcher::SetLinearResolution(double _resolution)
+{
+  this->resolution = _resolution;
+}
 //////////////////////////////////////////////////
 void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
     const Ogre::RenderTargetEvent & /*_evt*/)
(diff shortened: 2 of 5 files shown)


ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -119,6 +127,15 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
...
       datablockMap;
+
+  /// \brief linear temperature resolution. Defaults to 10mK
+  private: double resolution = 0.01;
+
+  /// \brief thermal camera image format
+  private: PixelFormat format = PF_L16;
+
+  /// \brief thermal camera image bit depth
+  private: int bitDepth = 16u;
```
ahcorde, line 138:
```suggestion
  private: unsigned int bitDepth = 16u;
```

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -803,12 +857,32 @@ void Ogre2ThermalCamera::PostRender()
...
+    // \todo(anyone) add a new ConnectNewThermalFrame function that accepts
+    // a generic unsigned char array instead of uint16_t so we can do a direct
+    // memcpy of the data
+    for (unsigned int i = 0u; i < height; ++i)
+    {
+      for (unsigned int j = 0u; j < width; ++j)
+      {
+        unsigned int idx = i*width + j;
+        int v = static_cast<int>(this->dataPtr->thermalBuffer[idx]);
+        this->dataPtr->thermalImage[idx] = static_cast<uint16_t>(v);
```
ahcorde, line 872:
```suggestion
        this->dataPtr->thermalImage[idx] = static_cast<uint16_t>(this->dataPtr->thermalBuffer[idx]);
```

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -803,12 +857,32 @@ void Ogre2ThermalCamera::PostRender()
...
+      }
+    }
+  }
+  else
+  {
+    // fill thermal data
+    memcpy(this->dataPtr->thermalImage, this->dataPtr->thermalBuffer,
+        height*width*channelCount*bytesPerChannel);
```
ahcorde, line 880:
```suggestion
        height * width * channelCount * bytesPerChannel);
```

test/integration/thermal_camera.cc
```diff
@@ -256,6 +260,191 @@ void ThermalCameraTest::ThermalCameraBoxes(
...
+  box->SetLocalScale(unitBoxSize, unitBoxSize, unitBoxSize);
+
+  // set box temperature
+  float boxTemp = 310.0;
+  box->SetUserData("temperature", boxTemp);
+
+  root->AddChild(box);
+  {
+    // range is hardcoded in shaders
+    float boxTempRange = 3.0;
```
ahcorde, line 315:
```suggestion
    float boxTempRange = 3.0f;
```

ahcorde (changes requested):

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -263,18 +295,14 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
-              subItem->setCustomParameter(this->customParamIdx,
-                  Ogre::Vector4(color, 0, 0, 0.0));
-            }
+            // normalize temperature value
+            float color = (temp / this->resolution) / ((1 << bitDepth) - 1.0);
+
+            // set g, b, a to 0. This will be used by shaders to determine
+            // if particular fragment is a heat source or not
+            // see media/materials/programs/thermal_camera_fs.glsl
+            subItem->setCustomParameter(this->customParamIdx,
```
adlarkin, line 304: Do we no longer need the `if (!subItem->hasCustomParameter(this->customParamIdx))` check?

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -319,13 +347,18 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
           if (minTemperature && maxTemperature)
           {
             // make sure the temperature range is between [0, 655.35] kelvin
+            // for 16 bit format and 10mK resolution
```
adlarkin, line 350:
```suggestion
            // make sure the temperature range is between [min,max]
            // for the given pixel format and camera resolution
```
I think it's better to take out the explicit reference to 16 bit format and 10mK resolution, now that these parameters are configurable. That way, users aren't confused by the comment in case the resolution and pixel format are different.

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -803,12 +857,32 @@ void Ogre2ThermalCamera::PostRender()
...
+  {
+    // workaround for populating a 16bit image buffer with 8bit data
+    // \todo(anyone) add a new ConnectNewThermalFrame function that accepts
+    // a generic unsigned char array instead of uint16_t so we can do a direct
+    // memcpy of the data
+    for (unsigned int i = 0u; i < height; ++i)
+    {
+      for (unsigned int j = 0u; j < width; ++j)
+      {
+        unsigned int idx = i*width + j;
```
adlarkin, line 870:
```suggestion
        unsigned int idx = (i * width) + j;
```

ogre2/src/media/materials/programs/heat_signature_fs.glsl
```diff
@@ -35,12 +35,16 @@ out vec4 fragColor;
 uniform float minTemp = 0.0;
 uniform float maxTemp = 100.0;
 
+uniform int bitDepth;
+uniform float resolution;
+
 // map a temperature from the [0, 655.35] range to the [minTemp, maxTemp]
-// range (655.35 is the largest Kelvin value allowed)
+// range (655.35 is the largest Kelvin value allowed for 16 bit with
+// 10mK resolution)
```
adlarkin, line 43: Similar to my other comment about possible confusion when referring to 10mK resolution and 16 bit format, perhaps we should change this comment to say something like `map a temperature from the camera's [min,max] range to the user-defined [minTemp,maxTemp] range`

ogre2/src/media/materials/programs/thermal_camera_fs.glsl
```diff
@@ -74,7 +76,9 @@ void main()
     if (isHeatSource)
     {
       // heat is normalized so convert back to work in kelvin
-      temp = heat * 655.35;
+      // for 16 bit camera and 0.01 resolution:
+      //     ((1 << bitDepth) - 1)* resolution = 655.35
```
adlarkin, line 80:
```suggestion
      //     ((1 << bitDepth) - 1) * resolution = 655.35
```

test/integration/thermal_camera.cc
```diff
@@ -256,6 +260,191 @@ void ThermalCameraTest::ThermalCameraBoxes(
   ignition::rendering::unloadEngine(engine->Name());
 }
 
+//////////////////////////////////////////////////
+void ThermalCameraTest::ThermalCameraBoxes8Bit(
+    const std::string &_renderEngine)
+{
+  int imgWidth = 50;
+  int imgHeight = 50;
+  double aspectRatio = imgWidth/imgHeight;
```
adlarkin, line 269:
```suggestion
  double aspectRatio = imgWidth / imgHeight;
```

test/integration/thermal_camera.cc
```diff
@@ -256,6 +260,191 @@ void ThermalCameraTest::ThermalCameraBoxes(
...
+
+  root->AddChild(box);
+  {
+    // range is hardcoded in shaders
+    float boxTempRange = 3.0;
+    double farDist = 10.0;
+    double nearDist = 0.15;
+    double hfov = 1.05;
+    double minTemp = 253.0;
+    double maxTemp = 673.0;
```
adlarkin, line 320: If a 16-bit camera with .01 mK linear resolution has a maximum temperature of `655.35` kelvin, can an 8-bit camera have a maximum temperature that is higher than `655.35`? I'd think that the 8-bit camera would have more of a limited range than a 16-bit camera, but maybe I am misunderstanding the math.

Also, speaking of the math behind this, can a comment be added that explains how the values for `minTemp` and `maxTemp` were calculated?

Changelog.md
```diff
@@ -2,6 +2,11 @@
 
 ### Ignition Rendering 4.X
 
+### Ignition Rendering 4.x.x (202x-xx-xx)
+
+1. Add support for 8 bit thermal camera image format
+    * [Pull Request #235](https://github.com/ignitionrobotics/ign-rendering/pull/235)
```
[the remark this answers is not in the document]
iche033: we used to do this quite a bit in our PRs. I did it here just for convenience since we're planning on making another release soon. The `xx` are just placeholders

ogre2/src/Ogre2ThermalCamera.cc
iche033: the value of the `color` variable above is different for an 8 bit and a 16 bit camera. So when testing in a world containing both 16 bit and 8 bit cameras, we'll need to constantly override the values on each camera update.

test/integration/thermal_camera.cc
iche033: The `minTemp` and `maxTemp` are actually from physical thermal camera specs.

As an example,  the max temperature that can be represented by an 8 bit camera with 3 degree resolution is: 
* (2^bitDepth-1) * resolution = 255*3 = 765 degrees.

Here we have a user-defined max of 673.0. So any values above 673.0 will be clamped even when the image is able to store a higher number.

On the other hand, if user species a max temp of 800, which is above the limit (765) on what the 8 bit 3 resolution camera can produce, we can only return a max of 765 degrees

This FLIR page gave another example: https://www.flir.com/support-center/Instruments/what-is-the-relation-between-temperaturelinearresolution-and-sensorgainmode-temperature-range-for-the-flir-ax5-series/

> Even though the SensorGainMode is set to LowGainMode (–40 to +550°C), the camera signal will saturate around 382°C.

ogre2/src/Ogre2ThermalCamera.cc
iche033: 43c80a9

ogre2/src/Ogre2ThermalCamera.cc
iche033: 43c80a9

ogre2/src/Ogre2ThermalCamera.cc
iche033: 43c80a9

test/integration/thermal_camera.cc
iche033: 43c80a9

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -75,6 +75,14 @@ class Ogre2ThermalCameraMaterialSwitcher : public Ogre::RenderTargetListener
   /// \brief destructor
   public: ~Ogre2ThermalCameraMaterialSwitcher() = default;
 
+  /// \brief Set image format
+  /// \param[in] _format Inage format
```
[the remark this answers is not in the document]
iche033: 43c80a9

ogre2/src/Ogre2ThermalCamera.cc
iche033: 43c80a9

ogre2/src/Ogre2ThermalCamera.cc
iche033: 43c80a9

ogre2/src/media/materials/programs/heat_signature_fs.glsl
iche033: 43c80a9

ogre2/src/media/materials/programs/thermal_camera_fs.glsl
iche033: 43c80a9

test/integration/thermal_camera.cc
iche033: 43c80a9

test/integration/thermal_camera.cc
adlarkin: Thanks for the explanation. Should we add a note here that these min/max values are from actual physical thermal camera tests, and include a link to the specs in case we have to re-visit this test later?

iche033: 8a5bf29ea77db8d83618bc6bb93a5c8f7f29c16c

adlarkin (approved):
All comments have been addressed, and all checks have passed - LGTM!

ahcorde (approved):

PR #241, opened 2021-02-10
Fix setting particle size

iche033:
## Summary

Setting the size of the particles after the particle emitter is created had no effect. The solution is to destroy and recreate the particle system with the new size. 

I also updated the code so that the ParticleEmitter class does not create all different types of particle emitters (e.g. point, box, ellipsoid, etc) when it's initialized. When the user changes the emitter type, we'll destroy the existing emitter and create a new one with the specified type.

Note: my original intention was to have a flag to mark the emitter as dirty when the above params changed and let `PreRender()` handle the recreation of the emitter. However it was found that overriding `PreRender()` broke ABI so I worked around it with a manual `PreRenderImpl` call

## Checklist
- [ ] Added tests
- [ ] `codecheck` passed (See
  [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See
  [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review
[another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+)
to support the maintainers


ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -115,11 +123,16 @@ void Ogre2ParticleEmitter::Ogre2ParticleEmitter::SetType(
...
-    this->dataPtr->emitters[i]->setEnabled(false);
+  if (this->type == _type)
+    return;
 
-  this->dataPtr->emitters[_type]->setEnabled(true);
   this->type = _type;
+
+  this->dataPtr->emitterDirty = true;
+  // todo(anyone) remove this call. We had to do this since we can't override
+  // PreRender() as it breaks ABI. We should rename PreRenderImp to PreRender()
```
adlarkin, line 133:
```suggestion
  // PreRender() as it breaks ABI. We should rename PreRenderImpl to PreRender()
```

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -227,8 +237,14 @@ void Ogre2ParticleEmitter::SetParticleSize(
            << "Particle size values should be non-negative." << std::endl;
     return;
   }
-  this->dataPtr->ps->setDefaultDimensions(_size[0], _size[1]);
+
   this->particleSize = _size;
+
+  this->dataPtr->emitterDirty = true;
+  // todo(anyone) remove this call. We had to do this since we can't override
+  // PreRender() as it breaks ABI. We should rename PreRenderImp to PreRender()
```
adlarkin, line 245:
```suggestion
  // PreRender() as it breaks ABI. We should rename PreRenderImpl to PreRender()
```

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -413,7 +428,50 @@ void Ogre2ParticleEmitter::SetColorRangeImage(const std::string &_image)
...
+    this->Destroy();
+    this->CreateParticleSystem();
 
+    // make direct ogre calls here so we don't mark emitter as dirty again
+    this->dataPtr->ps->setDefaultDimensions(
+        this->particleSize[0], this->particleSize[1]);
+
+    this->SetEmitterSize(this->emitterSize);
+
+    // set other properties
```
adlarkin, line 453: I believe we need to call `SetDuration` and `SetEmitting` here as well, since these methods set properties for `this->dataPtr->emitter`.

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -424,16 +482,12 @@ void Ogre2ParticleEmitter::Init()
...
-  for (auto i = 0; i < EmitterType::EM_NUM_EMITTERS; ++i)
-  {
-    this->dataPtr->emitters[i] =
-      this->dataPtr->ps->addEmitter(kOgreEmitterTypes[i]);
-    this->dataPtr->emitters[i]->setEnabled(false);
-    this->dataPtr->emitters[i]->setDirection(Ogre::Vector3::UNIT_X);
-  }
-  this->dataPtr->emitters[EmitterType::EM_POINT]->setEnabled(true);
+  // Instantiate particle emitter and their default parameters.
+  // default in point emitter
```
adlarkin, line 486:
```suggestion
  // Emitter type is point unless otherwise specified
```

adlarkin (commented):
Looks good!

I would like to update the particle emitter test. Right now, the [current test](https://github.com/ignitionrobotics/ign-rendering/blob/cd86e7354ac5d0d44856919e94db7cfef13ae754/src/ParticleEmitter_TEST.cc) creates a particle emitter, checks to make sure it was created properly, modifies _all particle emitter properties_, and then checks the newly modified emitter to make sure that everything was update correctly. I think that we should add a test case where we create an emitter, but then modify _only the type_ and then _only the particle size_. That way, we can make sure that all of the other properties of the emitter are still correct if only the type and/or particle size is modified. I can take care of updating this test if you'd like.

ogre2/src/Ogre2ParticleEmitter.cc
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

ogre2/src/Ogre2ParticleEmitter.cc
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

ogre2/src/Ogre2ParticleEmitter.cc
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

ogre2/src/Ogre2ParticleEmitter.cc
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -115,11 +123,16 @@ void Ogre2ParticleEmitter::Ogre2ParticleEmitter::SetType(
     return;
   }
```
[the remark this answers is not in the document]
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -413,7 +428,50 @@ void Ogre2ParticleEmitter::SetColorRangeImage(const std::string &_image)
...
 {
   Ogre2Visual::Init();
+  this->CreateParticleSystem();
+}
+
+//////////////////////////////////////////////////
+void Ogre2ParticleEmitter::PreRenderImpl()
+{
+  // \todo(anyone) rename this function to PreRender() so it overrides function
+  // from base class
```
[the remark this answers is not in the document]
adlarkin: https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919

adlarkin:
I went ahead and addressed some of my feedback in https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919.

I wanted to follow up about my [previous comment](https://github.com/ignitionrobotics/ign-rendering/pull/241#pullrequestreview-589622558) regarding tests. So, it looks like the [particle emitter unit test](https://github.com/ignitionrobotics/ign-rendering/blob/cd86e7354ac5d0d44856919e94db7cfef13ae754/src/ParticleEmitter_TEST.cc) tests the data members that belong to the [`BaseParticleEmitter` class](https://github.com/ignitionrobotics/ign-rendering/blob/8bb934d394fdbd54cbda81b6fb530b68813bfdd9/include/ignition/rendering/base/BaseParticleEmitter.hh), but what we should really do here is add a test to ensure that the [`Ogre::ParticleEmitter` object in the `Ogre2ParticleEmitter` class](https://github.com/ignitionrobotics/ign-rendering/blob/a19125e60dafbc01f48f882465cda6b3b5615919/ogre2/src/Ogre2ParticleEmitter.cc#L47-L48) is being updated correctly whenever an emitter is re-created due to a type or particle size change. Is it possible to test the validity of this data member somehow, or is it impossible since this data member belongs to [`Ogre2ParticleEmitterPrivate`](https://github.com/ignitionrobotics/ign-rendering/blob/a19125e60dafbc01f48f882465cda6b3b5615919/ogre2/include/ignition/rendering/ogre2/Ogre2ParticleEmitter.hh#L102-L103)?

ogre2/src/Ogre2ParticleEmitter.cc
```diff
@@ -91,6 +94,11 @@ void Ogre2ParticleEmitter::Destroy()
...
   if (this->dataPtr->ps)
   {
+    this->dataPtr->ps->removeAllAffectors();
+    this->dataPtr->colorInterpolatorAffector = nullptr;
+    this->dataPtr->colorImageAffector = nullptr;
+    this->dataPtr->scalerAffector = nullptr;
+
     this->scene->OgreSceneManager()->destroyParticleSystem(
         this->dataPtr->ps);
     this->dataPtr->ps = nullptr;
```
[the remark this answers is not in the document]
iche033: ogre owns the pointer so we shouldn't need to delete it ourselves

iche033:
> Is it possible to test the validity of this data member somehow, or is it impossible since this data member belongs to Ogre2ParticleEmitterPrivate?

I'm trying to keep the tests generic so we can run them for different rendering engines. The other option is an integration test to test before and after changing type or particles, e.g. if you have a point particle emitter with a duration or 1 sec and then you change it to a box emitter, you should see the particles more spread out while only being enabled for 1 sec as well.

adlarkin:
> I'm trying to keep the tests generic so we can run them for different rendering engines.

Good point, I didn't think about that :+1: 

> The other option is an integration test to test before and after changing type or particles, e.g. if you have a point particle emitter with a duration or 1 sec and then you change it to a box emitter, you should see the particles more spread out while only being enabled for 1 sec as well.

That's not a bad idea. The only thing to consider with an integration test like this is that we should try to test all of the emitter properties, not just duration (particle size, velocity, scale rate, etc.), which I think would be very difficult to test. So, we can either write an integration test that checks for some of these things, or just leave everything as-is (i.e., just keep the unit test). I'm not sure what would be better.

iche033:
I'm leaning towards keeping the test as is given the amount of effort needed and time to get other particle effects changes in. Let's revisit this in the future. I've ticked an issue: https://github.com/ignitionrobotics/ign-rendering/issues/249

adlarkin:
Most of this looks good to me, but before merging, @iche033 can you take a look at my feedback/review comments above and changes I made in https://github.com/ignitionrobotics/ign-rendering/pull/241/commits/a19125e60dafbc01f48f882465cda6b3b5615919? In particular, I'd like to get your confirmation about [this change](https://github.com/ignitionrobotics/ign-rendering/pull/241#discussion_r575385678). If all the changes look good to you, then I am okay with approving and merging this.

iche033:
>  If all the changes look good to you, then I am okay with approving and merging this.

the changes look good, thanks!

adlarkin (approved):

PR #243, opened 2021-02-10
Handle non-positive object temperatures

adlarkin:
Related to https://github.com/ignitionrobotics/ign-gazebo/pull/621. This handles temperatures that are <= 0 kelvin for the thermal camea.

## Checklist
- [ ] Added tests
- [ ] Added example world and/or tutorial
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+archived%3Afalse+) to support the maintainers

### DIFF
--- a/ogre2/src/Ogre2ThermalCamera.cc
+++ b/ogre2/src/Ogre2ThermalCamera.cc
@@ -266,6 +266,7 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
       if (tempAny.index() != 0 && !std::holds_alternative<std::string>(tempAny))
       {
         float temp = -1.0;
+        bool foundTemp = true;
         try
         {
           temp = std::get<float>(tempAny);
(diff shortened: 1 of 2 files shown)


iche033:
I made some tweaks in a3a88548060f7d17d0211bfa6bc9cdc9e12ade8e that clamps neg temperature to 0 and enable our shaders work with 0 temp value. This should make the test in https://github.com/ignitionrobotics/ign-gazebo/pull/621 pass

adlarkin:
> I made some tweaks in [a3a8854](https://github.com/ignitionrobotics/ign-rendering/commit/a3a88548060f7d17d0211bfa6bc9cdc9e12ade8e) that clamps neg temperature to 0

I just took a look and tested this out, it works for me! I'm marking this as ready for review. If we merge https://github.com/ignitionrobotics/ign-gazebo/pull/621, this will need to be merged and we will need to make a patch release of `ign-rendering` before making the related `ign-gazebo` release.

adlarkin:
As mentioned in https://github.com/ignitionrobotics/ign-gazebo/pull/621#issuecomment-777028822, I am holding off on merging this for now since it would effect `ign-gazebo` and `ign-rendering` releases that are currently in-process at the time of this writing. I will mark this as a draft and re-open it once the releases are complete.

ogre2/src/Ogre2ThermalCamera.cc
```diff
@@ -286,30 +287,36 @@ void Ogre2ThermalCameraMaterialSwitcher::preRenderTargetUpdate(
...
               temp = -1.0;
+              foundTemp = false;
             }
           }
         }
 
-        // only accept positive temperature (in kelvin)
-        if (temp >= 0.0)
+        // if a non-positive temperature was given, clamp it to 0
+        if (foundTemp && temp < 0.0)
```
nkoenig, line 296: This only clamps the `temp` value when `foundTemp` is true. Is it possible for `foundTemp` to be false and `temp` to be < 0.0? If so, should you handle that case?

adlarkin: > Is it possible for foundTemp to be false and temp to be < 0.0?

I don't think so. The only way `foundTemp` is set to true is if `temperature` is defined/attached to a visual, which would mean that `<temperature>` needs to be specified in the SDF file in the `ign-gazebo` context. If `<temperature>` is not specified, then this code should run (background object where object colors are used to define temperatures - since we're using rgb here, it shouldn't be possible to have an rgb value that is < 0): https://github.com/ignitionrobotics/ign-rendering/blob/8e0de0dd3e499c0fc0191a2cd8b794b0aab69980/ogre2/src/Ogre2ThermalCamera.cc#L387-L417

iche033 (approved):

PR #189, opened 2020-12-22
Thermal Camera: Visualize objects with a heat signature

adlarkin:
Requires https://github.com/ignitionrobotics/ign-gazebo/pull/49 and makes use of #211.

This expands the current capabilities of the thermal camera to allow for rendering of objects with a heat signature. Currently, only objects with a uniform temperature are rendered by the thermal camera (objects with a heat signature can have varying temperature).

To learn how to use/test these updates, please see the related tutorial: https://github.com/ignitionrobotics/ign-sensors/pull/79

Testing can be done by running:
```
ign gazebo -r thermal_camera.sdf
```

---

TODO
- [x] Update thermal camera unit tests (after further inspection, these don't need to be updated)
- [x] Update thermal camera integration tests
- [x] Address any TODOs in the code (look at [files changed](https://github.com/ignitionrobotics/ign-rendering/pull/189/files))

Signed-off-by: Ashton Larkin <ashton@openrobotics.org>

PR #246, opened 2021-02-12
4 ➡️ 5

chapulina:
Port `ign-rendering4` to `main`.

Branch comparison: https://github.com/ignitionrobotics/ign-rendering/compare/main...ign-rendering4

**Note to maintainers**: Remember to **Merge** with commit (not squash-merge or rebase)


adlarkin (approved):

chapulina:
This new test failure could be related to #189 or #235:

```
/Users/jenkins/workspace/ignition_rendering-ci-pr_any-homebrew-amd64/ign-rendering/test/integration/thermal_camera.cc:233
The difference between boxTemp and temp is 14, which exceeds boxTempRange, where
boxTemp evaluates to 310,
temp evaluates to 296, and
boxTempRange evaluates to 3.
```

adlarkin:
> This new test failure could be related to #189 or #235

I took a look at homebrew CI. The failure is for the `ThermalCameraBoxesUniformTemp` test, which was actually already implemented before #189 and #235. That being said though, #189 and #235 did change some of the code slightly for processing uniform temperatures, so maybe these PRs are indeed related to this new failure. It's interesting that this failure only happened on MacOS and not Ubuntu. All checks passed on #235 as well, so I am not sure what's going on here.

@iche033 any ideas?

chapulina:
> this failure only happened on MacOS and not Ubuntu. 

I haven't looked closely into the code, but when this happens, I usually look for anything in the test that relies on timing, because the Homebrew CI machines are usually less powerful than the Ubuntu ones and may require more sleeping.

In any case, I retriggered the build to see if the test passes this time. I don't think we need to block this PR on a flaky test, but I think it would be good to address the flakiness back on `ign-rendering4`.

adlarkin:
> I retriggered the build to see if the test passes this time

It looks like the test still failed on MacOS CI. I'm taking a closer look at this and will post an update here if I find anything that needs to be addressed. Let's hold off on merging this for now in case something needs to be fixed.

iche033:
hmm not sure why. The tests seem to be passing in ign-rendering4, e.g. in https://github.com/ignitionrobotics/ign-rendering/pull/241. May we've changed something in edifice.

adlarkin:
I tried to do some debugging about the new test failure. I opened up two draft PRs in order to trigger/compare CI:
* #247: Branches off of this PR in order to do more digging into the specifics of the test failure
* #248: Branches off of `main` (edifice) to see what the current CI results are without this forward port

I found that all tests pass on #248, but I am still seeing the MacOS failure on #247 even after trying a few things. This makes me think that the new test failure is due to a combination of change(s) in this forward port and changes in edifice, since tests were passing on the dome and edifice branches without this forward port.

The interesting thing is that based on my testing in #247, it looks like only one pixel in the image fails the check, and I have no idea why: https://github.com/ignitionrobotics/ign-rendering/pull/247#issuecomment-778495448

I'm out of ideas at this point. Does anyone have any thoughts on what may be going on here?

chapulina:
I'll merge this to get some PRs moving forward while you peeps are looking into the thermal camera failure.

adlarkin:
> I'll merge this to get some PRs moving forward while you peeps are looking into the thermal camera failure.

I create an issue for this so that it can be tracked: https://github.com/ignitionrobotics/ign-rendering/issues/253

PR #269, opened 2021-03-09
Expose particle scatter ratio parameter

iche033:

# 🎉 New feature

## Summary

used by https://github.com/ignitionrobotics/ign-gazebo/pull/674

This lets users configure the particle scatter ratio parameter that was previously hardcoded in a couple different sensor src files.

The `particleScatterRatio` param determines the amount of scattering effect applied to sensors (depth camera and lidars). The higher the value, the more "scattering" (noise) is added to the depth readings.  This value was fixed and now it's configuration though the `ParticleEmitter`'s `UserAny` object

## Test it

see https://github.com/ignitionrobotics/ign-gazebo/pull/674

## Checklist
- [x] Added tests

### DIFF
--- a/ogre2/src/Ogre2ParticleNoiseListener.cc
+++ b/ogre2/src/Ogre2ParticleNoiseListener.cc
@@ -87,7 +87,7 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
           static_cast<float>(particleStddev));
 
       // get particle scatter ratio value from particle emitter user data
-      // set pass that to the shaders
+      // and pass that to the shaders
       Ogre::Any userAny = ps->getUserObjectBindings().getUserAny();
       if (!userAny.isEmpty() && userAny.getType() == typeid(unsigned int))
       {
@@ -103,39 +103,41 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
         }
         Ogre2VisualPtr ogreVisual =
             std::dynamic_pointer_cast<Ogre2Visual>(result);
-
-        std::string particleScatterRatioKey = "particle_scatter_ratio";
-        Variant particleScatterRatioAny =
-            ogreVisual->UserData(particleScatterRatioKey);
-        if (particleScatterRatioAny.index() != std::variant_npos)
+        if (ogreVisual)
         {
-          float ratio = -1.0;
-          try
+          const std::string particleScatterRatioKey = "particle_scatter_ratio";
+          Variant particleScatterRatioAny =
+              ogreVisual->UserData(particleScatterRatioKey);
+          if (particleScatterRatioAny.index() != std::variant_npos)
           {
-            ratio = std::get<float>(particleScatterRatioAny);
-          }
-          catch(...)
-          {
+            float ratio = -1.0;
             try
             {
-              ratio = std::get<double>(particleScatterRatioAny);
+              ratio = std::get<float>(particleScatterRatioAny);
             }
             catch(...)
             {
               try
               {
-                ratio = std::get<int>(particleScatterRatioAny);
+                ratio = std::get<double>(particleScatterRatioAny);
               }
-              catch(std::bad_variant_access &e)
+              catch(...)
               {
-                ignerr << "Error casting user data: " << e.what() << "\n";
-                ratio = -1.0;
+                try
+                {
+                  ratio = std::get<int>(particleScatterRatioAny);
+                }
+                catch(std::bad_variant_access &e)
+                {
+                  ignerr << "Error casting user data: " << e.what() << "\n";
+                  ratio = -1.0;
+                }
               }
             }
-          }
-          if (ratio > 0)
-          {
-            this->particleScatterRatio = ratio;
+            if (ratio > 0)
+            {
+              this->particleScatterRatio = ratio;
+            }
           }
         }
       }
--- a/ogre2/src/Ogre2ParticleNoiseListener.cc
+++ b/ogre2/src/Ogre2ParticleNoiseListener.cc
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Open Source Robotics Foundation
+ * Copyright (C) 2021 Open Source Robotics Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


ogre2/src/Ogre2ParticleNoiseListener.hh
```diff
@@ -40,9 +40,19 @@ namespace ignition
...
+      /// \brief Pointer to camera
       private: Ogre::Camera *ogreCamera = nullptr;
+
+      /// \brief Pointer to ogre matieral with shaders for applying particle
+      /// scattering effect to sensors
       private: Ogre::MaterialPtr ogreMaterial;
+
+      /// \brief Particle scatter ratio. This is used to determine the ratio of
+      /// particles that will detected by sensors
+      private: double particleScatterRatio = 0.65;
```
adlarkin, line 55:
```suggestion
      /// \brief Particle scatter ratio. This is used to determine the ratio of
      /// particles that will detected by sensors. Increasing the ratio increases
      /// the scatter of the particles, making it appear less dense. Decreasing
      /// the ratio decreases the scatter of the particles, making it appear
      /// more dense. This value should be > 0.
      private: float particleScatterRatio = 0.65;
```

I think it'd be good to add a note about how the value of this parameter modifies behavior in case we need to re-visit this later (feel free to modify the documentation as needed if my explanation here is incorrect or unclear).

Also, I think we can change this to be of type `float`, since shaders and the data retrieved from ORGE use floats, right?

test/integration/depth_camera.cc
```diff
@@ -678,6 +678,69 @@ void DepthCameraTest::DepthCameraParticles(
     EXPECT_LT(pointParticleAvg, pointAvg);
     EXPECT_LT(depthParticleAvg, depthAvg);
 
+    // test setting particle scatter ratio
+    // reduce particle scatter ratio - this creates a "less dense" particle
+    // emitter so we should have larger depth values on avg since fewers
+    // depth readings are occluded by particles
```
adlarkin, line 684: If we reduce the scatter ratio, doesn't that mean there is less scatter, which means that the emitter is actually more dense, so depth values on average would be less? Based on my understanding/interpretation, the scatter ratio essentially determines how much scatter exists in the particles (lower ratio = less scatter, higher ratio = more scatter). I left another comment about documenting the behavior/interpretation of the scatter ratio parameter, hopefully my interpretation was correct there.

ogre2/src/Ogre2ParticleNoiseListener.cc
```diff
@@ -82,6 +85,63 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
           pass->getFragmentProgramParameters();
       psParams->setNamedConstant("particleStddev",
           static_cast<float>(particleStddev));
+
+      // get particle scatter ratio value from particle emitter user data
+      // set pass that to the shaders
```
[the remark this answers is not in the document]
iche033: fixed. db02aea

ogre2/src/Ogre2ParticleNoiseListener.cc
```diff
@@ -82,6 +85,63 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
...
+        {
+          result = this->scene->VisualById(
+              Ogre::any_cast<unsigned int>(userAny));
+        }
+        catch(Ogre::Exception &e)
+        {
+          ignerr << "Ogre Error:" << e.getFullDescription() << "\n";
+        }
+        Ogre2VisualPtr ogreVisual =
+            std::dynamic_pointer_cast<Ogre2Visual>(result);
```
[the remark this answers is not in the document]
iche033: yep done. db02aea

ogre2/src/Ogre2ParticleNoiseListener.cc
```diff
@@ -82,6 +85,63 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
...
+
+        std::string particleScatterRatioKey = "particle_scatter_ratio";
```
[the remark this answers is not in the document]
iche033: db02aea

src/ParticleEmitter_TEST.cc
```diff
@@ -91,6 +91,9 @@ void ParticleEmitterTest::CheckBasicAPI()
   math::Color    expectedColorEnd        = ignition::math::Color::White;
   double         expectedScaleRate       = 1;
   std::string    expectedColorRangeImage = "";
+  // Particle scatter ratio is stored in user data
+  // \todo(anyone) Add API to set scatter ratio
```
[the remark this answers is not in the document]
iche033: this unfortunately requires adding a virtual function in the base class and breaks ABI by changing the vtable. So we would need to do it an unreleased version.

ogre2/src/Ogre2ParticleNoiseListener.hh
iche033: added and tweaked note. db02aea

test/integration/depth_camera.cc
iche033: > If we reduce the scatter ratio, doesn't that mean there is less scatter, which means that the emitter is actually more dense
the emitter would actually be less dense - more rays will be casting through the particles (due to less occlusion). So the average depth value will be higher (longer rays).  On the other hand, a higher scatter ratio means more dense and more rays will be bounced back by the particles, resulting in shorter depth values. I've tweaked the description to clarify how this param works.

ogre2/src/Ogre2ParticleNoiseListener.hh
```diff
@@ -51,8 +51,13 @@ namespace ignition
       private: Ogre::MaterialPtr ogreMaterial;
 
       /// \brief Particle scatter ratio. This is used to determine the ratio of
-      /// particles that will detected by sensors
-      private: double particleScatterRatio = 0.65;
+      /// particles that will detected by sensors. Increasing the ratio
```
adlarkin, line 54:
```suggestion
      /// particles that will be detected by sensors. Increasing the ratio
```

test/integration/depth_camera.cc
adlarkin: That makes sense to me. Thanks for the explanation!

ogre2/src/Ogre2ParticleNoiseListener.hh
adlarkin: resolved in https://github.com/ignitionrobotics/ign-rendering/pull/269/commits/0c48d1685d691d45b55f248acfd4813b0fe1f350

src/ParticleEmitter_TEST.cc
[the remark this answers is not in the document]
adlarkin: Good point, I didn't realize we had to add a virtual function. I went ahead and made a note of breaking ABI and needing to do this in an unreleased version in https://github.com/ignitionrobotics/ign-rendering/pull/269/commits/0c48d1685d691d45b55f248acfd4813b0fe1f350

adlarkin (approved):

iche033:
CI is green, merging

PR #275, opened 2021-03-19
Add API for particle emitter scatter ratio

adlarkin:

# 🦟 Bug fix

Addresses a missing API that was mentioned [here](https://github.com/ignitionrobotics/ign-rendering/pull/264#discussion_r596898756). This breaks ABI, so this is targeting Edifice before release.

## Checklist
- [x] Added tests

### DIFF
--- a/ogre2/src/Ogre2ParticleNoiseListener.cc
+++ b/ogre2/src/Ogre2ParticleNoiseListener.cc
@@ -21,6 +21,7 @@
 #include "ignition/rendering/ogre2/Ogre2RenderTypes.hh"
 #include "ignition/rendering/ogre2/Ogre2Scene.hh"
 #include "ignition/rendering/ogre2/Ogre2Visual.hh"
+#include "ignition/rendering/ogre2/Ogre2ParticleEmitter.hh"
 
 #include "Ogre2ParticleNoiseListener.hh"
 
@@ -88,6 +89,7 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
 
       // get particle scatter ratio value from particle emitter user data
       // and pass that to the shaders
+      float scatterRatio = 0.65f;
       Ogre::Any userAny = ps->getUserObjectBindings().getUserAny();
       if (!userAny.isEmpty() && userAny.getType() == typeid(unsigned int))
       {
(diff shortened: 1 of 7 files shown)


ogre2/src/Ogre2ParticleNoiseListener.cc
```diff
@@ -101,48 +102,16 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
...
-            if (ratio > 0)
-            {
-              this->particleScatterRatio = ratio;
-            }
-          }
+          psParams->setNamedConstant("particleScatterRatio",
+              emitterPtr->ParticleScatterRatio());
         }
+        // TODO(adlarkin) set the scatter ratio in shaders to some default
+        // value of emitterPtr is null?
```
adlarkin, line 113: Any thoughts about this? I believe that right now, since we are using uniforms for the `particleScatterRatio` variable in the shaders, this would default to 0 unless otherwise specified:

https://github.com/ignitionrobotics/ign-rendering/blob/41086328e7abd56756f26942d69a8aa70eb4efef/ogre2/src/media/materials/programs/gpu_rays_1st_pass_fs.glsl#L40

If we don't set a default, I think that this would mean that no scattering occurs when `emitterPtr` is null, since this check should result in `false` (this actually might be correct behavior - if so, I don't think we need to handle the case of setting anything in shaders if `emitterPtr` is null):

https://github.com/ignitionrobotics/ign-rendering/blob/41086328e7abd56756f26942d69a8aa70eb4efef/ogre2/src/media/materials/programs/gpu_rays_1st_pass_fs.glsl#L93-L95

In reality, `emitterPtr` shouldn't be null here, but I think it could be worth handling that scenario just in case.

adlarkin (commented):
@iche033 this is ready for review. I just have one question, which I tried to explain below.

ogre2/src/Ogre2ParticleNoiseListener.cc
adlarkin: After looking at your original approach some more, I tried to follow that in https://github.com/ignitionrobotics/ign-rendering/pull/275/commits/945fd09008773629acbc54e2b9d4d7a79d871072. I believe that this could resolve this comment, let me know what you think.

ogre2/src/Ogre2ParticleNoiseListener.cc
```diff
@@ -88,6 +89,7 @@ void Ogre2ParticleNoiseListener::preRenderTargetUpdate(
 
       // get particle scatter ratio value from particle emitter user data
       // and pass that to the shaders
+      float scatterRatio = 0.1f;
```
iche033, line 92: the default value before the change was 0.65 (in ogre2/src/Ogre2ParticleNoiseListener.hh) so I think we should keep that consistent

iche033 (commented):
nice, looks good to me. just one minor comment. 

thanks for adding the new API!

ogre2/src/Ogre2ParticleNoiseListener.cc
adlarkin: Fixed in https://github.com/ignitionrobotics/ign-rendering/pull/275/commits/531ab80c95bd8cf0c85982453f7fc0c64a1ab56d

adlarkin:
We'll need to update the related code in `ign-gazebo` as well, but I think that we might need to wait to do that. In particular, we can now use these new API calls in `ign-gazebo/src/rendering/SceneManager.cc` (https://github.com/ignitionrobotics/ign-gazebo/pull/674), but we will need to wait for that PR to be merged and forward-ported first.

iche033 (approved):

PR #200, opened 2021-01-19
Added capsule geometry

ahcorde:
Added capsule geometry. Related with this issue https://github.com/ignitionrobotics/ign-rendering/issues/197

 - Ogre
capsule_ogre

 - Ogre2
capsule_ogre2

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #276, opened 2021-03-19
Suggestions to #200

chapulina:
* Change `Create` to `Update` to reflect better what the function is used for
* Made `ogre1` and `ogre2` code very similar - the only difference are the `2`s
* Changed the `Update` function so that the existing mesh is updated even when a new mesh doesn't need to be created (in case we add other properties that make `capsuleDirty` without altering the mesh's name)
* Got the markers to work with `ign-gazebo`, opening a PR soon.

### DIFF
--- a/include/ignition/rendering/Capsule.hh
+++ b/include/ignition/rendering/Capsule.hh
@@ -28,9 +28,8 @@ namespace ignition
   {
     inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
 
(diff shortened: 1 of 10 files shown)


chapulina:
I recommend hiding whitespace changes when reviewing, there aren't as many changes as it seems.

iche033:
I think the Capsule_TEST failed due to segfault on shutdown

chapulina:
> I think the Capsule_TEST failed due to segfault on shutdown

I fixed the test by removing the call to destroy from the destructor. I don't know if things are being correctly cleaned up though, we may have memory leaks. Something to revisit after the release.

PR #296, opened 2021-04-01
Reduce lidar data discretization

iche033:

# 🦟 Bug fix

## Summary

For lidars with high sample count, there is apparent discretization in the data generated. The discretization effect is more evident at long distances. This is mainly due to the resolution of the textures used  in the 1st pass, which were hardcoded to 1024x1024. For a lidar with high sample count, e.g. 10000, the 1024 texture size is too low. The 2nd pass uses a texture with the actual sample count (10000) but it's just sampling from this 1024x1024 texture. 

This PR changes the way the 1st pass texture size is set - It is now based on the sample count and field of view of the lidar.  Lidars with low sample count will now also use smaller textures to save memory. The max texture size is capped to 4K in order to run on most graphics cards. 

Here is a before and after comparison of point cloud visualization in RVIZ showing the discretization effect for a 360 degrees lidar with 10000 horizontal samples.

Before: 1st pass texture resolution fixed to 1024
lidar_discretization_before

After: 1st pass texture resolution capped to 4096
lidar_discretization_after
 

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/actor_animation/Main.cc
+++ b/examples/actor_animation/Main.cc
@@ -181,17 +181,17 @@ int main(int _argc, char** _argv)
 
   // Expose engine name to command line because we can't instantiate both
   // ogre and ogre2 at the same time
-  std::string engine("ogre");
+  std::string ogreEngineName("ogre");
   if (_argc > 1)
   {
-    engine = _argv[1];
+    ogreEngineName = _argv[1];
   }
 
   common::Console::SetVerbosity(4);
   std::vector<std::string> engineNames;
   std::vector<CameraPtr> cameras;
 
-  engineNames.push_back(engine);
+  engineNames.push_back(ogreEngineName);
 
   std::vector<VisualPtr> visuals;
   ic::SkeletonPtr skel = nullptr;
--- a/examples/camera_tracking/Main.cc
+++ b/examples/camera_tracking/Main.cc
@@ -118,12 +118,20 @@ int main(int _argc, char** _argv)
 {
   glutInit(&_argc, _argv);
 
+  // Expose engine name to command line because we can't instantiate both
+  // ogre and ogre2 at the same time
+  std::string ogreEngineName("ogre");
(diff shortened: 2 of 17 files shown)


ahcorde (approved):

iche033:
I did some profiling to see the impact of increasing to use a 4096 texture and found that there is some noticeable changes.

For the case of a 360 degrees lidar with 4K textures (worst case), the VRAM usage went up by around 1.5GB. Manual calculation seems to be consistent with observations:

4096:
* depth texture: 4096 * 4096 * 4 (bytes per channel) * 3 (no. of channels per by pixel) * 6 (cube faces) = 1208 MB
* particle texture: 2048 * 2048 * 4 * 3 * 6 = 302 MB
* total: ~1.5GB

1024:
* depth texture: 1024 * 1024 * 4 * 3 * 6 = 75 MB
* particle texture: 512 * 512 * 4 * 3 * 6 = 19 MB
* total: < 100 MB

There is also some noticeable performance hit in the `PostRender` call:
* 4096: ~0.02s
* 1024: ~0.002s

-------------------

I did one more test with a texture size limit of 2048 and the results are:
* VRAM: ~400MB
* `PostRender`: ~0.0045s

Here's the lidar visualization of using 2048 sized textures:

lidar_discretization_2048

-------------------

iche033:
Based on the profiling results above, I'm leaning towards capping the texture size to 2048 (or keeping in as 1024) instead of 4096 to avoid a significant change in performance in a released version of ign-rendering

iche033:
After some discussion offline, we decided to keep the limit to `1024` due to the impact it has on VRAM usage. 42f7466

So the changes here would no longer reduce lidar discretization for sensors with high sample count. I think the changes are still worth getting int because it makes the 1st pass texture resolution configurable based on sample size and so reduces VRAM usage for lidars with low sample count.

PR #299, opened 2021-04-03
Improve build times by reducing included headers

darksylinc:
## Summary

Ogre2Includes.hh is a very convenient header but it includes too many
Ogre headers, which heavily affects build times.

For certain .cc files compilation time was reduced by 1.2 seconds on an
i7 7700 @3.60Ghz

A full build was of ign-rendering5 went from 1m 57s to 1m 46s (I only
touched ogre2, not ogre; a full rebuild of ign-rendering5 includes
both; optix was not included in my config).

This amounts to an 11 second improvement for a full build meaning a 9.4%
reduction (more if we consider ogre1 module is affecting measuring)

The solution is to simply include the Ogre headers that are needed, instead
of relying on Ogre2Includes.hh; and making sure Ogre2Includes.hh is never
included in a *.hh file.

## Checklist

This is a simple header inclusion change; it does not alter runtime functionality.
Note that merging this change into other branches such as the WIP 2.2 might cause a few compiler errors as there are fewer headers included. But those build errors should be trivial fix.

### DIFF
--- a/ogre2/src/Ogre2DynamicRenderable.cc
+++ b/ogre2/src/Ogre2DynamicRenderable.cc
@@ -32,6 +32,18 @@
 #include "ignition/rendering/ogre2/Ogre2RenderEngine.hh"
 #include "ignition/rendering/ogre2/Ogre2Scene.hh"
 
+#ifdef _MSC_VER
+  #pragma warning(push, 0)
+#endif
+#include <OgreItem.h>
+#include <OgreMesh2.h>
+#include <OgreMeshManager2.h>
+#include <OgreSceneManager.h>
+#include <OgreSubMesh2.h>
+#ifdef _MSC_VER
+  #pragma warning(pop)
+#endif
+
 /// \brief Private implementation
 class ignition::rendering::Ogre2DynamicRenderablePrivate
 {
@@ -131,8 +143,8 @@ void Ogre2DynamicRenderable::DestroyBuffer()
   if (this->dataPtr->vbuffer)
     delete [] this->dataPtr->vbuffer;
 
-  Ogre::Root *root = Ogre2RenderEngine::Instance()->OgreRoot();
-  Ogre::RenderSystem *renderSystem = root->getRenderSystem();
+  Ogre::RenderSystem *renderSystem =
+      this->dataPtr->sceneManager->getDestinationRenderSystem();
   Ogre::VaoManager *vaoManager = renderSystem->getVaoManager();
 
   if (!vaoManager)
@@ -187,8 +199,8 @@ void Ogre2DynamicRenderable::UpdateBuffer()
   if (!this->dataPtr->dirty)
     return;
 
-  Ogre::Root *root = Ogre2RenderEngine::Instance()->OgreRoot();
-  Ogre::RenderSystem *renderSystem = root->getRenderSystem();
+  Ogre::RenderSystem *renderSystem =
+      this->dataPtr->sceneManager->getDestinationRenderSystem();
 
   Ogre::VaoManager *vaoManager = renderSystem->getVaoManager();
   if (!vaoManager)
--- a/ogre2/include/ignition/rendering/ogre2/Ogre2Capsule.hh
+++ b/ogre2/include/ignition/rendering/ogre2/Ogre2Capsule.hh
@@ -23,7 +23,6 @@
 
 #include "ignition/rendering/base/BaseCapsule.hh"
 #include "ignition/rendering/ogre2/Ogre2Geometry.hh"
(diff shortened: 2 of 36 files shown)


darksylinc:
I had to resubmit the PR because it was accidentally done onto the wrong branch, which triggered all CIs to fail.

ogre2/src/Ogre2DynamicRenderable.cc
```diff
@@ -131,8 +137,8 @@ void Ogre2DynamicRenderable::DestroyBuffer()
   if (this->dataPtr->vbuffer)
     delete [] this->dataPtr->vbuffer;
 
-  Ogre::Root *root = Ogre2RenderEngine::Instance()->OgreRoot();
-  Ogre::RenderSystem *renderSystem = root->getRenderSystem();
+  Ogre::RenderSystem *renderSystem =
+      this->dataPtr->sceneManager->getDestinationRenderSystem();
```
chapulina, line 141: This change seems unrelated to reducing the headers, is it needed as part of this PR?

chapulina (commented):
Nice! I just have a question about a non-header change.

ogre2/src/Ogre2DynamicRenderable.cc
AllianceDarkSylinc: Actually yes. Including OgreRoot.h is not necessary because RenderSystem can be obtained from another pointer

darksylinc: Actually yes. Including OgreRoot.h is not necessary because RenderSystem can be obtained from another pointer

chapulina:
This PR is exposing some compiler warnings on Windows coming from upstream. We had suppressed those on [206](https://github.com/ignitionrobotics/ign-rendering/pull/206/files#diff-17b3ed8084f89242c1a1dd9d3f6ec0d28a9998f8773561f92f9f544741ff2613) with pragmas on `Ogre2Includes.hh`. Now that we're using more granular includes, I think we'll need to suppress the warnings separately on each file.

darksylinc:
Ahh I see.

I'll have to compile on Linux and see those warnings. If they're trivial I may be able to fix upstream

chapulina:
> I'll have to compile on Windows and see those warnings

You should be able to see them on CI here:

https://build.osrfoundation.org/job/ign_rendering-pr-win/1547/msbuild/

> If they're trivial I may be able to fix upstream

That would be awesome to get on the 2.2 version that we're about to build against, if possible :smiley:

iche033:
> That would be awesome to get on the 2.2 version that we're about to build against, if possible ￼ 

+1

since ign-rendering4 depends on 2.1, we'll probably still need to add those pragma's in this PR around places where we include ogre headers.

darksylinc:
- I've re-targetted the PR to ign-rendering5 as requested
- Added #pragmas to disable warnings coming from Ogre headers in MSVC
- Fixed warnings [in upstream](https://github.com/OGRECave/ogre-next/commit/32ba36e8749bde782ccacfff67be1c3261c6cf81)
- I don't know why MSVC claims [these 5 new warnings](https://build.osrfoundation.org/job/ign_rendering-pr-win/1557/msbuild/new/) given that these warnings should've been in the previous version as well.

iche033:
For the 5 windows warnings, I think we can just give it the same treatment as we did in the ogre 1.x case:
https://github.com/ignitionrobotics/ign-rendering/blob/cd6e89f0e528b27955bc33c4e37ad7f3b9df6e25/ogre/include/ignition/rendering/ogre/OgreRenderTargetMaterial.hh#L29

here's a diff

```
diff --git a/ogre2/include/ignition/rendering/ogre2/Ogre2RenderTargetMaterial.hh b/ogre2/include/ignition/rendering/ogre2/Ogre2RenderTargetMaterial.hh
index e2e4cbd3..a5fbb197 100644
--- a/ogre2/include/ignition/rendering/ogre2/Ogre2RenderTargetMaterial.hh
+++ b/ogre2/include/ignition/rendering/ogre2/Ogre2RenderTargetMaterial.hh
@@ -31,6 +31,11 @@
   #pragma warning(pop)
 #endif
 
+#ifdef _MSC_VER
+ #pragma warning(push)
+ #pragma warning(disable:4275)
+#endif
+
 namespace ignition
 {
   namespace rendering
@@ -106,4 +111,8 @@ namespace ignition
   }
 }
 
+#ifdef _MSC_VER
+ #pragma warning(pop)
+#endif
+
 #endif
```

darksylinc:
Updated again. The windows CI now passes!

chapulina (approved):
Nice!

iche033 (approved):

PR #303, opened 2021-04-10
Add shadows to Ogre2DepthCamera without crashing

darksylinc:
Ogre2Scene::PreRender was using a hack via Ogre2RenderTexture::SetCamera
to get compositors rebuilt when lights (and hence shadow mapping)
changed. This is now done properly.

The code changes seem large but they're actually not:
- Ogre2RenderTarget::UpdateShadowNode has been moved to Ogre2Scene (cut
and paste)
- Public interface's Camera::SetShadowsNodeDefDirty was added
- Ogre2Camera and Ogre2DepthCamere overload SetShadowsNodeDefDirty
- Private Ogre2DepthCamera::CreateWorkspaceInstance function was added

## Summary

Before this commit, RGBD cameras did not have shadows, because enabling it would cause crashes or memory corruption.

After this commit, RGBD cameras have shadows, and should be safe to keep enabled.

I was asked to port this changes to ign-rendering4, however I could not make it work because even without my changes, RGBD cameras' outputs in normal ign-rendering4 was garbage. You can view my attempt to port to ign-rendering4 in [matias-depthcamera-shadow-fix4](https://github.com/darksylinc/ign-rendering/tree/matias-depthcamera-shadow-fix4).

Therefore this PR is made against ign-rendering5

## Checklist


include/ignition/rendering/Camera.hh
```diff
@@ -282,6 +282,12 @@ namespace ignition
       /// \return Texture Id of type GLuint.
       public: virtual unsigned int RenderTextureGLId() const = 0;
 
+      /// \brief Notifies us that the shadow node definition is about to be
+      /// updated. This means our compositor workspace must be destroyed
+      /// because the shadow node definition it's using will become a
+      /// dangling pointer otherwise
+      public: virtual void SetShadowsNodeDefDirty() = 0;
```
iche033, line 289: in ignition we guarantee ABI compatibility for released versions. So adding a virtual function here unfortunately breaks ABI. Can you move this function to Ogre2Camera and Ogre2DepthCamera (I think those are the only 2 classes that need them) and make them non virtual? Then in `Ogre2Scene::PreRender`, we should be able to just call the derived class' `SetShadowsNodeDefDirty` directly.

ogre2/include/ignition/rendering/ogre2/Ogre2RenderTarget.hh
```diff
@@ -134,10 +137,6 @@ namespace ignition
       /// \brief Update the render pass chain
       protected: virtual void UpdateRenderPassChain();
 
-      /// \brief Create a compositor shadow node with the same number of shadow
-      /// textures as the number of shadow casting lights
-      protected: void UpdateShadowNode();
```
iche033, line 139: We'll need to deprecate this function first before removing it since it's exposed to the users. Can you mark this function as deprecated?

```
protected: void IGN_DEPRECATED(5) UpdateShadowNode();
```

The implementation can just be a warning msg, something like:

```
ignwarn << "Ogre2RenderTarget::UpdateShadowNode() is deprecated and replaced by Ogre2Scene:UpdateShadowNode()" << std::endl;
```

iche033 (commented):
Nice I can see shadows in an RGBD camera now! I have couple of comments about ABI compatibility and deprecating a function.

One issue I noticed is that when I try to add a new shadow-casting light (to trigger a rebuild of compositor), ign gazebo crashes. This can be reproduced with the following steps:
1. run the [sensors_demo.sdf](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/examples/worlds/sensors_demo.sdf) world: `ign gazebo -v 4 -r sensors_demo.sdf`
2. open the `Lights` plugin by going to top right menu  (three dots) and select `Lights`
3. insert a spot light into the scene (needs to be spot light as we disabled point light shadows due to another shadows issue)
4. ign gazebo crashes

include/ignition/rendering/Camera.hh
darksylinc: Ouch!!! (Swears internally)

Then we can implement this version on unstable branch, while stable uses downcasting and calling the function directly

darksylinc: How do you normally handle this?

#ifdefs e.g.

```
#if IGN_VERSION > 4
public: virtual void SetShadowsNodeDefDirty() = 0;
#endif

#if IGN_VERSION > 4
camera->SetShadowsNodeDefDirty();
#else
 // downcast version
#endif
```

or just each branch its own piece of code?

iche033: > or just each branch its own piece of code?


yeah we'll need to have different code. The process goes like:
1. add downcast version in ign-rendering5 + add a note in appropriate places, e.g. in Ogre2Camera and Ogre2DepthCamera's `SetShadowsNodeDefDirty` function to mention that this is done for ABI compatibility.
2. When this PR gets merged, someone (e.g. me) will do a merge forward from ign-rendering5 to the unstable branch, `main`
3. During the merge forward, I can take care of changing the implementation to use this version

ogre2/include/ignition/rendering/ogre2/Ogre2RenderTarget.hh
darksylinc: > One issue I noticed is that when I try to add a new shadow-casting light (to trigger a rebuild of compositor), ign gazebo crashes. This can be reproduced with the following steps:

Cannot reproduce. Adding new lights works perfectly fine for me. I'm even trying on full debug. Do you have a callstack?

Perhaps something didn't re-build cleanly, since as you said the ABI changed and if a component wasn't properly rebuilt, at some point the wrong virtual function(s) will be called.

darksylinc: Nevermind. Repro found.

darksylinc: Well I am shocked!

We hit an Ogre 2.1 bug that was fixed in 2.2; fortunately it's easy to workaround from Gazebo, and that workaround can be removed in 2.2

darksylinc:
OK I've updated the PR with the following fixes:

1. Rebased to latest head of branch
2. Fix crash on the server after adding a spotlight
3. Avoid breaking the ABI. Commit https://github.com/darksylinc/ign-rendering/commit/0585a30b403d1c8799cb9cc34aac7658757faebd (the one with `#SetShadowsNodeDefDirtyABI`) should be reverted when merging this forward to newer versions

ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
```diff
@@ -118,6 +122,28 @@ namespace ignition
...
+      // necessary there
+      //
+      // See https://github.com/ignitionrobotics/ign-rendering/
+      // pull/303#pullrequestreview-635228897
+      // Repro:
+      //  1.run the sensors_demo.sdf world: ign gazebo -v 4 -r sensors_demo.sdf
+      //  2.open Lights plugin, top right menu (3 dots) and select Lights
+      //  3.insert a spot light into the scene (needs to be spot light)
+      //  4.ign gazebo crashes
+      private: void removeWorkspaceCrashWorkaround();
```
iche033, line 145: can you change capitalize this function, i.e. `RemoveWorkspaceCrashWorkaround`? private functions are also capitalized in Igntiion.

ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -304,6 +304,7 @@ void Ogre2DepthCamera::Destroy()
   }
   if (this->dataPtr->ogreCompositorWorkspace)
   {
+    removeWorkspaceCrashWorkaround();
```
iche033, line 307: add `this->` in front of member function names

ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -1154,6 +1165,38 @@ double Ogre2DepthCamera::FarClipPlane() const
...
+void Ogre2DepthCamera::SetShadowsNodeDefDirty()
+{
+  if (!this->dataPtr->ogreCompositorWorkspace)
+    return;
+
+  auto engine = Ogre2RenderEngine::Instance();
+  auto ogreRoot = engine->OgreRoot();
+  Ogre::CompositorManager2 *ogreCompMgr = ogreRoot->getCompositorManager2();
+
+  removeWorkspaceCrashWorkaround();
```
iche033, line 1178: add this-> in front of member function names

iche033 (commented):
Nice! the workaround fixes the crash for me. Changes look good. I just have very minor coding style comments. Otherwise this is good to go.

ogre2/include/ignition/rendering/ogre2/Ogre2DepthCamera.hh
darksylinc: Aggh, sorry. Force of habit.

I'm aware of this style but I've unconsciously wrote it that way. I'll fix them

darksylinc:
Coding style changes done

iche033 (approved):

PR #307, opened 2021-04-18
Fix race condition when rendering the UI (#304)

### LINKED ISSUE #304: Race condition between Qt and Ogre2 in presentation
ign-rendering uses the same pattern as provided [textureinthread](https://doc.qt.io/archives/qt-5.11/qtquick-scenegraph-textureinthread-example.html) sample from Qt. (the code can be [seen in repo](https://github.com/qt/qtdeclarative/tree/5.15/examples/quick/scenegraph/textureinthread))

## What should happen

The pattern is the following (ideal, i.e. what should happen):

1. Worker Thread (e.g. Ogre2) is producer
2. GUI thread (Qt) is consumer
3. There are two FBOs, A and B

```
1. Qt requests Ogre2 to start rendering
2. Ogre renders to FBO B, signals it's done to Qt thread
3. Qt thread atomically swaps A and B
4. Qt thread presents B
5. Qt requests Ogre2 to start rendering
6. Ogre renders to FBO A, signals it's done to Qt thread
7. Qt thread atomically swaps A and B
8. Qt thread presents A
9. Repeat step 1
```

This is basically a double-buffer scheme maintained by hand.

It is even described in Qt's sample (emphasis mine):
QtSample

## What is actually happening

1. Qt requests Ogre2 to start rendering
2. Ogre renders to FBO A, signals it's done to Qt thread
3. Qt thread presents A
4. Repeat step 1

This is causing a race condition because A is both being rendered to while also being used internally as a texture by Qt. This can cause all sorts of weird behaviors or maybe even crashes when Qt UI is used.

Most likely how bad this bug is depends on which driver is being used (e.g. Mesa radeonsi vs Mesa Intel, vs Windows Intel, vs proprietary AMD, vs NVIDIA proprietary, vs Nouveau) and the workload balance and processing speed of the CPU (like all race conditions)

## Why it went unnoticed

This bug went unnoticed because Gazebo defaults to using 8x MSAA (but due to bugs in setup, MSAA is not actually used, that's another problem) and the FBO passed to Qt is the resolved one. Due to this, the window for the race condition is extremely small (because FBO is only written to during the resolve operation, which is only a fraction of the whole render).

However once I disable MSAA the problem becomes much more apparent, notice the background flicker:

https://user-images.githubusercontent.com/3395130/114423489-9a847c80-9b8d-11eb-9398-8233669468a2.mp4

## Environment
* OS Version: All affected
* Source or binary build?
 Source. Probably all branches where Ogre2 is used

## Is Ogre1 affected?

I don't know. I did not evaluate. The possibility is high, given that part of the buggy behavior comes from ign-gui, which is agnostic to both Ogre1 and Ogre2

## Is OptiX affected?

Unknown.

## Tasks

 - [x] ~Ensure FBO juggling is thread safe (https://github.com/ignitionrobotics/ign-rendering/pull/307) ~. We're now forcing serialization. See https://github.com/ignitionrobotics/ign-gazebo/pull/774
 - [x] ~Add "debug mode" for blocking (serializing) Qt and Ogre2 threads for decent RenderDoc debugging (otherwise API calls become interleaved and is extremely confusing)~-  We're now forcing serialization.
 - [x] Fix compositor setup regarding textures. See https://github.com/ignitionrobotics/ign-rendering/pull/313
 - [x] Fix MSAA usage. See https://github.com/ignitionrobotics/ign-rendering/pull/313
 - [x] Stretch goal: Add single threaded mode (some API analyzer tools can be picky). Update: This is no longer optional. See https://github.com/ignitionrobotics/ign-gazebo/pull/774

## Affected files

Most of the bugs affect the files coming from:

 - Ogre2RenderTarget.cc
 - Ogre2Camera.cc (since Ogre2Camera::RenderTextureGLId redirects to Ogre2RenderTarget's GLId)
 - Scene3D.cc from [ign-gui](https://github.com/ignitionrobotics/ign-gui)

# Other

I will be the one fixing this bug. This ticket is for tracking progress

darksylinc:
## Summary

This fix also depends on on a fix in ign-gazebo so that it now can call
SwapFromThread, otherwise the race condition will still happen.

At the current moment, the implementation is very wasteful creating many
more RenderTargets than needed (because 2 workspaces are created).
This issue will be fixed in the next commit

Affects ignitionrobotics/ign-rendering#304

# 🦟 Bug fix

Fixes #304 

## Checklist

### DIFF
--- a/include/ignition/rendering/Camera.hh
+++ b/include/ignition/rendering/Camera.hh
@@ -279,8 +279,13 @@ namespace ignition
       /// \brief Get the OpenGL texture id associated with the render texture
       /// used by this camera. A valid id is returned only if the underlying
       /// render engine is OpenGL based.
-      /// \return Texture Id of type GLuint.
+      /// \return Texture Id of type GLuint. Returned value is
+      /// valid until the next SwapFromThread() call
       public: virtual unsigned int RenderTextureGLId() const = 0;
+
+      /// \brief Informs this Camera we're done updating from worker thread
+      /// and for this iteration of the loop
+      public: virtual void SwapFromThread() = 0;
 
       /// \brief Add a render pass to the camera
       /// \param[in] _pass New render pass to add
--- a/include/ignition/rendering/RenderTarget.hh
+++ b/include/ignition/rendering/RenderTarget.hh
@@ -93,6 +93,10 @@ namespace ignition
       /// \return Render pass at the specified index
       public: virtual RenderPassPtr RenderPassByIndex(unsigned int _index)
           const = 0;
+
+      /// \brief Informs this RenderTarget we're done updating from worker
+      /// thread and for this iteration of the loop
+      public: virtual void SwapFromThread() = 0;
     };
 
     /* \class RenderTexture RenderTexture.hh \
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -165,6 +165,9 @@ namespace ignition
       public: virtual unsigned int RenderTextureGLId() const override;
 
       // Documentation inherited.
+      public: virtual void SwapFromThread() override;
+
+      // Documentation inherited.
       public: virtual void AddRenderPass(const RenderPassPtr &_pass) override;
 
       // Documentation inherited.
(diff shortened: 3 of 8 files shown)


chapulina:
Hi @darksylinc, are you still planning to work on this PR? Let us know how we can help. Thanks!

darksylinc:
Hi!

Not at the time. The race condition was fixed by another PR.

The PR that fixed it forced serialization (which is like brute forcing this thing). This PR wanted to preserve parallelism.

However there are several issues unaddressed (synchronizing important events like window resizes, fixing how Qt/Ign threads pass around data to avoid potential deadlocks), our efforts are being focused on higher priority tasks. What we have now is "good enough"

PR #313, opened 2021-05-01
Fix FSAA in UI and lower VRAM consumption

darksylinc:
## PR NOTE 1

This fixes a regression from ign-rendering4.
In ign-rendering4 FSAA worked fine because it was relying on compositor scripts. This broke when the compositor script was replaced with C++ code.

The VRAM consumption issue affects both branches, but I think it's just fine to make it ign-rendering5 and onwards only

## PR NOTE 2 (important)

~How do I test the camera gaussian filter? It's affected directly (i.e. if renderTargetResultsIdx is set to the wrong value then the image won't look like the filter was applied, or some other similar issue) and I had no idea how to test it.~

**Update:** ~CI says gaussian filter test failed. How shocking. I should check that test locally~

**Update 2:** Found the bug and fixed it. ~Though DepthGaussianNoise test is still failing and I'm still looking into it~

**Update 3:** DepthGaussianNoise is failing in my machine because the test is buggy. Opened Ticket https://github.com/ignitionrobotics/ign-rendering/issues/315

## PR NOTE 3

Affected classes are mostly Ogre2RenderTarget, Ogre2RenderTexture and Ogre2DepthCamera

## PR NOTE 4

Currently fixing ABI breakage.

## Before vs After

Before
After

## Comments

FSAA was being requested however due to how the compositor was setup,
this effect was not taking effect.

Additionally, the Compositor setup was improved to lower memory
consumption. Originally the setup was taken from Ogre samples which
assume they will ultimately be rendering to a window.

However this is not the case and thus IGN was creating 3 render targets
(two for ping-ponging between postprocess FXs + one for storing the
final result)
This was optimized so that we only create 2 render targets: two for
ping-ponging between postprocess FXs and we pick at runtime which one is
storing the final result via the new variable renderTargetResultsIdx

Further performance optimizations could be made in Ogre 2.2 to improve
unnecessary MSAA resolving when doing postprocess, though considering
there's currently only one postprocessing effect (the Gaussian filter) I
doubt this optimization would make much of a difference

## Checklist


darksylinc:
Converting this PR back to DRAFT since multiple issues were found

darksylinc:
I fixed the issues I found so it's no longer a draft again.

However I wanted to see if the ABI checker would tell me ~what I broke~ if I broke something, but the ABI checker seems to be malfunctioning :(

Are there instructions on how to run it locally?

iche033:
> However I wanted to see if the ABI checker would tell me what I broke if I broke something, but the ABI checker seems to be malfunctioning :(

something is wring with this particular build machine. I've triggered the build again on a different machine:
https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/569/

> Are there instructions on how to run it locally?

We don't have public instructions I think. We use [abi-compliance-checker](https://lvc.github.io/abi-compliance-checker/). You can probably extract the cmds that we use by looking at the console output of one of the successful ABI builds:
https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/564/console

```
abi-compliance-checker -lang C++ -lib ign-rendering -old pkg.xml -new devel.xml
```

The contents of `pkg.xml` and `devel.xml` are also in the console output.

darksylinc:
Apparently `isRenderWindow` is the only function that broke ABI. So I fixed it.

But I could use a 2nd pair of eyes.

If everything's ok it should be ready for merge

iche033:
can you take a look at the latest ABI report from our CI? It found a few more issues:
https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/577/API_5fABI_20report/

Usually the `Added Symbols` issues are harmless. It's mainly the `Removed Symbols` and `Problems with Data Types` that cause ABI incompatibility.

ogre2/include/ignition/rendering/ogre2/Ogre2RenderTarget.hh
```diff
@@ -115,8 +115,16 @@ namespace ignition
...
       public: void SetShadowsNodeDefDirty();
 
-      /// \brief Get a pointer to the ogre render target
-      public: virtual Ogre::RenderTarget *RenderTarget() const = 0;
+      /// \brief Returns true if this is a render window
+      /// TODO(anyone): this function should be virtual.
+      /// We didn't do it to preserve ABI.
+      /// Look in commit history for '#Ogre2IsRenderWindowABI' to
+      /// see changes made and revert
+      public: bool isRenderWindow() const;
```
iche033, line 123: capitalize `I`, i.e. `IsRenderWindow`

ogre2/include/ignition/rendering/ogre2/Ogre2RenderTarget.hh
```diff
@@ -133,7 +141,10 @@ namespace ignition
           Ogre::CompositorWorkspace *_workspace,
           const std::string &_workspaceDefName,
           const std::string &_baseNode, const std::string &_finalNode,
-          const std::vector<RenderPassPtr> &_renderPasses, bool _recreateNodes);
+          const std::vector<RenderPassPtr> &_renderPasses,
+          bool _recreateNodes,
+          Ogre::Texture *(*_ogreTextures)[2],
+          bool _isRenderWindow);
```
iche033, line 147: add doxygen comments for these new params

we'll need to keep the existing function since adding the args here breaks API/ABI.

darksylinc:
> Usually the `Added Symbols` issues are harmless. It's mainly the `Removed Symbols` and `Problems with Data Types` that cause ABI incompatibility.

Mmm....

`Ogre2RenderTarget::UpdateRenderPassChain` changed signature, so it counts as 1 removal, 1 add.
I can keep two copies, deprecated the old one, and make it throw (since the new code can**not** work without the missing arguments). My questions are:

1. If client application is directly calling Ogre2RenderTarget::UpdateRenderPassChain, it will throw. Does this count as ABI breakage? The client app at this point was interfacing directly with a non-virtual function and AFAIK this is not guaranteed (by Gazebo) to stay stable.
2. Or is this just to fix potential unrelated ABI breakage? (i.e. removal of function causes unknown crashes in unrelated functions)

Removal of `Ogre2RenderTexture::RenderTarget` is controversial, because it is a virtual function. What happened was that this:

```
class Ogre2RenderTarget
{
  virtual OgrePointer* RenderTarget() = 0; // was pure
};

class Ogre2RenderTexture : public Ogre2RenderTarget
{
  virtual OgrePointer* RenderTarget() override;
};
```

Became this:

```
class Ogre2RenderTarget
{
  virtual OgrePointer* RenderTarget(); // no longer pure
};

class Ogre2RenderTexture : public Ogre2RenderTarget
{
  // Nothing here
};
```

From a vtable point of view nothing relevant changed (calling Ogre2RenderTarget::RenderTarget() should still work), but trying to directly call Ogre2RenderTexture::RenderTarget() (i.e. bypassing the vtable) would fail.

I can play on the safe side here and reimplement Ogre2RenderTexture::RenderTarget.

Btw I'm a bit worried about this error:

1. `The layout of v-table has been changed for unknown reason.	Call of any method in this class may result in crash or incorrect behavior of applications.` is this normal?

iche033:
> If client application is directly calling Ogre2RenderTarget::UpdateRenderPassChain, it will throw. Does this count as ABI breakage? The client app at this point was interfacing directly with a non-virtual function and AFAIK this is not guaranteed (by Gazebo) to stay stable.

We can deprecate old function but instead of making it throw, just print a warning msg to tell users to use the new function (Ignition has a no throw policy which is probably not well documented). We can do something like this to prevent the warning msg being printed out repeatedly: https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/include/ignition/gazebo/components/Component.hh#L124-L132

> Or is this just to fix potential unrelated ABI breakage? (i.e. removal of function causes unknown crashes in unrelated functions)

In the `UpdateRenderPassChain` particular case, I think it would be less of a problem. Other issues related to vtable changes and data types found by the ABI build is likely to cause unknown / undefined behavior and crashes. 

> The layout of v-table has been changed for unknown reason. Call of any method in this class may result in crash or incorrect behavior of applications

We've seen this before and the consensus is that it that we need to fix this issue to avoid unexpected crashes. My guess is that it's related to changes done to the `RenderTarget()` function. 

> I can play on the safe side here and reimplement Ogre2RenderTexture::RenderTarget

This is one thing I would try first to see if it helps fix the v-table issue.

The ABI checker also picked up issues with the new `ogreTexture[2]` var and the change to existing variable `materialApplicator` which caused changes in memory layout.

Fixing ABI is not fun and usually involves many workarounds :/ I can also take a look at fixing some of these ABI issues later this week if you haven't gotten to it yet.

darksylinc:
I updated the code to be ABI compatible.

Maaan I'm beat. This is 20% work, 80% figuring out ABI issues.
If this keeps up I'll just target a newer version.

My ABI checker says 100% compatibility. Let's cross fingers CI agrees.

iche033 (commented):
this should get rid of the build warnings:

```
diff --git a/ogre2/src/Ogre2RenderTarget.cc b/ogre2/src/Ogre2RenderTarget.cc
index a0e8fda6..59b66397 100644
--- a/ogre2/src/Ogre2RenderTarget.cc
+++ b/ogre2/src/Ogre2RenderTarget.cc
@@ -647,10 +647,11 @@ void Ogre2RenderTarget::UpdateRenderPassChain()
 
 void Ogre2RenderTarget::UpdateRenderPassChain(
-    Ogre::CompositorWorkspace *_workspace, const std::string &_workspaceDefName,
-    const std::string &_baseNode, const std::string &_finalNode,
-    const std::vector<RenderPassPtr> &_renderPasses,
-    bool _recreateNodes)
+    Ogre::CompositorWorkspace * /*_workspace*/,
+    const std::string & /*_workspaceDefName*/,
+    const std::string & /*_baseNode*/, const std::string & /*_finalNode*/,
+    const std::vector<RenderPassPtr> & /*_renderPasses*/,
+    bool /*_recreateNodes*/)
 {
   ignwarn << "Warning: This Ogre2RenderTarget::UpdateRenderPassChain "
           << "overload is deprecated" << std::endl;
@@ -856,9 +857,12 @@ void Ogre2RenderTarget::SyncOgreTextureVars()
 // Ogre2RenderTexture
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 Ogre2RenderTexture::Ogre2RenderTexture()
 {
 }
+#pragma GCC diagnostic pop
 
 Ogre2RenderTexture::~Ogre2RenderTexture()
@@ -917,7 +921,10 @@ Ogre::RenderTarget *Ogre2RenderTexture::RenderTarget() const
 void Ogre2RenderTexture::SetOgreTexture(Ogre::Texture *_ogreTexture)
 {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
   this->ogreTexture = _ogreTexture;
+#pragma GCC diagnostic pop
 }
 
 /////////////////////////////////////////////////
```

iche033:
Confirmed that the FSAA issue is fixed. Thanks for iterating! 

> Maaan I'm beat. This is 20% work, 80% figuring out ABI issues.
> If this keeps up I'll just target a newer version.

sure, if the changes are looking like they are not trivial to make to a release version without breaking ABI, you can target `main` first. We would then evaluate afterwards to see if it's worth backporting them.

darksylinc:
> looking at the data type problem description and the vtable, I think the issue could be due to Ogre2RenderTarget::RenderTarget() no longer being pure virtual. I don't think there's a behavior change since the logic is just moved to the base class and we are not re-implementing a virtual function. This may be ok.

I suspect the same. It passed on my machine, but it seems it's not consistent. Just to be sure I made it pure virtual again and let's see what the abi checker says this time.

iche033:
latest ABI build is green :)

the INTEGRATION_render_pass test failed in the latest ubuntu build. I haven't seen that before. I just retriggered another [build](https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/1460/) and it failed again. I haven't been able to reproduce it locally - could be a timing issue. This led me to try out running the [ogre2_demo](https://github.com/ignitionrobotics/ign-rendering/tree/bd4aa2b1dfaf0861cf5b3b613166bfec0fb42815/examples/ogre2_demo) in examples again and noticed that the mouse movement is laggy / jumpy. See gif below. I don't see this happen on `ign-rendering5` branch. If I enable the noise render pass (post processing) effect by pressing `P`, the lag is gone, not sure why

ign-rendering-fsaa-lag

darksylinc:
> the INTEGRATION_render_pass test failed in the latest ubuntu build
It's been failing randomly (but quite frequently) on my machine. I was gonna take a look at some point as I feared it could be a problem with the PostRender branch, but interesting to see that it repros here instead!

It may mean that this patch accidentally tries to show the wrong texture, containing uninitialized or old results, which would explain why the error fails (i.e. it contains whatever was in memory)

> This led me to try out running the ogre2_demo in examples again and noticed that the mouse movement is laggy / jumpy

This further fuels this theory, if the same texture is shown twice. Definitely worth taking a look. Thanks!

darksylinc:
I'm afraid I couldn't repro any of your problems:

1. INTEGRATION_render_pass passes
2. ogre2_demo does not lag

I tried ogre2_demo on the following machines:

1. AMD Ryzen 5900X + AMD Radeon RX 560 2GB (mesa drivers) + Ubuntu 18.04 LTS (5.11.0-051100-generic)
2. Intel i7 7700 + AMD Radeon HD 7770 (mesa drivers) + Ubuntu 20.04 LTS (5.8.0-50-generic)
3. Intel i7 7700 + NVIDIA GeForce 1060 3GB (proprietary drivers) + Ubuntu 20.04 LTS (5.8.0-50-generic)

I noticed the Intel i7 machine keeps randomly complaining about a double free / mem. corruption (sometimes it complains at startup and crashes before the windows comes up, other times it complains during shutdown, other times it doesn't complain at all) but that may have been some build mistake which I am going to make a clean build now to be sure. If problem persists it may be worth looking into for me.

But overall I could not repro none of the two problems :(

darksylinc:
@iche033 Although I could not repro the bug you mentioned, I fixed a heap corruption that could _definitely_ cause the behavior you were seeing on your machine. Could you try again?

As for the other two bugs I found, they're unrelated to my fixes so I'll create a PR. I think I'll split it into two more PRs because both are simple, but one of them may break ABI (I'm not sure how we can avoid ABI breakage there, the root of the problem is in the syntax declaration)

iche033:
ok thanks for looking into the `ogre2_demo` problem and fixing the memory issues.

I did more testing and found that I was also able to reproduce the problem in latest `ign-rendering4` branch so it's not introduced by this PR, sorry for the noise! I tested with the 2 memory fixes but the problem is still there. I ticketed #341 with more info added there. I think it came down to high GPU usage that's causing the lag. Surprisingly, the GPU usage drops when post-processing effect enabled - I would have thought that it would go up instead of down. 

In any case, since it's not introduced by this PR, the changes look good here. The `INTEGRATION_render_pass` test is also passing in the last two builds. I think this is good to go.

iche033 (approved):

darksylinc:
Yyyyeeeeessss!!

This PR was starting to drive me a little nuts.

> I think it came down to high GPU usage that's causing the lag

Ahh!!! This reminds me that I noticed `Ogre2RenderTarget::UpdateRenderPassChain` ended up calling `recreateAllNodes` every frame. It didn't seem like that's intended behavior because it should notice nothing changed and hence early out.

That could be causing the lag. I was going to investigate that, then forgot.

iche033:
> Ahh!!! This reminds me that I noticed Ogre2RenderTarget::UpdateRenderPassChain ended up calling recreateAllNodes every frame. 

that's it! I've created a PR to fix it: https://github.com/ignitionrobotics/ign-rendering/pull/342

darksylinc:
> that's it! I've created a PR to fix it: #342

Awesome! I would've probably taken longer to figure that out

PR #320, opened 2021-05-08
Helper function to get a scene

chapulina:
# 🎉 New feature

## Summary

Most applications will only have one rendering engine loaded at a time, and only one scene within that. This helper function gets the first scene that can be found in any rendering engine.

This pattern is being repeated on several `ign-gui` and `ign-gazebo` plugins, so it would be nice to reduce duplication. For example:

* [Screenshot](https://github.com/ignitionrobotics/ign-gui/blob/ffbdb26ce519f85b273d57527a11470dae3838e1/src/plugins/screenshot/Screenshot.cc#L184-L229)
* [Grid3D](https://github.com/ignitionrobotics/ign-gazebo/blob/f795b657683f629133e7de056771959786a150fd/src/gui/plugins/grid_config/GridConfig.cc#L152-L187)
* [VisualizeLidar](https://github.com/ignitionrobotics/ign-gazebo/blob/a41cf80049c07977079d0b215398dc4225b63038/src/gui/plugins/visualize_lidar/VisualizeLidar.cc#L151-L186)
* etc

This PR could target `ign-rendering3`, I'm targeting `main` because I have other PRs that will depend on it. Once this has been reviewed I can rebase and target 3.

## Test it

Take a look at the added tests.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/rendering/RenderingIface.hh
+++ b/include/ignition/rendering/RenderingIface.hh
@@ -24,6 +24,7 @@
 
 #include "ignition/rendering/config.hh"
 #include "ignition/rendering/Export.hh"
+#include "ignition/rendering/RenderTypes.hh"
 
 namespace ignition
 {
@@ -131,6 +132,17 @@ namespace ignition
     /// \param[in] _paths The list of the plugin paths
     IGNITION_RENDERING_VISIBLE
     void setPluginPaths(const std::list<std::string> &_paths);
+
+    /// \brief Most applications will only have one rendering engine loaded
+    /// at a time, and only one scene within that. This helper function gets
+    /// the first scene that can be found in the first loaded rendering engine.
+    ///
+    /// It's not recommended to call this function when there's more than one
+    /// engine or scene.
+    ///
+    /// \return Pointer to a scene that was found, null if no scene is loaded.
+    IGNITION_RENDERING_VISIBLE
+    ScenePtr sceneFromFirstRenderEngine();
     }
   }
 }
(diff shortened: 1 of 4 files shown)


ahcorde (changes requested):
LGTM, just one minor comment, what do you think about use another name for the method? `getFirstInitializedScene` ?

I can see an error in windows 

```bash
47: [31m[Err] [C:\Jenkins\workspace\ign_rendering-pr-win\ws\ign-rendering\src\RenderEngineManager.cc:336] [m[31mInvalid render-engine index: [m[31m1000000[m[31m
47: [m[       OK ] RenderingIfaceTest.RegisterEngine (67 ms)
```

chapulina:
> what do you think about use another name for the method? getFirstInitializedScene ?

I  don't know if that name would be necessarily true for the current implementation. For example:

1. Register engine EA
2. Register engine EB
3. Create scene on SB on EB
4. Create scene on SA on EA

SB was initialized before SA, but we'll return SA. Moreover, if we don't do step 4, we will return null.

ahcorde:
@osrf-jenkins retest this please

ahcorde (approved):

iche033:
> I don't know if that name would be necessarily true for the current implementation.

How about `getFirstRenderEngineScene` to be more specific?

chapulina:
> How about getFirstRenderEngineScene to be more specific

The trick is whether the "first" applies to the render engine, to the scene, or to both...

I don't have a better suggestion though. I can go with that if you think there isn't a lot of ambiguity. I'd just drop the `get`.

iche033:
`sceneFromFirstRenderEngine`? 

I think either one works

chapulina:
> sceneFromFirstRenderEngine?

Ok, I went with that: e19bd7353d94552db111704f621c4efe627fda95

chapulina:
> Once this has been reviewed I can rebase and target 3.

Done, this is now targeted at v3

iche033 (approved):
looks good to me.

PR #325, opened 2021-05-21
Changed calculation for range clipping

Lobotuerk:
Signed-off-by: Tomas Lorente <jtlorente@ekumenlabs.com>

# 🦟 Bug fix

Fixes https://github.com/osrf/subt/issues/342

## Summary
This changes clipping to be based on distance to the focal point instead of distance to the focal plane.

## Checklist
- [ ] Added tests


ogre2/src/media/materials/programs/depth_camera_fs.glsl
```diff
@@ -139,7 +139,8 @@ void main()
   }
 
   // clamp xyz and set rgb to background color
-  if (point.x > far - tolerance)
+  if (point.x * point.x + point.y * point.y >
+      (far - tolerance) * (far - tolerance))
```
iche033, line 143: how about just `length(point) > far - tolerance` to be consistent with how the gpu lidar clips ranges? 

Similarly for clipping points within near clip plane: `length(point) < near + tolerance`

Lobotuerk: Makes sense, I had no idea there was a length function. Good catch

Lobotuerk: Does 9bfe9d1 works?

iche033: yeah that looks fine. `INTEGRATION_depth_camera` test caught this change and is failing. Can you update the test?

Can you also add a note about this change in the [Migration.md](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering5/Migration.md) guide?

Lobotuerk:
Given that the issue is about longer ranges, and that closer clipping is giving some issues that go farther than this PR, this is going to tackle only far clipping and not near clipping. The near clipping seems to be related to https://github.com/ignitionrobotics/ign-rendering/pull/356#issuecomment-870429869 , would need that fixed before near clipping can be done like far clipping

ogre2/src/media/materials/programs/depth_camera_fs.glsl
Lobotuerk: Where should I note the change in the migration guide? 5 -> 6 ?

iche033: yeah, you can add a new section `Ignition Rendering 5.0 to 6.0`

ogre2/src/media/materials/programs/depth_camera_fs.glsl
```diff
@@ -139,7 +139,7 @@ void main()
   }
 
   // clamp xyz and set rgb to background color
-  if (point.x > far - tolerance)
```
iche033, line 142: can you make the same change to `depth_camera_final_fs.glsl`?

Lobotuerk: If I change this, the tests start falling again. expecting -inf and getting inf https://github.com/ignitionrobotics/ign-rendering/blob/lobotuerk/depthRangeChange/test/integration/depth_camera.cc#L324-L326

Lobotuerk: Changing both that and https://github.com/ignitionrobotics/ign-rendering/blob/lobotuerk/depthRangeChange/test/integration/depth_camera.cc#L309-L311 to`maxVal` make all tests pass, but I don't think that makes sense

iche033: I think the problem is that because `point` was previously clamped to `-inf` so `length(point)` becomes `+inf`. So in `depth_camera_final_fs.glsl` we can change to something like:

```
if (!isinf(point.x) && length(point) > far - tolerance)
```

which just skips this additional clamping if it's already been clamped in the previous pass.

Lobotuerk: 34267e4 should take care of this

ogre2/src/media/materials/programs/depth_camera_fs.glsl
Lobotuerk: 34267e4 should take care of this

iche033 (approved):
looks good to me.

Issue #134, opened 2020-08-31
COLLADA Rendering Failure: instance_geometry Conflict


jaelrod:
I have the need to programmatically generate some COLLADA files to simulate some dynamic assets in Gazebo. Loosely following COLLADA specification, I am declaring a simple triangles mesh in library_geometries and trying to instantiate that same geometry multiple times with different transformation matrices and materials applied to it. Attached is a file that demonstrates the below screenshots.
[mesh.dae.txt](https://github.com/ignitionrobotics/ign-rendering/files/5152607/mesh.dae.txt)

This is the mesh loaded into Blender, looking as expected:
![image](https://user-images.githubusercontent.com/64549955/91767245-b2743680-eb90-11ea-9484-3cd8d2356a9f.png)

However, the same exact COLLADA file loaded into Ignition Gazebo (v3.2.0) looks like the following. The rendered mesh has only one plane instead of two, and that plane takes the shape of the instance_geometry that is declared first and the bind_material of the instance_geometry declared second:
![image](https://user-images.githubusercontent.com/64549955/91767796-9755f680-eb91-11ea-9196-761227462978.png)

I seem to be able to work around this by defining duplicate geometries inside of library_geometries for each instance_geometry, but that sort of defeats the point of being able to reuse recurring shapes from the geometry library.

chapulina:
Thank you for ticketing the issue. The fix may need to go into [ignition::common::ColladaLoader](https://github.com/ignitionrobotics/ign-common/blob/master/graphics/src/ColladaLoader.cc), which is in a different library.

jaelrod:
Sounds good. My apologies for placing the issue in the wrong repo. After ticketing this yesterday, I encountered a separate but similar issue, so I'll put that one there. Is COLLADA the preferred asset representation format for Ignition Gazebo? Is there better support for a different format?

chapulina:
> My apologies for placing the issue in the wrong repo.

No worries. There's a possibility that the issue is coming from here, I just mentioned the collada loader for completeness.

> Is COLLADA the preferred asset representation format for Ignition Gazebo?

There's also support for OBJ, you may have better luck with that, depending on the asset. For animations, we also support BVH.

PR #329, opened 2021-05-28
Segmentation Camera

AmrElsersy:
### 🎉 New feature

related to [#134](https://github.com/ignitionrobotics/ign-sensors/issues/134)

## Summary
Segmentation Camera to capture semantic information, Implemented using OGRE 2 by using the material switching to provide Semantic & Panoptic Segmentation

## Semantic Segmentation

- Pixels of same label from different items, have the same color & id

## Panoptic Segmentation

- Pixels of same label from different items, have different color & id
- 1 channel for label id & 2 channels for instance id

## Features:
- Supports both Semantic Ids map(where each pixel has a value = label id) & Colored map(where each pixel has a color) for visualization purposes 
- The user label each item by setting its UserData
- Each unlabeled item is considered as a background
- Added example in ign-rendering/examples to try it easily


include/ignition/rendering/Scene.hh
```diff
@@ -657,6 +657,35 @@ namespace ignition
       public: virtual ThermalCameraPtr CreateThermalCamera(
                   unsigned int _id, const std::string &_name) = 0;
 
+      /// \brief Create new depth camera. A unique ID and name will
```
mjcarroll, line 660:
```suggestion
      /// \brief Create new segmentation camera. A unique ID and name will
```

include/ignition/rendering/Scene.hh
```diff
@@ -657,6 +657,35 @@ namespace ignition
...
+      /// automatically be assigned to the camera.
+      /// \return The created camera
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera() = 0;
+
+      /// \brief Create new depth camera with the given ID. A unique name
```
mjcarroll, line 665:
```suggestion
      /// \brief Create new segmentation camera with the given ID. A unique name
```

include/ignition/rendering/Scene.hh
```diff
@@ -657,6 +657,35 @@ namespace ignition
...
+      /// will automatically be assigned to the camera. If the given ID is
+      /// already in use, NULL will be returned.
+      /// \param[in] _id ID of the new camera
+      /// \return The created camera
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera(
+                  unsigned int _id) = 0;
+
+      /// \brief Create new depth camera with the given name. A unique ID
```
mjcarroll, line 673:
```suggestion
      /// \brief Create new segmentation camera with the given name. A unique ID
```

include/ignition/rendering/Scene.hh
```diff
@@ -657,6 +657,35 @@ namespace ignition
...
+      /// will automatically be assigned to the camera. If the given name is
+      /// already in use, NULL will be returned.
+      /// \param[in] _name Name of the new camera
+      /// \return The created camera
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera(
+                  const std::string &_name) = 0;
+
+      /// \brief Create new depth camera with the given name. If either the
```
mjcarroll, line 681:
```suggestion
      /// \brief Create new segmentation camera with the given name. If either the
```

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
+/*
+ * Copyright (C) 2017 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/GlutWindow.hh
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/GlutWindow.hh
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#ifndef IGNITION_RENDERING_EXAMPLES_THERMAL_CAMERA_GLUTWINDOW_HH_
+#define IGNITION_RENDERING_EXAMPLES_THERMAL_CAMERA_GLUTWINDOW_HH_
```
mjcarroll, line 18:
```suggestion
#ifndef IGNITION_RENDERING_EXAMPLES_SEGMENTATION_CAMERA_GLUTWINDOW_HH_
#define IGNITION_RENDERING_EXAMPLES_SEGMENTATION_CAMERA_GLUTWINDOW_HH_
```

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,214 @@
+/*
+ * Copyright (C) 2019 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,90 @@
...
+#ifndef IGNITION_RENDERING_SEGMENTATIONCAMERA_HH_
+#define IGNITION_RENDERING_SEGMENTATIONCAMERA_HH_
+
+#include <string>
+
+#include <ignition/common/Event.hh>
+#include "ignition/rendering/Camera.hh"
+#include "ignition/math/Color.hh"
+
+// Segmentation types for semantic/instance segmentation
```
mjcarroll, line 26: I think it would be helpful to document the behavior of each choice here.

include/ignition/rendering/base/BaseSegmentationCamera.hh
```diff
@@ -0,0 +1,140 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+#ifndef IGNITION_RENDERING_BASE_BASESegmentationCAMERA_HH_
+#define IGNITION_RENDERING_BASE_BASESegmentationCAMERA_HH_
```
mjcarroll, line 18:
```suggestion
#ifndef IGNITION_RENDERING_BASE_BASESEGMENTATIONCAMERA_HH_
#define IGNITION_RENDERING_BASE_BASESEGMENTATIONCAMERA_HH_
```

include/ignition/rendering/base/BaseSegmentationCamera.hh
```diff
@@ -0,0 +1,140 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ogre2/include/ignition/rendering/ogre2/Ogre2SegmentationCamera.hh
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ogre2/include/ignition/rendering/ogre2/Ogre2SegmentationCamera.hh
```diff
@@ -0,0 +1,142 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+
+#ifndef IGNITION_RENDERING_OGRE2_OGRE2SegmentationCAMERA_HH_
+#define IGNITION_RENDERING_OGRE2_OGRE2SegmentationCAMERA_HH_
```
mjcarroll, line 19:
```suggestion
#ifndef IGNITION_RENDERING_OGRE2_OGRE2SEGMENTATIONCAMERA_HH_
#define IGNITION_RENDERING_OGRE2_OGRE2SEGMENTATIONCAMERA_HH_
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,648 @@
+/*
+ * Copyright (C) 2018 Open Source Robotics Foundation
```
mjcarroll, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,648 @@
...
+
+/////////////////////////////////////////////////
+math::Color SegmentationMaterialSwitcher::LabelToColor(int64_t _label)
+{
+  if (_label == this->backgroundLabel)
+    return this->backgroundColor;
+
+  // use label as seed to generate the same color for the label
+  this->generator.seed(_label);
+  std::uniform_int_distribution<int> distribution(0, 255);
```
mjcarroll, line 188: As a future enhancement, it may be helpful to have a configurable colormap so that we can have more deterministic ordering.

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,648 @@
...
+  public: std::string workspaceDefinition;
+
+  /// \brief Render Texture to store the final segmentation data
+  public: Ogre::RenderTexture *ogreRenderTexture;
+
+  /// \brief Texture to create the render texture from.
+  public: Ogre::TexturePtr ogreTexture;
+
+  /// \brief Pixel Box to copy render texture data to a buffer
+  public: Ogre::PixelBox *pixelBox;
```
mjcarroll, line 380: Initialize all of these pointers to `nullptr`.  We've had bugs in the past when they weren't initialized.

mjcarroll (dismissed):
Overall looking really good.  A few notes:

* Copyright 2021 on newly-created files
* ifdefs get all caps
* Initialize pointers to `nullptr`

examples/segmentation_camera/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+cmake_minimum_required(VERSION 3.10.2 FATAL_ERROR)
```
ahcorde, line 1: Do you main to add support for windows in the CMakelists and example ? [here a reference](https://github.com/ignitionrobotics/ign-rendering/pull/291)

examples/segmentation_camera/GlutWindow.cc
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+  bool motionDirty = false;
+};
+struct mouseButton g_mouse;
+std::mutex g_mouseMutex;
+
+//////////////////////////////////////////////////
+void mouseCB(int _button, int _state, int _x, int _y)
+{
+  // ignore unknown mouse button numbers
+  if (_button >= 5)
```
ahcorde, line 97: magic number? add a constant

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+    if (!g_rayQuery)
+    {
+      ignerr << "Failed to create Ray Query" << std::endl;
+      return;
+    }
+  }
+
+  if (g_mouse.buttonDirty)
+  {
+    // std::cout << "(" << g_mouse.x << "," << g_mouse.y << ")" << std::endl;
```
ahcorde, line 151: remove

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+        2.0 * g_mouse.y / static_cast<double>(rayCamera->ImageHeight());
+
+    // g_rayQuery->SetFromCamera(rayCamera, ignition::math::Vector2d(nx, ny));
+    // g_target  = g_rayQuery->ClosestPoint();
+    // if (!g_target)
+    // {
+    //   // set point to be 10m away if no intersection found
+    //   g_target.point = g_rayQuery->Origin() + g_rayQuery->Direction() * 10;
+    //   return;
+    // }
```
ahcorde, line 165: remove

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+      double amount = -(scroll * factor) * (distance / 5.0);
+      for (ir::CameraPtr camera : g_cameras)
+      {
+        g_viewControl.SetCamera(camera);
+        g_viewControl.SetTarget(g_target.point);
+        g_viewControl.Zoom(amount);
+      }
+    }
+  }
+
```
ahcorde, line 184:
```suggestion

```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+  if (g_mouse.motionDirty)
+  {
+    g_mouse.motionDirty = false;
+    auto drag = ignition::math::Vector2d(g_mouse.dragX, g_mouse.dragY);
+
+    // left mouse button pan
+    if (g_mouse.button == GLUT_LEFT_BUTTON && g_mouse.state == GLUT_DOWN)
+    {
+      for (ir::CameraPtr camera : g_cameras)
```
ahcorde, line 194:
```suggestion
      for (auto &camera : g_cameras)
```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+      {
+        g_viewControl.SetCamera(camera);
+        g_viewControl.SetTarget(g_target.point);
+        g_viewControl.Pan(drag);
+      }
+    }
+    else if (g_mouse.button == GLUT_MIDDLE_BUTTON && g_mouse.state == GLUT_DOWN)
+    {
+      for (ir::CameraPtr camera : g_cameras)
```
ahcorde, line 203:
```suggestion
      for (auto &camera : g_cameras)
```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+          * distance * tan(vfov/2.0) * 6.0);
+      for (ir::CameraPtr camera : g_cameras)
+      {
+        g_viewControl.SetCamera(camera);
+        g_viewControl.SetTarget(g_target.point);
+        g_viewControl.Zoom(amount);
+      }
+    }
+  }
+
```
ahcorde, line 229:
```suggestion
```

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,376 @@
...
+}
+
```
ahcorde, line 231:
```suggestion
```

examples/segmentation_camera/GlutWindow.hh
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/Main.cc
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,214 @@
...
+  camera->SetLocalPosition(0.0, 0.0, 0.5);
+  camera->SetLocalRotation(0.0, 0.0, 0.0);
+  camera->SetImageWidth(800);
+  camera->SetImageHeight(600);
+  camera->SetImageFormat(PixelFormat::PF_R8G8B8);
+  camera->SetFarClipPlane(10);
+  camera->SetNearClipPlane(1.0);
+  camera->SetAspectRatio(1.333);
+  camera->SetHFOV(IGN_PI / 2);
+  // camera->SetBackgroundColor(math::Color(255,255,255));
```
ahcorde, line 159:
```suggestion
```

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,214 @@
...
+  camera->EnableColoredMap(true);
+  // camera->SetSegmentationType(SegmentationType::Panoptic);
```
ahcorde, line 161:
```suggestion
```

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,214 @@
...
+    {
+      CameraPtr camera = createCamera(engineName);
+      if (camera)
+      {
+        cameras.push_back(camera);
+      }
+    }
+    catch (...)
+    {
+      // std::cout << ex.what() << std::endl;
```
ahcorde, line 208:
```suggestion
```

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,214 @@
...
+//////////////////////////////////////////////////
+int main(int _argc, char** _argv)
+{
+  glutInit(&_argc, _argv);
+
+  common::Console::SetVerbosity(4);
+  std::vector<std::string> engineNames;
+  std::vector<CameraPtr> cameras;
+
+  engineNames.push_back("ogre2");
```
ahcorde, line 194: can you include this logic?

https://github.com/WilliamLewww/ign-rendering/blob/cc618741a43d86eaab6d0419a205f5d862e336f2/examples/actor_animation/Main.cc#L180-L188

include/ignition/rendering/SegmentationCamera.hh
ahcorde, line 2:
```suggestion
 * Copyright (C) 20121 Open Source Robotics Foundation
```

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,90 @@
...
+        ConnectNewSegmentationFrame(
+          std::function<void(const uint8_t *, unsigned int, unsigned int,
+          unsigned int, const std::string &)>  _subscriber) = 0;
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
+      public: virtual void SetSegmentationType(SegmentationType _type) = 0;
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
```
ahcorde, line 68: Review

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,90 @@
...
+      public: virtual void EnableColoredMap(bool _enable) = 0;
+
+      /// \brief Set color for background & unlabeled items in the colored map
+      /// \param[in] _color Color of background & unlabeled items
+      public: virtual void SetBackgroundColor(math::Color _color) = 0;
+
+      /// \brief Set label for background & unlabeled items in the semantic map
+      /// \param[in] _color label of background & unlabeled items
```
ahcorde, line 76: review

include/ignition/rendering/base/BaseScene.hh
```diff
@@ -438,6 +453,18 @@ namespace ignition
                    return ThermalCameraPtr();
                  }
 
+      /// \brief Implementation for creating a segmentation camera.
+      /// \param[in] _id Unique id
+      /// \param[in] _name Name of segmentation camera
```
ahcorde, line 458: add \return

include/ignition/rendering/base/BaseSegmentationCamera.hh
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

include/ignition/rendering/base/BaseSegmentationCamera.hh
```diff
@@ -0,0 +1,140 @@
...
+  {
+    inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
+
+    template <class T>
+    class BaseSegmentationCamera :
+      public virtual SegmentationCamera,
+      public virtual BaseCamera<T>,
+      public virtual T
+    {
+      protected: BaseSegmentationCamera();
```
ahcorde, line 39: add docs or at least `// Documentation inherited`

ogre2/include/ignition/rendering/ogre2/Ogre2SegmentationCamera.hh
ahcorde, line 2:
```suggestion
 * Copyright (C) 2021 Open Source Robotics Foundation
```

ogre2/include/ignition/rendering/ogre2/Ogre2SegmentationCamera.hh
```diff
@@ -0,0 +1,142 @@
...
+      /// \brief Create dummy render texture. Needed to satisfy inheritance
+      /// and to set image's dims
+      public: virtual void CreateRenderTexture();
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
+      public: void SetSegmentationType(SegmentationType _type);
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
```
ahcorde, line 107: review

ogre2/include/ignition/rendering/ogre2/Ogre2SegmentationCamera.hh
```diff
@@ -0,0 +1,142 @@
...
+      public: void EnableColoredMap(bool _enable);
+
+      /// \brief Set color for background & unlabeled items in the colored map
+      /// \param[in] _color Color of background & unlabeled items
+      public: void SetBackgroundColor(math::Color _color);
+
+      /// \brief Set label for background & unlabeled items in the semantic map
+      /// \param[in] _color label of background & unlabeled items
```
ahcorde, line 115: review

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,648 @@
...
+      private: std::map<int64_t, bool> takenColors;
+
+      /// \brief keep track of the labels which is already colored
+      /// usful for coloring items in semantic mode in LabelToColor()
+      /// key: label id. value: always True
+      private: std::map<int64_t, bool> coloredLabel;
+
+      private: std::default_random_engine generator;
+
+      // private: std::uniform_int_distribution<int> distribution(1,2);
```
ahcorde, line 117: remove

ahcorde (changes requested):

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: @mjcarroll do you mean storing the color for each label id of all 255 id ?

examples/segmentation_camera/GlutWindow.cc
AmrElsersy: This code is needed, it is not working now(it crashes on setting the query from the camera), but i am working on fixing it :D

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,218 @@
...
+  if (_argc > 1)
+  {
+    ogreEngineName = _argv[1];
+  }
+
+  common::Console::SetVerbosity(4);
+  std::vector<std::string> engineNames;
+  std::vector<CameraPtr> cameras;
+
+  engineNames.push_back("ogre2");
```
ahcorde, line 199:
```suggestion
  engineNames.push_back(ogreEngineName);
```

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,94 @@
...
+        ConnectNewSegmentationFrame(
+          std::function<void(const uint8_t *, unsigned int, unsigned int,
+          unsigned int, const std::string &)>  _subscriber) = 0;
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
+      public: virtual void SetSegmentationType(SegmentationType _type) = 0;
+
+      /// \brief Set Segmentation Type (Semantic / Instance)
+      /// \param[in] _type Segmentation Type (Semantic / Instance)
```
ahcorde, line 72:
```suggestion
      /// \param[in] _enable ..
```

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,94 @@
...
+      public: virtual void EnableColoredMap(bool _enable) = 0;
+
+      /// \brief Set color for background & unlabeled items in the colored map
+      /// \param[in] _color Color of background & unlabeled items
+      public: virtual void SetBackgroundColor(math::Color _color) = 0;
+
+      /// \brief Set label for background & unlabeled items in the semantic map
+      /// \param[in] _color label of background & unlabeled items
```
ahcorde, line 80:
```suggestion
      /// \param[in] _label label of background & unlabeled items
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+}
+
+/////////////////////////////////////////////////
+bool SegmentationMaterialSwitcher::IsTakenColor(math::Color _color)
+{
+  // get the int value of the 24 bit color
+  int64_t colorId = _color.R() * 255 * 255 + _color.G() * 255 + _color.B();
+
+  if (this->takenColors.count(colorId))
+    return true;
```
ahcorde, line 171:
```suggestion
  if (this->takenColors.count(colorId))
  {
      return true;
  }
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+    this->Name();
+  auto backgroundColor = Ogre2Conversions::Convert(
+    this->dataPtr->materialSwitcher->backgroundColor);
+
+  // basic workspace consist of clear pass with the givin color &
+  // a render scene pass to the givin render texture
+  this->dataPtr->ogreCompositorManager->createBasicWorkspaceDef(
+    this->dataPtr->workspaceDefinition,
+    backgroundColor
+    );
```
ahcorde, line 518:
```suggestion
  this->dataPtr->ogreCompositorManager->createBasicWorkspaceDef(
    this->dataPtr->workspaceDefinition, backgroundColor);
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+    this->dataPtr->buffer,
+    width, height, channelCount,
+    Ogre::PixelUtil::getFormatName(this->dataPtr->format)
+  );
+}
+
+/////////////////////////////////////////////////
+uint8_t *Ogre2SegmentationCamera::SegmentationData() const
+{
+    return this->dataPtr->buffer;
```
ahcorde, line 583:
```suggestion
  return this->dataPtr->buffer;
```

ahcorde (changes requested):

examples/segmentation_camera/CMakeLists.txt
AmrElsersy: No, I don't know how to add it :D, i just copied that from the examples

ahcorde (dismissed):
check linters https://github.com/ignitionrobotics/ign-rendering/pull/329/checks?check_run_id=2712808360

include/ignition/rendering/base/BaseScene.hh
```diff
@@ -296,6 +296,21 @@ namespace ignition
...
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera(
+        const unsigned int _id) override;
+
+      // Documentation inherited.
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera(
+        const std::string &_name) override;
+
+      // Documentation inherited.
+      public: virtual SegmentationCameraPtr CreateSegmentationCamera(
+        const unsigned int _id, const std::string &_name) override;
```
adlarkin, line 312: It looks like this breaks ABI: https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/598/API_5fABI_20report/compat_report.html#Type_Binary_Problems_High

According to the [ABI guide from KDE](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B):

> You cannot...
> 
> add new virtual functions to non-leaf classes as this will break subclasses. Note that a class designed to be sub-classed by applications is always a non-leaf class.
>
> Reason: the addition of a new virtual function to a class that is non-leaf (that is, there is at least one class deriving from this class) changes the layout of the virtual table (the virtual table is basically a list of function pointers, pointing to the functions that are active in this class level). To accommodate the new virtual, the compiler must add a new entry to this table, but existing derived classes won't know about it and will not have the entry in their virtual tables.

So, unless there's a way for us to achieve this functionality without adding any virtual methods to `BaseScene`, we may need to re-target this for Fortress :grimacing:

adlarkin: > the addition of a new virtual function to a class that is non-leaf (that is, there is at least one class deriving from this class) changes the layout of the virtual table

I actually believe that the ABI breaking code are the pure virtual methods added to the `Scene` class, since the `Scene` class is a non-leaf class (I accidentally commented on the `BaseScene` code, which could be considered a leaf class):

https://github.com/ignitionrobotics/ign-rendering/blob/3b8c51ea827996f56331ef7ff08b45f67ccc78ed/include/ignition/rendering/Scene.hh#L660-L687

iche033: ok yeah these will break ABI and It would be difficult to work around it. I recommend targeting at `main`

adlarkin: I went ahead and updated this PR to target `main` and updated the segmentation example `CMakeLists.txt` to depend on `ign-rendering6` in https://github.com/ignitionrobotics/ign-rendering/pull/329/commits/8813531a2df28210d292445afc4f5081806b3f8d: https://github.com/ignitionrobotics/ign-rendering/blob/8813531a2df28210d292445afc4f5081806b3f8d/examples/segmentation_camera/CMakeLists.txt#L3

I believe everything should be up-to-date for Fortress now, but it looks like DCO is giving a lot of complaints: https://github.com/ignitionrobotics/ign-rendering/pull/329/checks?check_run_id=2765358606

I know we can set DCO to manually pass, but there seem to be a lot more failures in the check I linked above than usual. @iche033, any thoughts?

iche033: not sure how DCO got into that state but it's complaining about commits outside of this PR. You can probably just set DCO to pass on this one.

adlarkin: > not sure how DCO got into that state

It looks like it had to do with the fact that I reset the branch of this PR to `main` and committed this change before updating the PR itself to target main. I pushed another commit to address some warnings, and now, DCO seems fine. So, I think we can mark this as resolved :+1:

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+/////////////////////////////////////////////////
+SegmentationMaterialSwitcher::~SegmentationMaterialSwitcher()
+{
+}
+
+/////////////////////////////////////////////////
+bool SegmentationMaterialSwitcher::IsTakenColor(math::Color _color)
+{
+  // get the int value of the 24 bit color
+  int64_t colorId = _color.R() * 255 * 255 + _color.G() * 255 + _color.B();
```
adlarkin, line 168: Is this supposed to be a unique ID? Is there any chance that 2 different RGB combinations could yield the same ID with this formula?

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+  int r = distribution(this->generator);
+  int g = distribution(this->generator);
+  int b = distribution(this->generator);
+
+  math::Color color(r, g, b);
+
+  if (this->type == SegmentationType::Semantic)
+  {
+    // if the label is colored before return the color
+    // don't check fo taken colors in that case, all items
```
adlarkin, line 201:
```suggestion
    // don't check for taken colors in that case, all items
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+    // with the same label will have the same color
+    if (this->coloredLabel.count(_label))
+      return color;
+  }
```
adlarkin, line 205: This can be consolidated into a single `if` statement:
```c++
if (this->type == SegmentationType::Semantic && this->coloredLabel.count(_label))
  return color;
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+      {
+        ignerr << "Ogre Error:" << e.getFullDescription() << "\n";
+      }
+      Ogre2VisualPtr ogreVisual = std::dynamic_pointer_cast<Ogre2Visual>(
+        visual);
+
+      // get class user data
+      Variant labelAny = ogreVisual->UserData(this->labelKey);
+
+      int label;
```
adlarkin, line 248:
```suggestion
      // get class user data
      Variant labelAny = ogreVisual->UserData(this->labelKey);
      int label;
```
nit: whitespace

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+            math::Color color = this->LabelToColor(label);
+            subItem->setCustomParameter(1, Ogre::Vector4(
+              color.R(), color.G(), color.B(), 1.0));
+          }
+          else
+          {
+            // labels ids material(each pixel has item's label)
+            float labelColor = label / 255.0;
+            subItem->setCustomParameter(1, Ogre::Vector4(
+              labelColor, labelColor, labelColor, 1.0));
```
adlarkin, line 281: I'm concerned that this won't be very useful for users. Users want labels to be unique, and want labels to be something that is easy to work with (like an int). However, here, it looks like we are normalizing labels as floats between 0 and 1, right? If so, it would be difficult for users to compare pixels to see if they have the same label or not due to floating point roundoff error.

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+          subItem->setMaterial(this->plainMaterial);
+      }
+    }
+    itor.moveNext();
+  }
+
+  // reset the count & colors tracking
+  this->instancesCount.clear();
+  this->takenColors.clear();
+  this->coloredLabel.clear();
```
adlarkin, line 333: Is there a reason why we have to reset this after every pre-render? This makes me think that we'd have to re-compute colors each time around, and since we are using a random number generator to compute colors, this could mean that we don't have consistent colors for objects on each update.

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+    Ogre::Item *item = static_cast<Ogre::Item *>(object);
+    for (unsigned int i = 0; i < item->getNumSubItems(); i++)
+    {
+      Ogre::SubItem *subItem = item->getSubItem(i);
+      auto it = this->datablockMap.find(subItem);
+      if (it != this->datablockMap.end())
+        subItem->setDatablock(it->second);
+    }
+    itor.moveNext();
+  }
```
adlarkin, line 355: I think this could be simplified to look something like this:

https://github.com/ignitionrobotics/ign-rendering/blob/df77bf5ef752e5c0489973d9f856184df5a1d081/ogre2/src/Ogre2ThermalCamera.cc#L427-L432

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+}
+
+/////////////////////////////////////////////////
+class ignition::rendering::Ogre2SegmentationCameraPrivate
+{
```
adlarkin, line 360: nit: the `Ogre2SegmentationCameraPrivate` class definition should come right after the `SegmentationMaterialSwitcher` class definition. All of the method definitions for both the `SegmentationMaterialSwitcher` and `Ogre2SegmentationCameraPrivate` class should then come after each of these class definitions.

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+  /// \brief Material Switcher to switch item's material
+  /// with colored version for segmentation
+  public: SegmentationMaterialSwitcher *materialSwitcher {nullptr};
+
+  /// \brief Compositor Manager to create workspace
+  public: Ogre::CompositorManager2 *ogreCompositorManager {nullptr};
+
+  /// \brief Workspace to interface with render texture
+  public: Ogre::CompositorWorkspace *ogreCompositorWorkspace {nullptr};
```
adlarkin, line 369: Can we use `std::unique_ptr` for these types instead of raw pointers? Same comment applies to all of the other raw pointers being used in the `Ogre2SegmentationCameraPrivate` class

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+  this->ogreCamera->setFixedYawAxis(false);
+
+  this->ogreCamera->setAutoAspectRatio(true);
+  this->ogreCamera->setRenderingDistance(100);
+  this->ogreCamera->setProjectionType(Ogre::ProjectionType::PT_PERSPECTIVE);
+  this->ogreCamera->setCustomProjectionMatrix(false);
+}
+
+/////////////////////////////////////////////////
+void Ogre2SegmentationCamera::Destroy()
```
adlarkin, line 453: Make sure you are deleting all of the raw pointers in the `Ogre2SegmentationPrivate` class (`dataPtr` member variable) here. Or, if you use `std::unique_ptr` instead of raw pointers as I suggested in my other comment, you don't need to worry about this :slightly_smiling_face:

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+  /// \brief dummy render texture to set image dims
+  public: Ogre2RenderTexturePtr dummyTexture {nullptr};
+
+  /// \brief New Segmentation Frame Event to notify listeners with new data
+  public: ignition::common::EventT<void(const uint8_t *, unsigned int,
+        unsigned int, unsigned int, const std::string &)>
+        newSegmentationFrame;
+
+  /// \brief Image / Render Texture Format
+  public: Ogre::PixelFormat format = Ogre::PF_R8G8B8;
```
adlarkin, line 395: Can this variable be marked `const`?

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+  /// \brief buffer to store render texture data & to be sent to listeners
+  public: uint8_t *buffer = nullptr;
+
+  /// \brief dummy render texture to set image dims
+  public: Ogre2RenderTexturePtr dummyTexture {nullptr};
+
+  /// \brief New Segmentation Frame Event to notify listeners with new data
+  public: ignition::common::EventT<void(const uint8_t *, unsigned int,
+        unsigned int, unsigned int, const std::string &)>
+        newSegmentationFrame;
```
adlarkin, line 392: Can you document the template parameter of this variable? It looks like it's a callback method with a few arguments, so don't forget to document the arguments to the callback as well :slightly_smiling_face:

adlarkin (changes requested):
Can you add some tests for the segmentation camera as well? It would be great to have unit tests (in the `src/` directory) and an integration test (in the `test/integration/` directory).

examples/segmentation_camera/Main.cc
```diff
@@ -0,0 +1,216 @@
...
+  root->AddChild(sphere);
+
+  // create a sphere2
+  VisualPtr sphere2 = _scene->CreateVisual("sphere2");
+  sphere2->SetLocalPosition(5, 4, 2);
+  GeometryPtr sphereGeom2 = _scene->CreateSphere();
+  sphere2->AddGeometry(sphereGeom2);
+  sphere2->SetMaterial(red);
+  sphere2->SetUserData("label", 3);
+  root->AddChild(sphere2);
```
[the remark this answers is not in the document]
AmrElsersy: actually materials are useless here, i will remove it
the one that is shown the material of the material switcher, and these materials you mention will not appear anyway

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,374 @@
...
+  glutMouseFunc(mouseCB);
+  glutMotionFunc(motionCB);
+}
+
+//////////////////////////////////////////////////
+void printUsage()
+{
+  std::cout << "===============================" << std::endl;
+  std::cout << "  TAB - Switch render engines  " << std::endl;
+  std::cout << "  ESC - Exit                   " << std::endl;
```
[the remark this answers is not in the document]
AmrElsersy: ok i will remove the TAB, but ESC is working for me

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+      private: std::map<Ogre::SubItem *, Ogre::HlmsDatablock *> datablockMap;
+
+      /// \brief Ogre v1 material consisting of a shader that changes the
+      /// appearance of item to use a unique color for mouse picking
+      private: Ogre::MaterialPtr plainMaterial;
+
+      /// \brief Ogre v1 material consisting of a shader that changes the
+      /// appearance of item to use a unique color for mouse picking. In
+      /// addition, the depth check and depth write properties disabled.
+      private: Ogre::MaterialPtr plainOverlayMaterial;
```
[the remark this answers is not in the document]
AmrElsersy: Actually I copied that documentation from the implementation of all other material switchers, but don't know what is actually used for

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+      /// \brief Segmentation Type (Semantic / Instance)
+      private: SegmentationType type {SegmentationType::Semantic};
+
+      /// \brief True to generate colored map
+      /// False to generate labels ids map
+      private: bool isColoredMap;
+
+      /// \brief Keep track of num of instances of the same label
+      /// Key: label id, value: num of instances
+      private: std::map<int, int> instancesCount;
```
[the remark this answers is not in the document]
AmrElsersy: No, always >= 0

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: Yes, we have 3 channels each is 8 bit, that ID gets the decimal value of that 24 bit numbers, and there is no change to have the same ID twice

btw, this is the same way the panoptic is  handled in the official panoptic API proposed by COCO dataset (which introduced the panoptic format)

check that if you want to make sure
https://github.com/cocodataset/panopticapi/blob/master/panopticapi/utils.py#L30

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: It is in the range of [0-255], that normalization is just required to pass it to the function, but OGRE map it again to [0-255] in the buffer that the segmentation data is stored in

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: Yeah, It is calculated each iteration. but the same colors will be produced for the objects !
because in the random generator, I use the unique encoded ID of each item as a seed. so it generates the same color each time

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: mmm, I tried to make that but didn't succeed and got many errors because of OGRE, it expects in all functions to deal with raw pointers

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,647 @@
...
+
+      /// \brief Convert label of semantic map to a unique color for colored map
+      /// \param[in] _label id of the semantic map or encoded id of panoptic map
+      /// \return _color unique color in the colored map for that label
+      private: math::Color LabelToColor(int64_t _label);
+
+      /// \brief Check if the color is taken previously & store it if not taken
+      /// \param[in] _color Color to be checked
+      /// \return True if taken, False elsewhere
+      private: bool IsTakenColor(math::Color _color);
```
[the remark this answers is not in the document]
AmrElsersy: mmm, I think it will is also difficult to understand in "StoreColor" name :D 
but ok will take a look at it

examples/segmentation_camera/GlutWindow.cc
```diff
@@ -0,0 +1,374 @@
...
+    else if (g_mouse.button == GLUT_MIDDLE_BUTTON && g_mouse.state == GLUT_DOWN)
+    {
+      for (auto & camera : g_cameras)
+      {
+        g_viewControl.SetCamera(camera);
+        g_viewControl.SetTarget(g_target.point);
+        g_viewControl.Orbit(drag);
+      }
+    }
+    // right mouse button zoom
```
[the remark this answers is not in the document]
AmrElsersy: No, it is not working yet, it crashes when using the RayQuery, I am working on it, but can't solve it till now :D

adlarkin:
When I run `ign gazebo shapes.sdf` using this PR, it crashes and I see the following message:
```
[GUI] [Err] [BaseNode.hh:224] Cannot add self as a child node
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
```
@AmrElsersy would you mind taking a closer look at this?

AmrElsersy:
> @AmrElsersy would you mind taking a closer look at this?
 
Working for me normally @adlarkin

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,118 @@
...
+
+
+namespace ignition
+{
+  namespace rendering
+  {
+    /// \brief Segmentation types for Semantic / Panpoptic segmentation
+    enum class SegmentationType {
+
+      /// \brief Pixels of same label from different items
```
adlarkin, line 35:
```suggestion
    enum class SegmentationType {
      /// \brief Pixels of same label from different items
```
nit: remove whitespace

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,118 @@
...
+      /// have the same color & id.
+      Semantic = 0,
+
+      /// \brief Pixels of same label from different items, have different
+      /// color & id. 1 channel for label id & 2 channels for instance id
+      Panoptic = 1
```
adlarkin, line 41: The variable names in the enum should all be uppercase - here's an example: https://github.com/ignitionrobotics/ign-rendering/blob/7b9ac811c6f6489935681047987291c4cb938b6b/include/ignition/rendering/ShaderType.hh#L34-L53

You'll also need to update any other repositories that use this enum to be uppercase as well :grimacing:

AmrElsersy:
The problem of [Multi Link Models](https://github.com/ignitionrobotics/ign-gazebo/pull/853#issuecomment-882320969) is solved by considering all ogre items which have the same parent name, to have the same material in the material switcher


solved_panoptic

osrf-triage:

https://community.gazebosim.org/t/gsoc-2021-machine-learning-extension-to-ignition-gazebo/1070/1

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,858 @@
...
+  this->colorToLabel.clear();
+  this->datablockMap.clear();
+  auto itor = this->scene->OgreSceneManager()->getMovableObjectIterator(
+      Ogre::ItemFactory::FACTORY_TYPE_NAME);
+
+  // Used for multi-link models, where each model has many ogre items but
+  // belongs to the same object, and all of them has the same parent name
+  std::string prevParentName = "";
+
+  // Store the ogre objects in a vecotr
```
iche033, line 291: `vecotr` -> `vector`

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,858 @@
...
+  auto engine = Ogre2RenderEngine::Instance();
+  auto ogreRoot = engine->OgreRoot();
+  this->dataPtr->ogreCompositorManager = ogreRoot->getCompositorManager2();
+
+  this->dataPtr->workspaceDefinition = "SegmentationCameraWorkspace_" +
+    this->Name();
+  auto backgroundColor = Ogre2Conversions::Convert(
+    this->dataPtr->materialSwitcher->backgroundColor);
+
+  // basic workspace consist of clear pass with the givin color &
```
iche033, line 619: `givin` -> `given`

include/ignition/rendering/base/BaseSegmentationCamera.hh
```diff
@@ -0,0 +1,181 @@
...
+    //////////////////////////////////////////////////
+    template <class T>
+    bool BaseSegmentationCamera<T>::IsColoredMap()
+    {
+      return false;
+    }
+
+    //////////////////////////////////////////////////
+    template <class T>
+    void BaseSegmentationCamera<T>::SetBackgroundColor(math::Color)
```
iche033, line 155: `const math::Color &`

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,117 @@
...
+          std::function<void(const uint8_t *, unsigned int, unsigned int,
+          unsigned int, const std::string &)>  _subscriber) = 0;
+
+      /// \brief Set Segmentation Type
+      /// \param[in] _type Segmentation Type
+      public: virtual void SetSegmentationType(SegmentationType _type) = 0;
+
+      /// \brief Get Segmentation Type
+      /// \return Segmentation Type
+      public: virtual SegmentationType GetSegmentationType() = 0;
```
iche033, line 78: Omit `Get` prefix. I think the problem you'll run into is that the function name conflicts with the enum type `SegmentationType`. For cases like this, this function can just be named `Type()`, [similar example here](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering5/include/ignition/rendering/Marker.hh#L104-L109)

include/ignition/rendering/SegmentationCamera.hh
```diff
@@ -0,0 +1,117 @@
...
+
+      /// \brief Enable Color map mode to generate colored semantics
+      /// \param[in] _enable True to generate colored map, False to generate
+      /// label id map
+      public: virtual void EnableColoredMap(bool _enable) = 0;
+
+      /// \brief Enable Color map mode to generate colored semantics
+      /// \return True if colored map, False if label id map
+      public: virtual bool IsColoredMap() = 0;
```
iche033, line 87: can this be `const`?

examples/segmentation_camera/GlutWindow.cc
iche033: what's the status of ray query issue?

How about adding a todo note if the issue will not be fixed in this PR?

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,858 @@
...
+      else if (this->type == SegmentationType::PANOPTIC)
+      {
+        auto itemName = visual->Name();
+        std::string parentName;
+
+        // Visuals that are added without a name (through rendering API)
+        // will have the name "scene::Visual(VISUAL_ID)", use it as a name
+        auto unNamedVisualIndex = itemName.find("scene::Visual");
+        if (unNamedVisualIndex != std::string::npos)
+          parentName = itemName.substr(0, itemName.find_first_of(")") + 1);
```
iche033, line 375: instead of searching for a hardcoded name string, how about checking to see if the visual shares the same parent as the previous visual, something like:

```
auto parent = visual->Parent();
if (parent && parent == prevParent)
  isMultiLink = true;
```

ogre2/src/Ogre2SegmentationCamera.cc
```diff
@@ -0,0 +1,858 @@
...
+
+  /// \brief Convert label of semantic map to a unique color for colored map
+  /// \param[in] _label id of the semantic map or encoded id of panoptic map
+  /// \param[in] _isMultiLink bool used to skip the taken color check if the
+  /// label is for a multi link model, as all links should have the same color
+  /// \return _color unique color in the colored map for that label
+  private: math::Color LabelToColor(int64_t _label,
+    bool _isMultiLink = false);
+
+  /// \brief Check if the color is taken previously
```
iche033, line 74: can you add a small comment that it also adds the color to the list if it's not taken?

iche033 (commented):
Did a first pass review. Nice use of ogre features.

include/ignition/rendering/base/BaseSegmentationCamera.hh
```diff
@@ -0,0 +1,181 @@
...
+      public: virtual bool IsColoredMap();
+
+      // Documentation inherited
+      public: virtual void SetBackgroundColor(math::Color _color);
+
+      // Documentation inherited
+      public: virtual void SetBackgroundLabel(int _label);
+
+      // Documentation inherited
+      public: virtual math::Color BackgroundColor();
```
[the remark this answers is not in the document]
adlarkin: > keep the function signature consistent

It also looks like this method should return a reference (`math::Color &`), not a copy: https://github.com/AmrElsersy/ign-rendering/blob/f4f376101f48bf4586630ab69d0e031123e8a1fc/include/ignition/rendering/SegmentationCamera.hh#L99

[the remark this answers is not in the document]
AmrElsersy: I cannot make it a reference as it needs to be implemented in the BaseSegmentationCamera class, which should return a referece for the color when we implement the function ... but we don't have that color yet, it is in the dataPtr object, so when I try to return a reference to a local variable it gives me a warning .. so i make it return a copy

ogre2/src/Ogre2SegmentationCamera.cc
AmrElsersy: @iche033 
mmm, but It is not the direct parent in all cases, I check if all the visuals has the same top parent, and it might be not the direct parent

for example, if we have a model which contains many links, so the visuals names will be something like:
`model::link1::visual1` & `model::link2::visual2` , so I need to check the top parent which is `model` but not the direct parent which is `link1` or `link2`

iche033: I see, how about adding a convenient function to get the top level model visual?

```cpp
////////////////////////////////////////////////
VisualPtr Ogre2SegmentationMaterialSwitcher::TopLevelVisual(VisualPtr _visual)
{ 
  if (!_visual)
    return _visual;
  VisualPtr p = _visual;
  while (p->Parent() && p->Parent() != _visual->Scene()->RootVisual())
    p = std::dynamic_pointer_cast<Visual>(p->Parent());
  return p;
}
```

then you can get the parent name here by calling 

```cpp
   parentName = this->TopLevelVisual(visual)->Name();
```

this is just to prevent relying on hardcoded name prefixes in case the visual naming style changes in the future.

src/SegmentationCamera_TEST.cc
```diff
@@ -0,0 +1,90 @@
...
+    igndbg << "Engine '" << _renderEngine
+              << "' is not supported" << std::endl;
+    return;
+  }
+
+  ScenePtr scene = engine->CreateScene("scene");
+  ASSERT_NE(nullptr, scene);
+
+  SegmentationCameraPtr camera(scene->CreateSegmentationCamera());
+  EXPECT_TRUE(camera != nullptr);
```
adlarkin, line 55:
```suggestion
  ASSERT_NE(nullptr, camera);
```

src/SegmentationCamera_TEST.cc
```diff
@@ -0,0 +1,90 @@
...
+  const std::string &_renderEngine)
+{
+  // create and populate scene
+  RenderEngine *engine = rendering::engine(_renderEngine);
+  if (!engine)
+  {
+    igndbg << "Engine '" << _renderEngine
+              << "' is not supported" << std::endl;
+    return;
+  }
```
adlarkin, line 49: We should probably make sure that `_renderEngine` is ogre2 since the segmentation camera has only been implemented in ogre2

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+    g_buffer = new uint8_t[bufferSize];
+
+  memcpy(g_buffer, _data, bufferSize);
+
+  g_counter++;
+  g_mutex.unlock();
+}
+
+/// \brief Build the scene with 3 boxes besides each other
+/// the 2 aside boxes have the same label & the middle is different
```
adlarkin, line 78:
```suggestion
/// the 2 outer boxes have the same label & the middle is different
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+void SegmentationCameraTest::SegmentationCameraBoxes(
+  const std::string &_renderEngine)
+{
+  // Optix is not supported
+  if (_renderEngine.compare("optix") == 0)
+  {
+    igndbg << "Engine '" << _renderEngine
+              << "' doesn't support depth cameras" << std::endl;
+    return;
+  }
```
adlarkin, line 125:
```suggestion
  // Only ogre2 is supported
  if (_renderEngine.compare("ogre2") != 0)
  {
    igndbg << "Engine '" << _renderEngine
              << "' doesn't support segmentation cameras" << std::endl;
    return;
  }
```
The only engine we currently support for segmentation camera is ogre2, so we should check for this before proceeding

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  // Setup ign-rendering with an empty scene
+  auto *engine = ignition::rendering::engine(_renderEngine);
+  if (!engine)
+  {
+    igndbg << "Engine '" << _renderEngine
+              << "' is not supported" << std::endl;
```
adlarkin, line 132:
```suggestion
    igndbg << "Failed to create scene with engine `" << _renderEngine << "`\n";
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+    return;
+  }
+
+  ignition::rendering::ScenePtr scene = engine->CreateScene("scene");
+  BuildScene(scene);
```
adlarkin, line 137:
```suggestion
  ignition::rendering::ScenePtr scene = engine->CreateScene("scene");
  ASSERT_NE(nullptr, scene);
  BuildScene(scene);
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+  camera->SetLocalPosition(0.0, 0.0, 0.0);
+  camera->SetLocalRotation(0.0, 0.0, 0.0);
+
+  int backgroundLabel = 23;
+  camera->SetBackgroundLabel(backgroundLabel);
+  EXPECT_EQ(camera->BackgroundLabel(), backgroundLabel);
+
+  camera->SetSegmentationType(SegmentationType::ST_SEMANTIC);
+
+  EXPECT_EQ(camera->Type(), SegmentationType::ST_SEMANTIC);
```
adlarkin, line 152:
```suggestion
  camera->SetSegmentationType(SegmentationType::ST_SEMANTIC);
  EXPECT_EQ(camera->Type(), SegmentationType::ST_SEMANTIC);
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  camera->EnableColoredMap(false);
+  EXPECT_FALSE(camera->IsColoredMap());
+
+  int width = 320;
+  int height = 240;
+  double aspectRatio = width/height;
```
adlarkin, line 159:
```suggestion
  double aspectRatio = static_cast<double>(width) / static_cast<double>(height);
```
In order to get proper division here, the ints need to be casted to doubles before dividing. Otherwise, integer division will occur, resulting in `1`, and then `1` will be converted to a double when assigned to `aspectRatio`. Here's an example of what I'm describing: https://godbolt.org/z/6dcMKqhrW

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+  // Set a callback on the  camera sensor to get a Segmentation camera frame
+  ignition::common::ConnectionPtr connection =
+      camera->ConnectNewSegmentationFrame(
+          std::bind(OnNewSegmentationFrame,
+          std::placeholders::_1, std::placeholders::_2, std::placeholders::_3,
+          std::placeholders::_4, std::placeholders::_5));
+  EXPECT_NE(nullptr, connection);
+
+  // Update once to create image
+  camera->Update();
```
adlarkin, line 178:
```suggestion
  // Update once to create image
  camera->Update();
  EXPECT_EQ(1, g_counter);
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+  // check the label
+  EXPECT_EQ(leftLabel  , 1);
+  EXPECT_EQ(middleLabel, 2);
+  EXPECT_EQ(rightLabel , 1);
+
+  // check if the first pixel(background) = the background label
+  int background = g_buffer[0];
+  EXPECT_EQ(background, backgroundLabel);
+
+  g_mutex.unlock();
```
adlarkin, line 210:
```suggestion
  // cast the unsigned char to unsigned int to read it
  uint8_t leftLabel =   g_buffer[leftIndex];
  uint8_t rightLabel =  g_buffer[rightIndex];
  uint8_t middleLabel = g_buffer[middleIndex];

  // check the label
  EXPECT_EQ(leftLabel  , 1);
  EXPECT_EQ(middleLabel, 2);
  EXPECT_EQ(rightLabel , 1);

  // check if the first pixel(background) = the background label
  int background = g_buffer[0];
  EXPECT_EQ(background, backgroundLabel);
```
I don't think you need to lock/unlock the mutex since you're not calling `camera->Update`

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  // Instance/Panoptic  Test
```
adlarkin, line 212:
```suggestion
  // Instance/Panoptic test
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+  camera->SetSegmentationType(SegmentationType::ST_PANOPTIC);
+
+  // Update once to create image
+  camera->Update();
```
adlarkin, line 216:
```suggestion
  // Update once to create image
  g_counter = 0;
  camera->Update();
  EXPECT_EQ(1, g_counter);
```

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+  EXPECT_EQ(leftLabel  , 1);
+  EXPECT_EQ(middleLabel, 2);
+  EXPECT_EQ(rightLabel , 1);
+
+  // instance count
+  EXPECT_EQ(middleCount, 1);
+  EXPECT_EQ(rightCount, 1);
+  EXPECT_EQ(leftCount, 2);
+
+  g_mutex.unlock();
```
adlarkin, line 242:
```suggestion
  // check the label
  EXPECT_EQ(leftLabel  , 1);
  EXPECT_EQ(middleLabel, 2);
  EXPECT_EQ(rightLabel , 1);

  // instance count
  EXPECT_EQ(middleCount, 1);
  EXPECT_EQ(rightCount, 1);
  EXPECT_EQ(leftCount, 2);
```
I don't think you need to lock/unlock the mutex since you're not calling `camera->Update`

test/integration/segmentation_camera.cc
```diff
@@ -0,0 +1,262 @@
...
+
+  // check the label
+  EXPECT_EQ(leftLabel  , 1);
+  EXPECT_EQ(middleLabel, 2);
+  EXPECT_EQ(rightLabel , 1);
+
+  // instance count
+  EXPECT_EQ(middleCount, 1);
+  EXPECT_EQ(rightCount, 1);
+  EXPECT_EQ(leftCount, 2);
```
adlarkin, line 240: Can you add a comment that explains why `rightCount` is 1 and `leftCount` is 2 instead of it being the other way around? Is it due to the names of the left and right boxes, and how they're processed in the material switcher?

ogre2/src/Ogre2SegmentationCamera.cc
adlarkin: @iche033 I went ahead and took your `TopLevelModelVisual` approach in https://github.com/ignitionrobotics/ign-rendering/pull/329/commits/4f0209d5697d7b58684a0772568c2e916fd1179f, and it seems to work well for models with multiple links. So, I'll mark this as resolved

ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
```diff
@@ -0,0 +1,321 @@
...
+        }
+        else
+        {
+          // 256 => 8 bits .. 255 => color percentage
+          float labelColor = label / 255.0;
+          float instanceColor1 = (instanceCount / 256) / 255.0;
+          float instanceColor2 = (instanceCount % 256) / 255.0;
+
+          customParameter = Ogre::Vector4(
+            instanceColor2, instanceColor1, labelColor, 1.0);
```
AmrElsersy, line 283: I think when we save the output image in the `ign-sensors::SegmentationCameraSensor` this will produce an Image with channels in the same order `(instanceColor2, instanceColor1, labelColor)` , and we don't want that, since the famous format of the panoptic map is to have the order `(labelColor, instanceColor1, instanceColor2)`

so why is that reversing ?

adlarkin: It was reversed as part of the changes made for `ogre2.2`. If I recall correctly, tests were failing before this change.

> the famous format of the panoptic map is to have the order `(labelColor, instanceColor1, instanceColor2)`

Are you saying that the label should be in the first channel, and the instance count should be the last two channels? If so, I think that the implementation you originally wrote is reverse. Looking at the integration test, the label is the last channel, and the instance count is the first 2 channels: https://github.com/AmrElsersy/ign-rendering/blob/segmentation/test/integration/segmentation_camera.cc#L215-L223

It's worth mentioning that although the order of the parameters were reversed for `Ogre2.2`, the code segment I linked above which checks for label/instance count did not change, so I believe that functionality is still the same.

adlarkin: Also, looking at your tutorial in `ign-sensors`, I see that the label is the last channel, and the instance count is in the first two channels: https://github.com/ignitionrobotics/ign-sensors/blob/7f8f045a7f39a13fccdb847b0aa55d70423cc3db/tutorials/05_segmentation_camera.md#processing-the-segmentation-sensor-via-ign-transport

So, I think that the change we had to make for Ogre2.2 is okay since this behavior of label in the last channel and instance count in first two channels is preserved

AmrElsersy: Yeah, It was reversed via the C++ API (when we loop over pixels), but when we get the image buffer and save it locally, the saved image was in the correct order (in the order of the parameters we passed `(labelColor, instanceColor1, instanceColor2)`) ... the order of the C++ API dosn't really matter, it is ok to be any order 

But ok I will check the order in the saved images again

adlarkin (approved):
The sensor itself seems to work well, but I did find a few issues with mouse functionality in the example. Since we want to get this in for Fortress, I'm approving this so that it can get merged, but I've ticketed an issue that outlines the mouse functionality issues in the example so that we can work on fixing them later: https://github.com/ignitionrobotics/ign-rendering/issues/414

adlarkin:
It looks like there's a new warning on ubuntu CI that's related to CMake configuration, but the warning doesn't provide any information, so I'm not sure how to fix it: https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/1799/cmake/new/

Has anyone seen this before?

iche033:
> It looks like there's a new warning on ubuntu CI that we should fix before merging:

could be related to https://github.com/ignitionrobotics/ign-rendering/issues/400, which is fixed in https://github.com/ignitionrobotics/ign-rendering/pull/411

ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
iche033: yeah it was reversed as the test was expecting the label to be in the 3rd channel. Same thing was done to the bounding camera when converting it to ogre 2.2. I never ran the tests before the conversion though. so I don't know if it was passing or not. How about adding a brief comment documenting what each component in the pixel is in the public API: https://github.com/ignitionrobotics/ign-rendering/blob/4ed1b40b0985b20e43388506080f1e268520e5d8/include/ignition/rendering/SegmentationCamera.hh#L62?

adlarkin:
> fixed in #411

Yeah, I think you're right. It looks like #411 hasn't been forward ported to main yet, so should we go ahead and merge this since the Ubuntu CI warning seems to be a false positive?

ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
adlarkin: > the test was expecting the label to be in the 3rd channel. Same thing was done to the bounding camera when converting it to ogre 2.2. I never ran the tests before the conversion though. so I don't know if it was passing or not.

Yes, you're right about the test expecting the label to be in the 3rd channel. The tests were passing before the conversion to ogre 2.2, so the changes you made for the ogre 2.2 conversion are okay :+1: 

> How about adding a brief comment documenting what each component in the pixel is in the public API?

Done: https://github.com/ignitionrobotics/ign-rendering/pull/329/commits/3e3711bd49c6641a5a78247eac71a93676e0481b

iche033 (approved):

PR #356, opened 2021-06-29
Fix lidar near plane clipping

iche033:

# 🦟 Bug fix

Fixes https://github.com/osrf/subt/issues/888

## Summary

In our GPU lidar implementation, the min range currently translates to the near clip plane of the cube camera used to sample the depth values of the scene. This is not correct. The problem is that objects that are near the corner of the image can be incorrectly clipped. 

For example, take a lidar with a min range of 1m.  An object is now placed at a position at `p = [0.9, 0.9, 0]` relative to the robot (0.9m in front of it and 0.9m to its left). The object should be visible because it is beyond the min range, i.e. `length(p)` is `sqrt(0.9*0.9 + 0.9*0.9) = 1.27`, which is larger than min range. However, the current implementation correctly clips `p` because the clipping plane is at 1m which is larger than the 0.9. 

To fix this, we set a smaller near clip plane for the cube camera and let the shaders handle the clipping of range values.

For comparison, take a look at the point clouds produced by the cerberus gagarin robot in https://github.com/osrf/subt/issues/888. The point clouds at the bottom is box shaped because of incorrect clipping. Here're the point clouds after the fix. It's now a spherical shape :

cerberus_gagarin_points

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -172,6 +172,9 @@ class ignition::rendering::Ogre2GpuRaysPrivate
   /// \brief Listener for setting particle noise value based on particle
   /// emitter region
   public: std::unique_ptr<Ogre2ParticleNoiseListener> particleNoiseListener[6];
+
+  /// \brief Actual near clip plane
+  public: float nearClipCube= 0.0;
 };
 
 using namespace ignition;
@@ -508,7 +511,7 @@ void Ogre2GpuRays::ConfigureCamera()
   this->SetRangeCount(this->RangeCount(), this->VerticalRangeCount());
 
   // Set ogre cam properties
-  this->dataPtr->ogreCamera->setNearClipDistance(this->NearClipPlane());
+  this->dataPtr->ogreCamera->setNearClipDistance(this->dataPtr->nearClipCube);
   this->dataPtr->ogreCamera->setFarClipDistance(this->FarClipPlane());
 }
 
@@ -628,9 +631,9 @@ void Ogre2GpuRays::Setup1stPass()
   // The projectParams is used to linearize depth buffer data
   // The other params are used to clamp the range output
   double projectionA = this->FarClipPlane() /
-      (this->FarClipPlane() - this->NearClipPlane());
-  double projectionB = (-this->FarClipPlane() * this->NearClipPlane()) /
-      (this->FarClipPlane() - this->NearClipPlane());
+      (this->FarClipPlane() - this->dataPtr->nearClipCube);
+  double projectionB = (-this->FarClipPlane() * this->dataPtr->nearClipCube) /
+      (this->FarClipPlane() - this->dataPtr->nearClipCube);
   projectionB /= this->FarClipPlane();
   psParams->setNamedConstant("projectionParams",
       Ogre::Vector2(projectionA, projectionB));
@@ -867,7 +870,7 @@ void Ogre2GpuRays::Setup1stPass()
     this->ogreNode->attachObject(this->dataPtr->cubeCam[i]);
     this->dataPtr->cubeCam[i]->setFOVy(Ogre::Degree(90));
     this->dataPtr->cubeCam[i]->setAspectRatio(1);
-    this->dataPtr->cubeCam[i]->setNearClipDistance(this->NearClipPlane());
+    this->dataPtr->cubeCam[i]->setNearClipDistance(this->dataPtr->nearClipCube);
     this->dataPtr->cubeCam[i]->setFarClipDistance(this->FarClipPlane());
     this->dataPtr->cubeCam[i]->setFixedYawAxis(false);
     this->dataPtr->cubeCam[i]->yaw(Ogre::Degree(-90));
@@ -1051,6 +1054,12 @@ void Ogre2GpuRays::Setup2ndPass()
 /////////////////////////////////////////////////////////
 void Ogre2GpuRays::CreateGpuRaysTextures()
 {
+  // make cube cam near clip smaller than specified and manually clip range
+  // values in 1st pass shader (gpu_rays_1st_pass_fs.glsl).
+  // This is so that we don't incorrectly clip the range values near the
+  // corners of the cube cam viewport.
+  this->dataPtr->nearClipCube = this->NearClipPlane() * 0.6f;
+
   this->ConfigureCamera();
   this->CreateSampleTexture();
   this->Setup1stPass();


ahcorde (changes requested):
```
  /github/workspace/ogre2/src/Ogre2GpuRays.cc:196:  Missing spaces around =  [whitespace/operators] [4]
```

peci1:
This PR almost completely breaks CTU_CRAS_NORLAB robots with Ouster lidars. I tested with Spot and MARV. I think it is because the newly defined near clip plane is inside the lidar's geometry. The lidar diameter is 0.0435 m, near clip is set to 0.1. If you take 0.1 * 0.6, it gives 0.06, which means the geometry should not obstruct the view, but apparently, it does (most of the time). I think there is something more related to https://github.com/ignitionrobotics/ign-sensors/issues/128 (see the screenshots with rotated robots). This PR makes the rotated behavior much worse.

Before: 
Snímek obrazovky pořízený 2021-06-29 10-58-13

Snímek obrazovky pořízený 2021-06-29 10-58-40

With this PR:

Snímek obrazovky pořízený 2021-06-29 10-42-37

Snímek obrazovky pořízený 2021-06-29 10-43-01

MARV is affected, too:

Snímek obrazovky pořízený 2021-06-29 11-07-21

Snímek obrazovky pořízený 2021-06-29 11-07-47

Just to get an idea how does the scaling parameter help, I also tried with `0.2 * NearClipPlane()`, but it seems the geometry obstruction is now complete:

Snímek obrazovky pořízený 2021-06-29 10-47-54

Snímek obrazovky pořízený 2021-06-29 10-48-24

Isn't there a simple way to tell the gpu lidar to ignore some geometries? At least those attached to the same link as the lidar sensor? (but that would make problems on models which are just a single link, like MARBLE_HUSKY). Or is it possible to turn on backface culling for the robot meshes?

Also, is there a way to visualize the output of the shaders (the depth images) before they go through the 1st and 2nd passes? That could help figuring out a better solution (or finding out where obstructing geometries are).

Increasing the near clip in SDF helps a bit, but still some false points are observed:

`0.1 -> 0.2`

rviz_screenshot_2021_06_29-11_17_16

`0.1 -> 0.5`

rviz_screenshot_2021_06_29-11_18_45

iche033:
> At least those attached to the same link as the lidar sensor? (but that would make problems on models which are just a single link, like MARBLE_HUSKY). 

yeah I think that may be a problem for other robots and cases where the geometry is meant to be visible.

> Or is it possible to turn on backface culling for the robot meshes?

the rendering engine should be doing culling, and just by looking at the Ouster lidar mesh, the mesh actually has triangles facing inwards, hence they are seen by the camera if it's beyond the near clip plane.

I think one solution is to set visibility mask/flags for the lidar and the ouster lidar visual. That way we can tell the sensor to ignore the ouster lidar mesh, and would also be a solution for ignitionrobotics/ign-sensors#128. This would require changes in sdf and ign-sensors, and the robot model.sdf. I'll think about this and see if there is an easier way

iche033:
> Increasing the near clip in SDF helps a bit, but still some false points are observed:

I noticed that. For testing, I removed the sensor mesh and the false points are gone. We def get some weird values when the sensor is inside another geometry.

peci1:
> We def get some weird values when the sensor is inside another geometry.

Which is, however, very common. I think this deserves a generally usable and good working solution. Or a tutorial for mesh design. Or an automatic warning in console when Gazebo detects all/most points from the lidar projected onto the minimum distance.

One general way I was thinking about was removing triangles which intersect the min-distance sphere (maybe a little inflated). But I don't know how to implement it.

darksylinc:
This problem sounds like it needs a custom piece in HlmsPbs (Ogre 2.1+) to apply custom clipping math.

Whenever the interpolated attribute of a triangle is gl_ClipDistance[i] < 0 (for any i) the triangle at that pixel will not be rendered.

This problem seems like it would need simple math:
`gl_ClipDistance[0] = distance( vertexWorldSpacePos, cameraPos ) - passBuf.minLidarRange;`

That way geometry will be clipped spherically, while being able to pull the near plane closer without worrying about geometry of the robot that's between the virtual camera pos and the lidar's min range.

## How do you want to proceed?

1. I create a demo showing how to customize HlmsPbs with a custom minLidarRange? (and someone else ports it to ign-rendering)
2. I customize HlmsPbs directly on ign-rendering, and someone points me which camera class needs minLidarRange and when the customization should be toggled on and when it should be toggled off again

iche033:
> I customize HlmsPbs directly on ign-rendering, and someone points me which camera class needs minLidarRange and when the customization should be toggled on and when it should be toggled off again

I think we can go for this option. It's great we can do this without modifying ogre code. The envCams in Ogre2GpuRays are the ones that will need this custom clipping:
https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/ogre2/src/Ogre2GpuRays.cc#L870

for now, custom clipping should always be on for these env cameras but not for other cameras. Ideally after the customization is done, we can easily apply them to other cameras later if needed, e.g. we are still considering if we need to do this for Ogre2DepthCamera.

Note that we are currently doing additional clipping in the shaders so I think we'll need to remove that logic: https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/ogre2/src/media/materials/programs/gpu_rays_1st_pass_fs.glsl#L125-L126

darksylinc:
> I think we can go for this option. It's great we can do this without modifying ogre code. The envCams in Ogre2GpuRays are the ones that will need this custom clipping:
> https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/ogre2/src/Ogre2GpuRays.cc#L870

Great!

> for now, custom clipping should always be on for these env cameras but not for other cameras. Ideally after the customization is done, we can easily apply them to other cameras later if needed, e.g. we are still considering if we need to do this for Ogre2DepthCamera.

Ok!

> Note that we are currently doing additional clipping in the shaders so I think we'll need to remove that logic: https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering4/ogre2/src/media/materials/programs/gpu_rays_1st_pass_fs.glsl#L125-L126

From what you're saying and what I can see, yes that test _should_ be redundant.

> I think the clip distance changes won't break API / ABI right?

You asked this one via email but I'm replying it here: I don't think so. API and ABI should not break.
The solution is also fairly simple, not complex code.

peci1:
The latest added commit (7fb493b) made a difference - it removed the ball around the lidar, but the far points are still not displayed with the robot rotated.

iche033:
Here's the latest result testing with a couple of different robots using [robots_near_clip.sdf](https://gist.github.com/iche033/7b52dfece48168f90a83b33ebd76aafa):
robots_near_clip

The lidar points are good at different angles, and the bottom of the lidar visualization is spherical shape (as opposed to box shaped). So I think this PR is good to go as is now.

Other notes:

I just realized that depth clamping is actually not enabled for the gpu lidar sensor. It was done for depth and thermal cameras in order to get the tests to pass - though it's not a desirable fix, see https://github.com/ignitionrobotics/ign-rendering/issues/395 for more details on issues with depth clamping enabled. However, turns out  `INTEGRATION_gpu_lidar` test is passing without enabling depth clamping. I think what's happening is that the `nearCubeClip` clip plane added in this PR is serving as a second near clip plane similar to the idea described in this comment: https://github.com/ignitionrobotics/ign-rendering/issues/395#issuecomment-917427922 so it's able to clip the sensor geometry while detecting objects closer than lidar's min range. This needs to be verified.

darksylinc:
There's a lot going on so I'll try to be as clear as I can be:

There's 3 things I can identify going on:

1. The original problem was caused by depth clamping being force enabled. That's why it didn't work correctly on ogre2, but it did work fine in ogre1. I took too long to notice this (I noticed after the Hlms customization was integrated...)
2. Additionally you pushed the near plane closer to the camera, because of the diagonals. This helps with the unit test as the near plane is now closer to camera
3. Additionally my Hlms customizations added a "spherical" clipping; which ignores depth clamping as it clips before sending vertices to the pixel shader. These customization is still useful, as it can be used to implement multi-tiered "near plane distances" as proposed in #395 to filter bad geometry setup blocking the sensors without having to fix the model.

iche033:
great, thanks for the clarification. I added more notes in to the code to document this.

iche033:
I increased the minimum texture size of the for the raw depth texture that data is sampled from. The currently setup does not work well for large FOV with low sample count. After testing with a 360 degree lidar and 4 samples, I found that I had to increase the min texture size to 128x128 (from 2x2) in order to produce more correct range results. 1f50023

darksylinc:
Rasterization-based GPUs (vs raytraced-based GPUs) aren't efficient when low resolutions are involved thus I wouldn't be surprised there's not much perf difference between 2x2 and 128x128, nonetheless it'd need to be profiled

PR #370, opened 2021-07-26
(out-of-date) Prevent default-constructed variants from holding a type

adlarkin:

# 🦟 Bug fix

Fixes https://github.com/ignitionrobotics/ign-rendering/pull/334#discussion_r647923514

## Summary
The [`std::variant` documentation](https://en.cppreference.com/w/cpp/utility/variant) states:
```
a default-constructed variant holds a value of its first alternative, unless that alternative is not default-constructible (in which case the variant is not default-constructible either).
```

This can be problematic if a user calls [`Visual::UserData`](https://github.com/ignitionrobotics/ign-rendering/blob/ignition-rendering3_3.5.0/include/ignition/rendering/Visual.hh#L125) with a key that does not exist, because this will [return a default-constructed variant](https://github.com/ignitionrobotics/ign-rendering/blob/ignition-rendering3_3.5.0/include/ignition/rendering/base/BaseVisual.hh#L347). The expected behavior should be that for calls to `Visual::UserData` with a key that does not exist, an "empty variant" is returned (i.e., a variant which holds no alternatives - see [`std::holds_alternative`](https://en.cppreference.com/w/cpp/utility/variant/holds_alternative)).

The fix proposed here is similar to what is proposed in https://github.com/ignitionrobotics/ign-rendering/pull/334#issuecomment-882382753. [`std::monostate`](https://en.cppreference.com/w/cpp/utility/variant/monostate) is used to guarantee empty alternatives for a default-constructed variant.

For a running example that shows the difference between the current behavior and proposed behavior using `std::monostate`, see https://godbolt.org/z/hcGM815Kh.

## Checklist
- [x] Added tests

### DIFF
--- a/include/ignition/rendering/Visual.hh
+++ b/include/ignition/rendering/Visual.hh
@@ -28,7 +28,15 @@ namespace ignition
   {
     inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
     //
-    using Variant = std::variant<int, float, double, std::string>;
+    /// \brief Alias for a variant that can holds various types of data.
+    /// The first type of the variant is std::monostate in order to prevent
+    /// default-constructed variants from holding a type (a default-constructed
+    /// variant is returned when a user calls Visual::UserData with a key that
+    /// doesn't exist for the visual. In this case, since the key doesn't
+    /// exist, the variant that is returned shouldn't hold any types - an
+    /// "empty variant" should be returned for keys that don't exist)
+    using Variant = std::variant<std::monostate, int, float, double,
+          std::string>;
 
     /// \class Visual Visual.hh ignition/rendering/Visual.hh
     /// \brief Represents a visual node in a scene graph. A Visual is the only
--- a/src/Visual_TEST.cc
+++ b/src/Visual_TEST.cc
@@ -17,6 +17,7 @@
 
 #include <gtest/gtest.h>
 #include <string>
+#include <variant>
 
 #include <ignition/common/Console.hh>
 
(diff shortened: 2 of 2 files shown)


include/ignition/rendering/Visual.hh
```diff
@@ -28,7 +28,15 @@ namespace ignition
...
-    using Variant = std::variant<int, float, double, std::string>;
+    /// \brief Alias for a variant that can holds various types of data.
+    /// The first type of the variant is std::monostate in order to prevent
+    /// default-constructed variants from holding a type (a default-constructed
+    /// variant is returned when a user calls Visual::UserData with a key that
+    /// doesn't exist for the visual. In this case, since the key doesn't
+    /// exist, the variant that is returned shouldn't hold any types - an
+    /// "empty variant" should be returned for keys that don't exist)
+    using Variant = std::variant<std::monostate, int, float, double,
+          std::string>;
```
chapulina, line 39: We can't change this without breaking ABI :confused: 

https://build.osrfoundation.org/job/ignition_rendering-abichecker-any_to_any-ubuntu_auto-amd64/646/API_5fABI_20report/

adlarkin: Ohhh you're right! I didn't even think about ABI here. I think that means we can't fix this issue in any released version then :disappointed_relieved: I wish we could at least add a method to the `Visual` class like `HasUserData`, and then encourage users to call this first before calling `UserData` in order to prevent using keys that don't exist. However, doing this would break ABI too, according to the [KDE ABI guide](https://community.kde.org/Policies/Binary_Compatibility_Issues_With_C%2B%2B):
```
You cannot...
* add new virtual functions to non-leaf classes as this will break subclasses
```

So, I don't think we can do anything about this in released versions. Should I re-target this to Fortress? If so, is there anything we should do about this for released versions (update documentation, create an issue, etc.)?

chapulina: Yeah I think we can target this at Fortress. For earlier versions, the only thing I can think of is creating our own "invalid variant" and documenting it. For example, we could return an `std::string("_invalid_")` or something like that. And document it...

adlarkin: > we could return a `std::string("_invalid_")` or something like that. And document it

I think this is a good idea, but my only concern is that this is technically a behavioral breaking change, which we try to stay away from for released versions. So, are we even able to do anything about this issue for released versions? Or can it only be addressed starting in Fortress?

chapulina: That's a good point, downstream users may be already compensating for the current behaviour, so we could break them. I'm ok targeting this at Fortress and documenting the migration.

adlarkin: Sounds good, I will close this since I have opened a new PR for fortress. See #371

adlarkin:
Closing this since it's an ABI break and we cannot fix this bug without a behavior breaking change - see https://github.com/ignitionrobotics/ign-rendering/pull/370#discussion_r677021521 for more details.

PR #358, opened 2021-07-02
UserData methods moved from Visual to Node

ahcorde:
# 🦟 Bug fix

Fixes https://github.com/ignitionrobotics/ign-rendering/issues/58

## Summary

This PR is part of the effort to consolidate `ign-gui` and `ign-gazebo` (related to this conversation https://github.com/ignitionrobotics/ign-gui/pull/231#discussion_r662702693) and it also targets this issue https://github.com/ignitionrobotics/ign-rendering/issues/58. 

The idea is to move the `UserData` and `SetUserData` methods to the Node class. I'm not sure how to proceed about the change in the API if I should tick-tock the methods in the Visual class, this change should not affect the API because `Node` is a parent class of `Visual`, it will break ABI but this PR is targeting `main`.

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #371, opened 2021-07-30
Prevent default-constructed variants from holding a type

adlarkin:

# 🦟 Bug fix

## Summary
This is a follow-up to #370 (see that PR for a description of the bug fix). Since #370 breaks ABI and we cannot make this change in released versions without changing behavior and potentially breaking downstream users (relevant discussion: https://github.com/ignitionrobotics/ign-rendering/pull/370#discussion_r677021521), this fix is targeted for the next unreleased version, which is Fortress.

I also added some missing test coverage/updated documentation related to #358.

## Checklist
- [x] Added tests
- [x] Updated migration guide (as needed)

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -32,6 +32,11 @@ release will remove the deprecated code.
     + Added `Scene::SetCameraPassCountPerGpuFlush`. Setting this value to 0 forces legacy behavior which eases porting.
     + Systems that rely on Graphics components like particle FXs and postprocessing are explicitly affected by Scene's Pre/PostRender. Once `Scene::PostRender` is called, the particle FXs' simulation is moved forward, as well as time values sent to postprocessing shaders. In previous ign-rendering versions each `Camera::Render` call would move the particle simulation forward, which could cause subtle bugs or inconsistencies when Cameras were rendering the same frame from different angles. Setting SetCameraPassCountPerGpuFlush to 0 will also cause these subtle bugs to reappear.
 
+1. **Visual.hh** and **Node.hh**
+    + `*UserData` methods and the `Variant` type alias have been moved from the `Visual` class to the `Node` class.
+    `Node::UserData` now returns no data for keys that don't exist (prior to Rendering 6.x, if
+    `Visual::UserData` was called with a key that doesn't exist, an `int` was returned by default).
+
 ## Ignition Rendering 4.0 to 4.1
 
 ## ABI break
--- a/src/Visual_TEST.cc
+++ b/src/Visual_TEST.cc
@@ -350,6 +350,29 @@ void VisualTest::UserData(const std::string &_renderEngine)
   value = visual->UserData(stringKey);
   EXPECT_EQ(stringValue, std::get<std::string>(value));
 
+  // bool
+  std::string boolKey = "bool";
+  bool boolValue = true;
+  visual->SetUserData(boolKey, boolValue);
+  value = visual->UserData(boolKey);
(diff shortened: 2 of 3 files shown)


src/Visual_TEST.cc
```diff
@@ -350,6 +350,29 @@ void VisualTest::UserData(const std::string &_renderEngine)
...
+
+  // test a key that does not exist (should return no data)
+  value = visual->UserData("invalidKey");
+  EXPECT_FALSE(std::holds_alternative<int>(value));
+  EXPECT_FALSE(std::holds_alternative<float>(value));
+  EXPECT_FALSE(std::holds_alternative<double>(value));
+  EXPECT_FALSE(std::holds_alternative<std::string>(value));
+  EXPECT_FALSE(std::holds_alternative<bool>(value));
+  EXPECT_FALSE(std::holds_alternative<unsigned int>(value));
+
```
adlarkin, line 375: The tests for the `*UserData` methods are still in `Visual_TEST.cc`, but since the `*UserData` methods have been moved to the `Node` class as of #358, should the tests be moved to `Node_TEST.cc` as well?

iche033: one of the reasons is that there is no `scene->CreateNode()` function and you can not just instantiate a node directly. It maybe worth considering letting users create nodes later as a lightweight version of visual. For now, I think it's fine to leave it here in Visual_TEST .

Migration.md
```diff
@@ -32,6 +32,11 @@ release will remove the deprecated code.
     + Added `Scene::SetCameraPassCountPerGpuFlush`. Setting this value to 0 forces legacy behavior which eases porting.
     + Systems that rely on Graphics components like particle FXs and postprocessing are explicitly affected by Scene's Pre/PostRender. Once `Scene::PostRender` is called, the particle FXs' simulation is moved forward, as well as time values sent to postprocessing shaders. In previous ign-rendering versions each `Camera::Render` call would move the particle simulation forward, which could cause subtle bugs or inconsistencies when Cameras were rendering the same frame from different angles. Setting SetCameraPassCountPerGpuFlush to 0 will also cause these subtle bugs to reappear.
 
+1. **Visual.hh** and **Node.hh**
+    + `*UserData` methods and the `Variant` type alias have been moved from the `Visual` class to the `Node` class.
+    `Node::UserData` now returns no data for keys that don't exist (prior to Rendering 6.x, if
+    `Visual::UserData` was called with a key that doesn't exist, an `int` was returned by default).
```
iche033, line 38: an `int` with a value of `0`?

iche033 (approved):
looks good to me.

Migration.md
adlarkin: Is it safe to always assume "value of `0`"? The [`std::variant` documentation](https://en.cppreference.com/w/cpp/utility/variant) states, _"a default-constructed variant holds a value of its first alternative"_, which makes me think that the `int` (the first alternative in this case) is also default-initialized. If that's the case, there's technically no guarantee that the `int` will be initialized to `0`. But maybe the variant initializes alternatives to a specific value? I'm not sure.

iche033: I'm not sure too. I saw the comment https://github.com/ignitionrobotics/ign-rendering/pull/334#issuecomment-882382753 that says it'll always return 0. Maybe that's not always true. We could leave the doc here as is if we don't know the answer.

adlarkin: Okay, I'll leave it as-is because I think it's `0` most of the time, but I'm not sure if that's a guarantee.

PR #388, opened 2021-08-24
Fix particle effect randomness

iche033:

# 🦟 Bug fix

## Summary

The randomness algorithm in the particle effect for has issue when `time` (which is used to add randomness) becomes large in the shaders. The fix is to just use the fractional part of the `time` value for adding randomness.

## To Test

To reproduce the problem, one way to test this is to artificially increase the `time` value in the shaders. You can do this by applying the patch below to this branch to simulate the old behavior when time is large:

```cpp
diff --git a/ogre2/src/media/materials/programs/depth_camera_fs.glsl b/ogre2/src/media/materials/programs/depth_camera_fs.glsl
index 0d3109d5..a22bd4b3 100644
--- a/ogre2/src/media/materials/programs/depth_camera_fs.glsl
+++ b/ogre2/src/media/materials/programs/depth_camera_fs.glsl
@@ -107,7 +107,7 @@ void main()
   if (particle.x > 0 && particleDepth < fDepth)
   {
     // apply scatter effect so that only some of the smoke pixels are visible
-    float r = rand(inPs.uv0 + vec2(rnd, rnd));
+    float r = rand(inPs.uv0 + vec2(rnd + 1e5, rnd + 1e5));
     if (r < particleScatterRatio)
     {
       // set point to 3d pos of particle pixel
```

Then run this example world:
https://gist.github.com/iche033/bcd3b7d3f4874e1e707e392d6dbb0aa0

The patch adds 100000 seconds (27.x hours) to time and you'll see bands across the rgbd camera :

rgbd_particles_large_t

The changes in this PR makes sure the random number never exceeds `1` and the noise pattern should stay distributed across the image:
rgbd_particles

## Checklist
- [ ] Added tests


darksylinc:
I knew there was something amiss with that line!

Your solution is a temporary workaround though. Once time becomes large enough, the fractional part becomes always 0 (and before that, it will always be 0 and 0.5, and before that...).

It's [easier to mentally visualize it](https://en.wikipedia.org/wiki/Half-precision_floating-point_format) with half floating point format

A better solution would be to fix it from C++ side of things.

If you're using Ogre's auto param, IIRC there was an option to make it periodic

iche033:
> A better solution would be to fix it from C++ side of things.

yeah you're right, once time becomes too large, the randomness goes away. I took the C++ approach and passed a random number to the shaders

ahcorde (approved):

PR #392, opened 2021-09-01
Add SetSize API for LidarVisual

iche033:

# 🎉 New feature

## Summary

Support setting point size for lidar visual. 

~Currently working for ogre 1.x only. I still need to figure out how to make it work with the ogre2 hlms system.~

**Update**: ogre2 is now working - updated to use ogre low level materials.

## Test it

1. Build and run the [lidar_visual](https://github.com/ignitionrobotics/ign-rendering/tree/lidar_point_size/examples/lidar_visual) example. 
2. Press `2` to toggle to point rendering and see the points are now slightly larger than before (size increased from 1 to 5):

lidar_visual_size_5

## Checklist
- [x] Added tests


include/ignition/rendering/LidarVisual.hh
```diff
@@ -171,6 +171,15 @@ namespace ignition
       /// \return The type for lidar visual
       public: virtual LidarVisualType Type() const = 0;
 
+      /// \brief Set size of the lidar visualization, e.g. size of lidar points
```
chapulina, line 174: Is there a unit for this? Is it "pixels" even though it's a double?

chapulina (commented):
Nice! Can we also make the point size in markers configurable? :pray:

include/ignition/rendering/LidarVisual.hh
iche033: updated doc. 43c4328 It's in pixels. It eventually comes down to setting `gl_PointSize` which is a float and I saw some examples setting it to a floating point value so I left it here as a double.

iche033:
> Can we also make the point size in markers configurable?

I created pull request #403 for setting marker point size. I notice that we don't have any gui plugin or msg field to configure the point size though. That'll maybe a god feature to add afterwards.

ogre2/src/Ogre2DynamicRenderable.cc
```diff
@@ -356,6 +359,11 @@ void Ogre2DynamicRenderable::UpdateBuffer()
       this->dataPtr->ogreItem->setCastShadows(
           this->dataPtr->material->CastShadows());
     }
+    else if (lowLevelMat)
+    {
+      this->dataPtr->ogreItem->getSubItem(0)->setMaterial(lowLevelMat);
+      this->dataPtr->ogreItem->setCastShadows(castShadows);
+    }
```
adlarkin, line 366: Is this needed? It looks like you're setting the material and castShadows to whatever was already stored here based on the `getCastShadows` and `getMaterial` calls above. So, I'm trying to understand if calling `setMaterial` and `setCastShadows` actually changes anything for `this->dataPtr->ogreItem`.

adlarkin (commented):
LGTM. I played with the example to change the value passed in to the `SetSize` call, and things behaved as expected with both ogre and ogre2. When running the example with ogre2, I saw the `Error casting user data: Unexpected index` message, but this went away when I tested the example with #396.

I left a few minor comments, and wouldn't mind having one more person take a look at this to confirm that the usage of materials/shaders here is okay.

examples/lidar_visual/Main.cc
```diff
@@ -213,6 +213,7 @@ LidarVisualPtr createLidar(ScenePtr _scene)
   lidar->SetMaxVerticalAngle(vMaxAngle);
   lidar->SetMaxRange(maxRange);
   lidar->SetMinRange(minRange);
+  lidar->SetSize(5.0);
```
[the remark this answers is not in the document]
iche033: good catch, I've updated the README file. 0d43d87

ogre2/src/Ogre2DynamicRenderable.cc
iche033: the [ogreItem->_initialise](https://github.com/ignitionrobotics/ign-rendering/blob/0d43d873ecabcf401153cab11be2d20f8f6dc74f/ogre2/src/Ogre2DynamicRenderable.cc#L351) call a few lines above resets those values so I had to set them again. I added a comment to explain this. 0d43d87

adlarkin (approved):
As I mentioned before, it'd be nice to have at least one more set of eyes on this since my rendering knowledge is still fairly elementary. But, from what I tested/reviewed, LGTM!

WilliamLewww (approved):

PR #396, opened 2021-09-07
Check if key exists in gpu lidar's user data

### LINKED ISSUE #394: Error casting user data: Unexpected index
Chasing [this bug](https://github.com/ignitionrobotics/ign-rendering/pull/272#issuecomment-913022327) where INTEGRATION_lidar_visual prints:

> [Err] [Ogre2GpuRays.cc:262] Error casting user data: Unexpected index

in stderr which aren't happening in Edifice which ended up not being a rendering bug, I noticed that if I force retroValue = -1 the items from INTEGRATION_lidar_visual still do get rendered using standard PBS shaders in both Edifice (2.1) and Fortress (2.2).

These reveals two possible bugs:

1. `laser_retro` **is not set anywhere**. except `ign-gazebo/src/rendering/SceneManager.cc` and `ign-rendering/test/integration/gpu_rays.cc` but neither of them are set during INTEGRATION_lidar_visual test. Due to [this change](https://github.com/ignitionrobotics/ign-rendering/pull/371/files), it appears that previously by default Variant would return int 0, whereas now a default Variant will throw when trying to cast to any type because none is set.

2. If I force retroValue = -1 the items from INTEGRATION_lidar_visual still do get rendered using standard PBS shaders even though the code later messes with visibility flags:
```cpp
if (retroValue >= 0)
{
  // set visibility flag so the camera can see it
  item->addVisibilityFlags(0x01000000);
  // ...
}
```

which suggests that the intended path is that items whose retroValue is not set should not be rendered; however they are being rendered.

## Environment
* OS Version: Ubuntu 18.04 LTS
* Source or binary build?
  Source. `main` branch

## Description
* Expected behavior: No errors printed on stderr. Likely appropriate objects should switch to use the custom low level material while the rest is not rendered (instead of being rendered using standard PBS).
* Actual behavior in `main`: Errors are printed on stderr and objects are being rendered using PBS shaders
* Behavior in Edifice: No stderr. All objects are switched to low level material because `laser_retro` defaults to 0 when unset. However if an object explicitly sets `laser_retro` to a negative value, objects will still be rendered using PBS, which is likely a bug.

## Steps to reproduce

1. Run INTEGRATION_lidar_visual in main

## Output

Lots of:

> [Err] [Ogre2GpuRays.cc:262] Error casting user data: Unexpected index

Test still passes anyway but looks like random chance. Inspecting the RenderTarget shows results are quite different

Reference result (uses custom low level material shader):
reference

Reference wireframe:
reference_wireframe

Failure result (uses PBS shaders. I don't think this element should be getting rendered at all!):
failure

Failure wireframe:
failure_wireframe

iche033:

# 🦟 Bug fix

Partially fixes #394

## Summary

The problem is described in #394. Summary is that the gpu lidar is flooding the console with `Error casting user data: Unexpected index` message because because objects in the scene do not have [laser_retro](http://sdformat.org/spec?ver=1.8&elem=visual#visual_laser_retro) set. This is actually ok as not all objects contain `laser_retro` values. This property is [set in ign-gazebo](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo5/src/rendering/SceneManager.cc#L281).

This PR adds a `HasUserData` API to check if a UserData key exists or not. This avoid additional parsing in gpu lidar to figure out if the user data contains valid value or not.

The second issue identified in #394 is that the visibility flags set to objects with `laser_retro` do not have effect. 
> the intended path is that items whose retroValue is not set should not be rendered; however they are being rendered.

That's found to be the case. This issue is not fixed in this PR as that turns out to require more work. I've removed the visibility mask/flag logic as they are not doing their job, and added a todo in the code.

## Checklist
- [x] Added tests


chapulina:
It's worth pointing out that since #371, `UserData` will return an invalid variant if it's unset. So downstream callers should be able to check something like `if (!userData.index() == 0)`. I don't think that's very user friendly though, so an explicit `HasUserData` function sounds like a good addition.

iche033:
> So downstream callers should be able to check something like if (!userData.index() == 0)

true, I added a check in the test in bade294 as a reminder that this also works.

adlarkin (approved):

chapulina (approved):

PR #418, opened 2021-09-21
Minor tweaks to visual cloning 

iche033:

Made a few minor tweaks to API and documentation

### DIFF
--- a/include/ignition/rendering/base/BaseMesh.hh
+++ b/include/ignition/rendering/base/BaseMesh.hh
@@ -100,12 +100,14 @@ namespace ignition
 
       public: virtual void PreRender() override;
 
+      // Documentation inherited.
       public: virtual GeometryPtr Clone() const override;
 
-      public: const MeshDescriptor &GetMeshDescriptor() const override;
+      // Documentation inherited.
+      public: void SetDescriptor(const MeshDescriptor &_desc) override;
 
-      public: virtual void CopyMeshDescriptor(
-                  const MeshDescriptor &_copy) override;
+      // Documentation inherited.
+      public: const MeshDescriptor &Descriptor() const override;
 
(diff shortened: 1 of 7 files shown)


include/ignition/rendering/base/BaseMesh.hh
```diff
@@ -370,18 +370,16 @@ namespace ignition
...
 
     //////////////////////////////////////////////////
     template <class T>
-    void BaseMesh<T>::CopyMeshDescriptor(const MeshDescriptor &_copy)
+    void BaseMesh<T>::SetDescriptor(const MeshDescriptor &_desc)
     {
-      this->meshDescriptor.meshName = _copy.meshName;
-      this->meshDescriptor.subMeshName = _copy.subMeshName;
-      this->meshDescriptor.centerSubMesh = _copy.centerSubMesh;
+      this->meshDescriptor = _desc;
```
adlarkin, line 382: Does this also copy the pointer to the mesh? As we discussed offline, I thought that we wanted to ignore storing/keeping track of this pointer

adlarkin (approved):
LGTM! Just one question before merging

include/ignition/rendering/base/BaseMesh.hh
iche033: yeah I took another look at the implementation and looks like the `common::Mesh` pointer is managed by the `common::MeshManager` and the pointer remains valid for the duration of the simulation session so it maybe ok. I don't have a strong opinion on this though

adlarkin: As long as the pointer is valid, I am fine with storing it. Since it's a pointer, it doesn't take up much memory anyways. I say go ahead and merge it!

PR #420, opened 2021-09-22
add bounding box sensor API stubs

adlarkin:
## Summary
I've added the public API for a bounding box sensor. Most of this is copy/pasted from #334, with a few changes:
1. There's only one constructor for the `rendering::BoundingBox` struct, which takes a type and label argument
2. The `rendering::BoundingBoxCamera::DrawBoundingBox` method takes the box color as an argument instead of defaulting to green

This PR just covers the breaking changes that are needed for the bounding box sensor so that we can add #334 to Fortress in a minor release. Some of the things I added here may not be breaking, but I wasn't sure, so I added whatever I thought could be ABI breaking. See #334 for an ogre2 implementation.

## Test it
No tests have been added here because tests are written for this in #334

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/rendering/BoundingBoxCamera.hh
+++ b/include/ignition/rendering/BoundingBoxCamera.hh
@@ -24,12 +24,15 @@
 #include <ignition/math/Color.hh>
 #include <ignition/math/Vector3.hh>
 
+#include "ignition/rendering/BoundingBox.hh"
 #include "ignition/rendering/Camera.hh"
 
 namespace ignition
 {
   namespace rendering
   {
+    inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
+    //
     /// \brief BoundingBox types for Visible / Full 2D Boxes / 3D Boxes
     enum class BoundingBoxType
     {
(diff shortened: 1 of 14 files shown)


include/ignition/rendering/BoundingBoxCamera.hh
```diff
@@ -0,0 +1,162 @@
...
+    class IGNITION_RENDERING_VISIBLE BoundingBoxCamera :
+      public virtual Camera
+    {
+      /// \brief Destructor
+      public: virtual ~BoundingBoxCamera() { }
+
+      /// \brief Get the BoundingBox data
+      /// \return Buffer of bounding boxes info (label, minX, minY, maxX, maxY)
+      public: virtual const std::vector<BoundingBox> &BoundingBoxData()
+              const = 0;
```
adlarkin, line 136: Would it be better to return a copy here instead of a reference in case users want to do things with the bounding box data? I also wonder if a copy is better so that if the bounding box data changes frequently, users don't have to worry about this since they have a copy

adlarkin (commented):
I'd like some feedback on an API decision I made before this gets merged. See my comment below.

include/ignition/rendering/BoundingBoxCamera.hh
chapulina: const ref is fine, the caller can always make a copy if they want to, and this way we're not forcing the copy.

iche033: I think this API gives user a cons ref and they also have the option to make a copy of it if they like right? So if they plan to store the vector and do something with the data, they'll just need to do an addition call to manually copy it, e.g. with std::copy

include/ignition/rendering/BoundingBoxCamera.hh
```diff
@@ -0,0 +1,162 @@
...
+
+      /// \brief Size of the box (width, height, depth), depth = 0 in 2D boxes
+      ignition::math::Vector3d size;
+
+      /// \brief Orientation of the 3D box in camera coord.
+      /// The 2D boxes are axis aligned (orientation = 0)
+      ignition::math::Quaterniond orientation;
+
+      /// \brief Label of the annotated object inside the box
+      uint32_t label;
```
chapulina, line 65: Just keep in mind that adding the members directly here, we won't be able to extend it in the future without breaking ABI. Using the PIMPL pattern would solve the issue.

include/ignition/rendering/BoundingBoxCamera.hh
```diff
@@ -0,0 +1,162 @@
...
+      /// occluded object
+      BBT_VISIBLEBOX2D = 1,
+
+      /// 3D oriented box
+      BBT_BOX3D = 2
+    };
+
+    /// \brief 2D or 3D Bounding box. It stores the
+    /// position / orientation / size info of the box and its label
+    struct BoundingBox
```
chapulina, line 49: I don't remember why we went with a custom struct, but I'll just bring up again that we could use `ignition::math::Box` (it doesn't have a `Vertices` function yet though :confused:

chapulina (commented):
Let's also add the dependency on `ign-math` eigen. It's best not to add dependencies to stable versions.

include/ignition/rendering/BoundingBoxCamera.hh
adlarkin: Sounds good, thanks for the feedback :+1: I will leave it as-is, returning a const ref

include/ignition/rendering/BoundingBoxCamera.hh
adlarkin: What if we had a struct like this:
```c++
struct BoundingBox
{
  /// \brief The bounding box
  ignition::math::Boxd box;

  /// \brief The label of the object represented by the bounding box
  uint32_t label;

  /// \brief Get the vertices of the box
  std::vector<math::Vector3d> Vertices() const
};
```

That way, we avoid some of the duplication but still add things like `label` and `Vertices` that `ignition::math::Box` doesn't currently have.

include/ignition/rendering/BoundingBoxCamera.hh
adlarkin: Is PIMPL used in `ign-rendering` since it has abstract interfaces in `include/ignition/rendering` and base implementations in `include/ignition/rendering/base`? I don't often see classes implemented in the `src` directory. But, I think using PIMPL is a good idea if we'd like to give ourselves flexibility down the road

chapulina: See the [HeightmapDescriptor](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering5/include/ignition/rendering/HeightmapDescriptor.hh) for an example. Let me know if I can help with anything

include/ignition/rendering/BoundingBoxCamera.hh
chapulina: I like that

adlarkin:
> Let's also add the dependency on ign-math eigen

Added in https://github.com/ignitionrobotics/ign-rendering/pull/420/commits/98a7bf10ee49c615371c13afa432d68b45fa79f0. I'm assuming that the ogre2 component can be updated in #334 since that's a plugin, right? Here's what I am referring to in #334: https://github.com/ignitionrobotics/ign-rendering/blob/4162fd7b95969691d6ef4918d795172a66e3053a/ogre2/src/CMakeLists.txt#L38

include/ignition/rendering/BoundingBoxCamera.hh
```diff
@@ -0,0 +1,162 @@
...
+    {
+      /// \brief Box type
+      ignition::rendering::BoundingBoxType type;
```
chapulina, line 52: Does the bounding box really need a type? Shouldn't the type belong to the camera?

include/ignition/rendering/BoundingBoxCamera.hh
chapulina: Done in a1103b52011a1545dc202370be90cd806169a55e

include/ignition/rendering/BoundingBoxCamera.hh
chapulina: We can decide this later now that the `BoundingBox` class is PIMPLized.

include/ignition/rendering/BoundingBoxCamera.hh
chapulina: The `BoundingBox` class is now on a separate file, so if the types get defined on this header, using it there would create a circular dependency

adlarkin: > Does the bounding box really need a type? Shouldn't the type belong to the camera?

I think you're right. The type should belong to the camera. I say we remove the `type` from the `BoundingBox` class for now. The nice thing is that since you modified this class to use PIMPL, if we find for whatever reason that we need to add this field back in later, we should be able to do so without issues.

chapulina (approved):
I hope that's all we need :crossed_fingers:

PR #441, opened 2021-09-29
Hide heightmap visibility for segmentation camera

adlarkin:
# 🦟 Bug fix

## Summary
The segmentation camera appears to show parts of heightmaps, even if they don't have a label. This shouldn't happen. I'm trying to make a temporary patch for this by disabling the heightmap visibility during material switching.

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
+++ b/ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
@@ -23,6 +23,7 @@
 
 #include <ignition/common/Console.hh>
 
+#include "ignition/rendering/ogre2/Ogre2Heightmap.hh"
 #include "ignition/rendering/ogre2/Ogre2Scene.hh"
 #include "ignition/rendering/ogre2/Ogre2Visual.hh"
 #include "ignition/rendering/RenderTypes.hh"
@@ -207,6 +208,21 @@ void Ogre2SegmentationMaterialSwitcher::cameraPreRenderScene(
       Ogre2VisualPtr ogreVisual = std::dynamic_pointer_cast<Ogre2Visual>(
         visual);
 
+      // segmentation doesn't work with heightmaps for now, so disable the
+      // visibility during material switching
+      // TODO(anyone) add support for segmentation with heightmaps
+      for (auto i = 0u; i < ogreVisual->GeometryCount(); ++i)
+      {
+        if (std::dynamic_pointer_cast<Heightmap>(
+              ogreVisual->GeometryByIndex(i)))
+        {
+          this->heightmapVisFlags[ogreVisual->Id()] =
+            ogreVisual->VisibilityFlags();
+          ogreVisual->SetVisibilityFlags(0x00000000);
+          break;
+        }
+      }
+
       // get class user data
       Variant labelAny = ogreVisual->UserData("label");
       int label;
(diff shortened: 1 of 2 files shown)


ogre2/src/Ogre2SegmentationMaterialSwitcher.cc
```diff
@@ -207,6 +208,21 @@ void Ogre2SegmentationMaterialSwitcher::cameraPreRenderScene(
       Ogre2VisualPtr ogreVisual = std::dynamic_pointer_cast<Ogre2Visual>(
         visual);
 
+      // segmentation doesn't work with heightmaps for now, so disable the
+      // visibility during material switching
+      // TODO(anyone) add support for segmentation with heightmaps
+      for (auto i = 0u; i < ogreVisual->GeometryCount(); ++i)
+      {
+        if (std::dynamic_pointer_cast<Heightmap>(
+              ogreVisual->GeometryByIndex(i)))
```
adlarkin, line 217: When testing this with the fortress demo world, it looks like this `if` statement is never true based on the debugging I've done. I'm not sure if this is the right way to check for if a visual is a heightmap, but my reason for taking this approach is because [a heightmap is a geometry](https://github.com/ignitionrobotics/ign-rendering/blob/1dbdef4e877a6356dcb1bc0478d2207308025c6b/include/ignition/rendering/Heightmap.hh#L32-L33), so I assume that if a visual has a heightmap geometry, then it's safe to say that the visual represents a heightmap.

Is what I'm doing here the correct way to check for a heightmap? If so, I'm not sure why this is never `true` when debugging. Or is there a better way to check for if a visual is a heightmap?

iche033: I think it failed to find the heightmap earlier in the code when looping through the ogre movable objects. It can't get the heightmaps because that's a custom ogre movable object that's not registered with the ogre scenemanager. 

I did a quick working around in my branch that sets a unique visibility flags for the heightmap in the Ogre2Heightmap class then set the Ogre2SegmentationCamera to a mask that does not contain the heightmap flags.

adlarkin: I just tested your branch, but I still see a part of the heightmap in the segmentation camera. It looks like your approach is similar to what's being done for things like ignoring the particle emitter output in the thermal camera, so I think it's probably on the right track. Should we close this PR and open a new one off of your branch, and then move this discussion to the new PR?

iche033: hmm I think it's fine, we can leave this PR open since my approach also didn't work.

Weird that I don't see the heightmap in the segmentation camera view. I'll try move it around more.

iche033: I just implemented a different approach. Can you update the `iche033/heightmap_segmentation_patch` branch and give that a try?

adlarkin: > Weird that I don't see the heightmap in the segmentation camera view. I'll try move it around more.

Did you not see any parts of the heightmap at all when testing `iche033/heightmap_segmentation_patch` with commit https://github.com/ignitionrobotics/ign-rendering/commit/b912ffd01638746207c97118889334eb55b2e953? I noticed a small part of the heightmap when testing with commit https://github.com/ignitionrobotics/ign-rendering/commit/b912ffd01638746207c97118889334eb55b2e953.

> I just implemented a different approach. Can you update the `iche033/heightmap_segmentation_patch` branch and give that a try?

It looks like the new approach (using commit https://github.com/ignitionrobotics/ign-rendering/commit/967118a982bb94f11b384a67fc63e3f98938fa81) hides the heightmap successfully for me!

iche033: > Did you not see any parts of the heightmap at all when testing iche033/heightmap_segmentation_patch with commit b912ffd?

I was testing with a minimal example world that I created but didn't see the heightmap in the segmentation camera view. I scrapped that and tested with the demo and I was able to reproduce the issue. Glad to hear that works now.

adlarkin: :+1: feel free to either open a new PR with your changes, or merge them in here. If it's in a separate PR, I can give it a final review/approval so that we can merge it in time for Fortress.

iche033: sure, I opened a new PR: https://github.com/ignitionrobotics/ign-rendering/pull/443.

adlarkin:
I'm closing this in favor of #443

PR #442, opened 2021-09-29
Don't clone visuals with an ignore flag

adlarkin:

## Summary
While working on https://github.com/ignitionrobotics/ign-gazebo/pull/1013, I noticed that there are times where certain visuals shouldn't be cloned. I've added support for setting a boolean `UserData` flag with a key of `skip-visual-clone` on a visual. If a visual has `UserData` with this key, it will not be cloned.

For an example of how this is useful, see the second half of https://github.com/ignitionrobotics/ign-gazebo/pull/1013#issuecomment-919564106.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/rendering/Visual.hh
+++ b/include/ignition/rendering/Visual.hh
@@ -149,6 +149,10 @@ namespace ignition
       /// \param[in] _newParent Parent of the cloned Visual. Set to nullptr if
       /// the cloned visual should have no parent.
       /// \return The visual. nullptr is returned if cloning failed.
+      /// \note If the visual to be cloned has boolean UserData with a key
+      /// "skip-visual-clone", nullptr is returned. If the visual to be cloned
+      /// does not have this UserData, but a child visual does, then the child
+      /// visuals with this UserData will not be cloned.
       public: virtual VisualPtr Clone(const std::string &_name,
                   NodePtr _newParent) const = 0;
     };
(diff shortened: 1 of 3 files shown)


include/ignition/rendering/Visual.hh
```diff
@@ -149,6 +149,10 @@ namespace ignition
       /// \param[in] _newParent Parent of the cloned Visual. Set to nullptr if
       /// the cloned visual should have no parent.
       /// \return The visual. nullptr is returned if cloning failed.
+      /// \note If the visual to be cloned has boolean UserData with a key
+      /// "skip-visual-clone", nullptr is returned. If the visual to be cloned
+      /// does not have this UserData, but a child visual does, then the child
+      /// visuals with this UserData will not be cloned.
```
iche033, line 155: oh I was thinking of also checking the user data in the ign-gazebo level (in SceneManager::CopyVisual function) and not inside ign-rendering we tend to use UserData internally in ign-rendering as a temporary workarounds.  I think the proper way for doing filtering when cloning is probably pass in a filter function that returns true if a particular visual should be cloned, e.g. similar to the predicate logic in `std::copy_if`. Something to consider in the next release as it break API / ABI.

How about not advertising this in the API doc so we don't promise this behavior and add a todo in the implementation that this is a workaround for ignoring visuals when cloning?

adlarkin:
I'm closing this PR because this functionality is now a part of `ign-gazebo`: see https://github.com/ignitionrobotics/ign-gazebo/pull/1103

PR #483, opened 2021-11-02
Enable ray query for depth, segmentation and thermal cameras

### LINKED ISSUE #414: Mouse functionality for segmentation camera demo is incomplete
## Environment
* OS Version: Ubuntu 18.04
* Source or binary build?
  Source: #329
    - Rendering plugin: ogre2

## Description
There are several mouse features that don't work with the segmentation camera demo:
1. Ray query code causes a crash: https://github.com/ignitionrobotics/ign-rendering/blob/e185ecd74c3dd979a06740d54fd010bf146eb15b/examples/segmentation_camera/GlutWindow.cc#L160
2. Mouse wheel scrool zoom doesn't work: https://github.com/ignitionrobotics/ign-rendering/blob/e185ecd74c3dd979a06740d54fd010bf146eb15b/examples/segmentation_camera/GlutWindow.cc#L170-L171
3. Right mouse button zoom doesn't work: https://github.com/ignitionrobotics/ign-rendering/blob/e185ecd74c3dd979a06740d54fd010bf146eb15b/examples/segmentation_camera/GlutWindow.cc#L206-L207

## Steps to reproduce

1. build the segmentation camera demo example:
```
cd ign-rendering/examples/segmentation_camera
mkdir build
cd build
cmake .. && make
```
2. Run the example and try to use the scroll wheel or right click button on your mouse (they won't work):
```
./segmentation_camera
```
3. Try to uncomment the ray query code (point 1 in the `Description` section above) and the rebuild/re-test the example. A crash occurs when you click on the GUI display

srmainwaring:
# 🦟 Bug fix

Fixes #414

## Summary

This PR enables depth, segmentation and thermal cameras to set a ray query for the `ogre` and `ogre2` render engines.

The approach Introduces mix-in classes to provide access to underlying Ogre objects in the `OgreCamera` / `Ogre2Camera` and related camera classes.

Update the mouse handling in the `segmentation_camera` and `thermal_camera examples`.

ogre_ray-query_thermal-camera

## Checklist
- [ ] Added tests
- [x] Updated migration guide (as needed)

### DIFF
--- a/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
@@ -97,10 +97,6 @@ namespace ignition
       public: virtual void Destroy() override;
 
       // Documentation inherited.
-      public: virtual Ogre::MovableObject* OgreMovableObject(
-          const char* _typename) const override;
-
-      // Documentation inherited.
       public: virtual Ogre::Camera *Camera() const override;
 
       /// \brief Get a pointer to the render target.
--- a/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
@@ -112,6 +112,7 @@ namespace ignition
       private: std::unique_ptr<OgreThermalCameraPrivate> dataPtr;
 
       private: friend class OgreScene;
+      private: friend class OgreRayQuery;
     };
     }
   }
--- a/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
@@ -117,7 +117,6 @@ namespace ignition
       private: std::unique_ptr<OgreThermalCameraPrivate> dataPtr;
 
       private: friend class OgreScene;
-      private: friend class OgreRayQuery;
     };
     }
   }
--- a/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
+++ b/ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
@@ -32,6 +32,7 @@
 #include "ignition/rendering/ogre/Export.hh"
 #include "ignition/rendering/ogre/OgreConversions.hh"
 #include "ignition/rendering/ogre/OgreIncludes.hh"
+#include "ignition/rendering/ogre/OgreObjectInterface.hh"
 #include "ignition/rendering/ogre/OgreRenderTarget.hh"
 #include "ignition/rendering/ogre/OgreRenderTypes.hh"
 #include "ignition/rendering/ogre/OgreScene.hh"
@@ -61,7 +62,8 @@ namespace ignition
     **/
     /// \brief Depth camera used to render thermal data into an image buffer
     class IGNITION_RENDERING_OGRE_VISIBLE OgreThermalCamera :
-      public BaseThermalCamera<OgreSensor>
+      public virtual BaseThermalCamera<OgreSensor>,
+      public virtual OgreObjectInterface
     {
       /// \brief Constructor
       protected: OgreThermalCamera();
@@ -93,6 +95,13 @@ namespace ignition
 
       // Documentation inherited
       public: virtual void Destroy() override;
+
+      // Documentation inherited.
+      public: virtual Ogre::MovableObject* OgreMovableObject(
+          const char* _typename) const override;
+
+      // Documentation inherited.
+      public: virtual Ogre::Camera *Camera() const override;
 
       /// \brief Get a pointer to the render target.
       /// \return Pointer to the render target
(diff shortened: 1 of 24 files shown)


srmainwaring:
Converting to draft: while this works I'm concerned that using RTTI will result in maintenance problems later.

Using a virtual function to access the `Ogre::Camera` would be better, but it's not clear to me the on best way to alter the inheritance structure or introduce a mix-in class.

iche033:
looks like the problem is that the depth, thermal, segmentation camera etc does not derive from `OgreCamera` / `Ogre2Camera` so it can't access the `camera` member variable. 

Not sure what the consequence of changing them to derive from `OgreCamera` / `Ogre2Camera` but that's one possible option. 

Another workaround I can think off is to add a function function called `OgreObject()` to `OgreSensor` / `Ogre2Sensor` that returns an `Ogre::MovableObject` object, and in the derived Ogre/Ogre2 classes just return the `Ogre::Camera` object. In the ray query class, you can then call the `OgreObject()` function and cast it to a `Ogre::Camera`.

open to other ideas.

srmainwaring:
> Another workaround I can think off is to add a function function called `OgreObject()` to `OgreSensor` / `Ogre2Sensor` that returns an `Ogre::MovableObject` object, and in the derived Ogre/Ogre2 classes just return the `Ogre::Camera` object. In the ray query class, you can then call the `OgreObject()` function and cast it to a `Ogre::Camera`

This might be a good way forward. It retains some type safety and doesn't require a change to the inheritance structure. All the ignition cameras contain the ogre camera object required so can implement the function. We could add a parameter to request the type of object needed (similar to the `getCustomObject()` method on the `Ogre::TextureGpu` class that we added for Metal support). That allows for other objects to be requested if needed later without a further change to the interface.  I'll update the PR with the changes for review.

iche033:
sounds good, thanks

srmainwaring:
I took a slightly different tack here: `OgreSensor` / `Ogre2Sensor` is concerned with scene nodes rather than movable objects. Rather than alter their purpose I've introduced small abstract mix-in classes to expose the underlying Ogre objects and implemented their methods in the cameras. In some cases these coincide with existing non-virtual methods on the camera class.

The mix-in class has two accessors. A specific one to access an `Ogre::Camera` and a general one to access an `Ogre::MovableObject`. I have not implemented the general case (so could remove it), but it may be useful to have if other underlying objects are required and you don't want to break ABI.

Coming up with a mix-in class name that makes sense and does not clash with existing objects is always a challenge, I went with `OgreObjectInterface` which is a bit clumsy and I'm not tied to it by any means. The camera access method has been chosen to coincide with any existing accessor already used (which happen to be different for the Ogre and Ogre2 render engines).

ogre2/include/ignition/rendering/ogre2/Ogre2ObjectInterface.hh
```diff
@@ -0,0 +1,50 @@
...
+#include "ignition/rendering/ogre2/Ogre2Includes.hh"
+
+namespace ignition
+{
+  namespace rendering
+  {
+    inline namespace IGNITION_RENDERING_VERSION_NAMESPACE {
+
+    /// \brief Mixin class to provide direct access to Ogre objects.
+    class Ogre2ObjectInterface
```
iche033, line 31: I think you'll need to add `IGNITION_RENDERING_OGRE2_VISIBLE` here for ignition to export the symbols, i.e.

`class IGNITION_RENDERING_OGRE2_VISIBLE Ogre2ObjectInterface`

iche033 (commented):
ok yeah your approach is fine. Given that the `OgreMovableObject` accessor function is currently not used, one suggestion is to make it non-pure-virtual and just return `nullptr` inline so that the derived classes do not have to implement it. I think it may break ABI if the derived classes decide to override the function later but I'm leaning towards reducing the overhead of adding new camera classes for now.

ogre2/include/ignition/rendering/ogre2/Ogre2ObjectInterface.hh
srmainwaring: Added export declarations and provided a default implementation for the `OgreMovableObject` method. Removed empty methods from the camera classes.

I'm not sure that a `const char* _typename` parameter is the best choice, something like the `Ogre::IdString` would be better but I'm not sure of the Ignition equivalent. Given that the function is a placeholder to signal intent, and that any changes to implement the method in camera classes would require a major release in any case, it's probably ok to leave it as is for now.

ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
```diff
@@ -112,6 +117,7 @@ namespace ignition
       private: std::unique_ptr<OgreThermalCameraPrivate> dataPtr;
 
       private: friend class OgreScene;
+      private: friend class OgreRayQuery;
```
iche033, line 120: is the `friend class` still needed now that it's using `OgreObjectInterface`? Same goes for other ogre camera headers

iche033 (commented):
works well for me. Just one minor comment, otherwise good to go.

ogre/include/ignition/rendering/ogre/OgreThermalCamera.hh
srmainwaring: You're right - they're not needed now. Originally I had the new members as protected but changed them to public to match the access level of `OgreCamera()` in `Ogre2Camera` etc. I'll remove all the friend classes that are no longer required.

iche033 (approved):

PR #503, opened 2021-12-04
Fix loading grayscale emissive map 

### LINKED ISSUE #499: Emissive map no longer works in ign-rendering6 (ogre2)
## Environment
* OS Version: Ubuntu 18.04
* Source or binary build?
  source build: `ign-rendering6` branch
 
## Description
* Expected behavior: objects are rendered with emissive map
* Actual behavior: emissive map has no effect

I think the feature is broken when we ported from ogre 2.1 to 2.2.

## Steps to reproduce
1. Test with ignition gazebo's fuel_textured_mesh.sdf world
  ```
  ign gazebo -v 4 fuel_textured_mesh.sdf
  ```
2. The tubes in the `Tube Light` model do not appear to be lit when emissive map is used.

Here is a comparison of the Tube Light model in Ignition Fortress (top) and Dome (bottom)

emissive_map_fortress_vs_dome

## Things I've tried

The call to set ogre2 emissive map is done here:
https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/ogre2/src/Ogre2Material.cc#L473

* I verified that `this->ogreDatablock->getUseEmissiveAsLightmap()` returns `false` and `this->ogreDatablock->_hasEmissive()` returns `true`
* setting emissive values (`this->ogreDatablock->setEmissive`) works 
* I tried debugging the code in [800.PixelShader_piece_ps.any](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/ogre2/src/media/Hlms/Pbs/Any/Main/800.PixelShader_piece_ps.any#L507-L526) for applying emissive map and it looks like the `emissive_map` property is not set hence it does not affect the final color.

iche033:
# 🦟 Bug fix

Fixes #499 

## Summary

Continuing the effort from #501, this PR fixes loading grayscale emissive map by converting it to RGB color texture so that the visual with the emissive map is not incorrectly rendered as red color.

Test by lauching ign gazebo with the `fuel_textured_mesh.sdf` world file.

Before:

tube_grayscale_emissive_before

After:

tube_grayscale_emissive_after

## Alternative considered

The other option considered is to some customization in the Pbs Hlms pixel shader and just apply only the r channel of the emissive texture to the final color, .e.g. [changing the code here](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/ogre2/src/media/Hlms/Pbs/Any/Main/800.PixelShader_piece_ps.any#L521) to `finalColour += emissiveCol.rrr;`. 

I went with the approach in this PR mainly to keep the changes within one function without touching the shaders so we don't deviate too much from the original Pbs shader code we took from ogre. This approach was also easier for me to to implement.

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -27,8 +27,6 @@
 #include <OgreHlmsManager.h>
 #include <OgreMaterialManager.h>
 #include <OgrePixelFormatGpuUtils.h>
-#include <OgreStagingTexture.h>
-#include <OgreTextureBox.h>
 #include <OgreTextureGpuManager.h>
 #include <Vao/OgreVaoManager.h>
 #ifdef _MSC_VER
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -650,16 +650,16 @@ void Ogre2Material::SetTextureMapImpl(const std::string &_texture,
     // check for 8 bit pixel
     if (img.BPP() == 8u)
     {
-      std::string p = common::parentPath(_texture);
+      std::string parentPath = common::parentPath(_texture);
       // set a custom name for the rgb texture by appending ign_ prefix
       std::string rgbTexName = "ign_" + baseName;
       baseName = rgbTexName;
-      std::string filename = common::joinPaths(p, rgbTexName);
+      std::string filename = common::joinPaths(parentPath, rgbTexName);
       auto tex = textureMgr->findTextureNoThrow(rgbTexName);
       if (!tex)
       {
         ignmsg << "Grayscale emissive texture detected. Converting to RGB: "
-            << rgbTexName << std::endl;
+               << rgbTexName << std::endl;
         // need to be 4 channels for gpu texture
         unsigned int channels = 4u;
         unsigned int size = img.Width() * img.Height() * channels;
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -541,7 +541,8 @@ void Ogre2Material::ClearLightMap()
   this->lightMapUvSet = 0u;
 
   // in ogre 2.2, we swtiched to use the emissive map slot for light map
-  this->ogreDatablock->setTexture(Ogre::PBSM_EMISSIVE, this->lightMapName);
+  if (this->ogreDatablock->getUseEmissiveAsLightmap())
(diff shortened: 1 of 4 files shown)


ogre2/src/Ogre2Material.cc
```diff
@@ -636,16 +637,82 @@ void Ogre2Material::SetTextureMapImpl(const std::string &_texture,
...
+      std::string p = common::parentPath(_texture);
+      // set a custom name for the rgb texture by appending ign_ prefix
+      std::string rgbTexName = "ign_" + baseName;
+      baseName = rgbTexName;
+      std::string filename = common::joinPaths(p, rgbTexName);
+      auto tex = textureMgr->findTextureNoThrow(rgbTexName);
+      if (!tex)
+      {
+        ignmsg << "Grayscale emissive texture detected. Converting to RGB: "
+            << rgbTexName << std::endl;
```
ahcorde, line 662:
```suggestion
        ignmsg << "Grayscale emissive texture detected. Converting to RGB: "
               << rgbTexName << std::endl;
```

ogre2/src/Ogre2Material.cc
```diff
@@ -636,16 +637,82 @@ void Ogre2Material::SetTextureMapImpl(const std::string &_texture,
...
+  // workaround for grayscale emissive texture
+  // convert to RGB otherwise the emissive map is rendered red
+  if (_type == Ogre::PBSM_EMISSIVE &&
+      !this->ogreDatablock->getUseEmissiveAsLightmap())
+  {
+    common::Image img(_texture);
+    // check for 8 bit pixel
+    if (img.BPP() == 8u)
+    {
+      std::string p = common::parentPath(_texture);
```
ahcorde, line 653:
```suggestion
      std::string parentDirectory = common::parentPath(_texture);
```
more descriptive name?

ahcorde (changes requested):

darksylinc:
Sounds like this is something that should be fixed from Upstream.

Detect if emissive is greyscale, then use a shader with the right swizzle

ogre2/src/Ogre2Material.cc
iche033: a6796ac

ogre2/src/Ogre2Material.cc
iche033: a6796ac

iche033:
great, thanks for implementing grayscale emissive map support upstream! I'm going to try get this in first, and remove later when I get time to port the fix over to our fork (and package / release new debs etc)

ahcorde (approved):

Issue #504, opened 2021-12-06
Port custom materials to Hlms pieces


darksylinc:
## Desired behavior

Right now specific passes [such as the Segmentation Camera](https://github.com/ignitionrobotics/ign-rendering/issues/444) iterates through all objects, changes their material to a custom low level materials, and renders that.

This works but has a few issues:

 - It cannot handle all details that HlmsPbs/HlmsUnlit does (e.g. HW skeletal animation)
 - It doesn't handle custom Hlms implementations (such as Terra's)
 - It can be slow to change the materials so often (switch it on every object to render, then restore them back)
 - It isn't API portable (it needs one material per API i.e. D3D11/Metal/GL)

Ogre +2.1 supports Hlms pieces, which allows users to add custom code that can be run before/during/after our default code runs.

This solves all outstanding issues:

 - It's fast
 - No need to switch materials (we'd still need to iterate through objects to disable those unwanted, unless we use visibility flags instead, but that's out of scope of this ticket)
 - It's API portable
 - It works with (nearly?) all Hlms implementations
 - All features supported by the underlying implementation that affect vertex position (i.e. HW skeletal animation) are automatically supported.

## Alternatives considered

This is the recommended way.

## Implementation suggestion

I'll be the one implementing this feature

iche033:
yeah that'll be nice go with the recommended way. Originally, we tried material schemes but couldn't get it to work in ogre 2.1. We then resorted to this workaround and have been doing it in many places of our code, e.g. [MaterialSwitcher](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/ogre2/src/Ogre2MaterialSwitcher.cc) used by selection buffer. So it would nice to go with the proper solution.

darksylinc:
The changes are currently on https://github.com/darksylinc/ign-rendering/tree/matias-hlmsCustom

## A few takes:

- It's going to be slighter harder than I realized; because we have to send custom data per object (not a huge deal given that it's only 1 set of float4)
- Iteration through objects is still needed so we can set those values to the object. What changes is that we no longer need to swap out materials (the rest of the benefit remain, e.g. API cross portability, vertex shader-based deformation such as HW skinning, etc)
  - Not iterating through objects every time we need the segmentation camera looks doable but that'd be a refactor from Ign side, rather than interface through Ogre. I won't be addressing that now; and anyone could do it

darksylinc:
[...]

PR #517, opened 2021-12-19
Port Segmentation Camera material to Hlms pieces

darksylinc:
This PR _partially_ implements #504 but is ready for inclusion.

I could add more stuff but I believe it's better to split it into smaller PRs to make review and testing more manageable

# 🦟 Bug fix

See #504 
(it doesn't close it, but merging this PR gets us closer)

## Summary

What was done:

- Ogre2SegmentationMaterialSwitcher now longer switches materials
  - It's still an important class because it sets the right values to each Renderable
  - But now the naming is obviously misleading
  - Hlms customizations take care of the shader part; which allows:
    - Faster performance for not having to switch materials twice per frame (Ogre-Next is not optimized for fast shader switching)
    - Features like skeletal skinning would not work with the old material switcher
 - Harmonized shader customizations. `HlmsPbsTerraShadows` and `Ogre2IgnHlmsCustomizations` were conflicting. They can now be applied simultaneously.
 - `HlmsPbs` and `HlmsUnlit` are now derived by `IgnHlmsPbs` & `IgnHlmsUnlit` to manage all listener-based customizations in a centralized way and provide per-renderable customizations (listener-based customizations can only affect objects in a global state; but cannot send data per renderable which is what Segmentation Camera needs)
 - `Ogre2IgnHlmsCustomizations` had PUBLIC visibility instead of PRIVATE. This was fixed. `Ogre2RenderEngine::HlmsCustomizations()` can return a `Ogre2IgnHlmsCustomizations&` reference but its header was never exposed to users
 - `Ogre2RenderEnginePrivate` is now explicitly HIDDEN (otherwise we get compiler warnings)

What is missing and will be implemented in follow up PRs:

 - Rename `Ogre2IgnHlmsCustomizations` to `Ogre2SphericalClipMinDistance`
 - Perform the same tasks for other material switchers such as `Ogre2SelectionBuffer`, `Ogre2ThermalCameraMaterialSwitcher`, `Ogre2LaserRetroMaterialSwitcher` and possibly others
 - Apply the same customizations to `HlmsTerra` (i.e. heightmaps)
   - This is the main endgoal, given that heightmaps can't be easily implemented through material switchers

## Checklist
- [ ] Added tests


darksylinc:
# Question:

Should `IgnHlmsPbs` & `IgnHlmsUnlit` be called `Ogre2IgnHlmsPbs` & `Ogre2IgnHlmsUnlit` instead?

ogre2/src/Ogre2IgnHlmsPbsPrivate.cc
```diff
@@ -0,0 +1,224 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include "Ogre2IgnHlmsPbsPrivate.hh"
+
+#include "Terra/Hlms/PbsListener/OgreHlmsPbsTerraShadows.h"
```
ahcorde, line 20: alphabetize

ogre2/src/Ogre2IgnHlmsPbsPrivate.hh
```diff
@@ -0,0 +1,178 @@
...
+        SceneManager *_sceneManager,
+        Hlms *_hlms) override;
+
+    /// \brief Tells Ogre the buffer data sent to GPU should be a little
+    /// bigger to fit our data we need to send
+    /// \param[in] _shadowNode see base class
+    /// \param[in] _casterPass see base class
+    /// \param[in] _dualParaboloid see base class
+    /// \param[in] _sceneManager see base class
+    public: virtual uint32 getPassBufferSize(
```
ahcorde, line 102: return ?

ogre2/src/Ogre2IgnHlmsPbsPrivate.hh
```diff
@@ -0,0 +1,178 @@
...
+        const Ogre::CompositorShadowNode *_shadowNode,
+        bool _casterPass, bool _dualParaboloid,
+        Ogre::SceneManager *_sceneManager) const override;
+
+    /// \brief Sends our custom data to GPU buffers that our
+    /// pieces activated in preparePassHash will need.
+    ///
+    /// Bytes written must not exceed what we informed in getPassBufferSize
+    /// \param _casterPass see base class
```
ahcorde, line 111:
```suggestion
    /// \param _shadowNode
    /// \param _casterPass see base class
    /// \param _dualParaboloid
```

ahcorde: Use in `\param`: `[in]`, `[inout]` or `[out]`

ogre2/src/Ogre2IgnHlmsUnlitPrivate.hh
```diff
@@ -0,0 +1,175 @@
...
+        bool _dualParaboloid,
+        SceneManager *_sceneManager,
+        Hlms *_hlms) override;
+
+    /// \brief Tells Ogre the buffer data sent to GPU should be a little
+    /// bigger to fit our data we need to send
+    /// \param[in] _shadowNode see base class
+    /// \param[in] _casterPass see base class
+    /// \param[in] _dualParaboloid see base class
+    /// \param[in] _sceneManager see base class
```
ahcorde, line 98: `\return` ?

ogre2/src/Ogre2IgnHlmsUnlitPrivate.hh
```diff
@@ -0,0 +1,175 @@
...
+    public: virtual uint32 getPassBufferSize(
+        const Ogre::CompositorShadowNode *_shadowNode,
+        bool _casterPass, bool _dualParaboloid,
+        Ogre::SceneManager *_sceneManager) const override;
+
+    /// \brief Sends our custom data to GPU buffers that our
+    /// pieces activated in preparePassHash will need.
+    ///
+    /// Bytes written must not exceed what we informed in getPassBufferSize
+    /// \param _casterPass see base class
```
ahcorde, line 108: _shadowNode and _dualParaboloid

ahcorde (changes requested):

iche033 (commented):
Nice, it's working well for me. 

One comment / question about rendering solid color mode - this is more important for your follow-up PR when making this work for Ogre2MaterialSwitcher (used by selection buffer). There was logic in the original implementation to check for overlay material as it's needed for the gizmo visual, i.e. the RGB axes need to be rendered as overlay so that it can be clicked on by the mouse. Would the new setup work for this use case?

darksylinc:
> One comment / question about rendering solid color mode - this is more important for your follow-up PR when making this work for Ogre2MaterialSwitcher (used by selection buffer).

If you mean this bit of code:

At init:
```cpp
// plain overlay material
this->plainOverlayMaterial =
    this->plainMaterial->clone("plain_color_overlay");
if (!this->plainOverlayMaterial->getTechnique(0) ||
    !this->plainOverlayMaterial->getTechnique(0)->getPass(0))
{
  ignerr << "Problem creating selection buffer overlay material"
      << std::endl;
  return;
}
Ogre::Pass *overlayPass =
    this->plainOverlayMaterial->getTechnique(0)->getPass(0);
Ogre::HlmsMacroblock macroblock(*overlayPass->getMacroblock());
macroblock.mDepthCheck = false;
macroblock.mDepthWrite = false;
overlayPass->setMacroblock(macroblock);
```

At evaluation per frame:

```cpp
// check if it's an overlay material by assuming the
// depth check and depth write properties are off.
if (!datablock->getMacroblock()->mDepthWrite &&
    !datablock->getMacroblock()->mDepthCheck)
{
  subItem->setMaterial(this->plainOverlayMaterial);
}
```

It seems like the actual basically boils down to:

```cpp
if( hasTransparency( originalMaterial ) ) // Basically checks mDepthWrite = false & mDepthCheck = false
{
   useSameButTransparentMaterial(); // Basically sets mDepthWrite = false & mDepthCheck = false
}
```

If that's the case then it should work, because the Hlms customization will honour the original material's settings.

**However this does raise an important thing to point out:** This PR will not always produce the exact same output.

The reason is that the original would take a Renderable, apply either `plainMaterial` or `plainOverlayMaterial` and render the frame with it.

However this customization will honour the (possibly complex, possibly not) macro- and blendblock settings from the original material. In some cases this is desirable and the previous implementation would actually be considered buggy.

In other cases it may not be desirable. Probably alpha/additive blending is not desirable. A possible solution would be to clone the materials of those objects (and only those) that are trying to apply blending (with one that does not).

iche033:
yes that's the use case I'm referring to.

> However this customization will honour the (possibly complex, possibly not) macro- and blendblock settings from the original material. In some cases this is desirable and the previous implementation would actually be considered buggy.

hmm going back to the selection buffer use case, if we have a renderable with a semi-transparent material (or material with some alpha blendblock), would its color change if it's in front of another solid object? For selection buffer / mouse picking to work, we'll need to make sure it's rendered the same unique solid color that's assigned to it.

ogre2/src/Ogre2IgnHlmsPbsPrivate.cc
darksylinc: I fixed the other errors, but I'm not sure I understand what I need to do here?

The whole snippet is:

```cpp
#include "Ogre2IgnHlmsPbsPrivate.hh"

#include "Terra/Hlms/PbsListener/OgreHlmsPbsTerraShadows.h"

#include <ignition/common/Filesystem.hh>
#include <ignition/common/Util.hh>
```

Alphabetically:

```cpp
#include "Ogre2IgnHlmsPbsPrivate.hh"

#include <ignition/common/Filesystem.hh>
#include <ignition/common/Util.hh>
#include "Terra/Hlms/PbsListener/OgreHlmsPbsTerraShadows.h"
```

Is that what you're asking?

Cheers

darksylinc:
> hmm going back to the selection buffer use case, if we have a renderable with a semi-transparent material (or material with some alpha blendblock), would its color change if it's in front of another solid object? For selection buffer / mouse picking to work, we'll need to make sure it's rendered the same unique solid color that's assigned to it.

Yeah that was I referring to by "undesirable". Definitely we'll need to turn blending off.

iche033 (approved):
looks good to me!

darksylinc:
Before merging this, I want to include the fix to turn blending off. That's not much work and should definitely be included in this PR.

darksylinc:
OK I fixed the problem with the blendblocks.

It should be ready for merge now.

iche033:
latest changes look good to me.

darksylinc:
I'm not certain what 'alphabetize' means in this context (https://github.com/ignitionrobotics/ign-rendering/pull/517#discussion_r772178166), thus I don't know what/how to fix

ogre2/src/Ogre2IgnHlmsPbsPrivate.cc
iche033: hmm I think the order should be ok based on [google's style guide](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes) which we've been somewhat loosely following. The related headers are included first in alphabetical order then other libraries headers in alphabetical order. 

Pinging @ahcorde again to see what he has in mind.

darksylinc:
Btw can I ask to merge 6 -> main before merging this PR?

I can rebase the changes.

It's because these Hlms changes is tipping its toes with PRs #522 #523 & #524 which have already been merged but are not yet in main.

iche033:
> Btw can I ask to merge 6 -> main before merging this PR? (preferably ASAP)

created PR #532, will merge when CI turns green

iche033:
Going to merge this to unblock other follow-up work. We can address the remaining style issue later (if that's still considered an issue)

PR #494, opened 2021-11-24
Per-point colors in point clouds

chapulina:
# 🎉 New feature

* Part of https://github.com/ignitionrobotics/ign-gazebo/issues/1156
* Needed by https://github.com/ignitionrobotics/ign-gui/pull/318

## Summary

We have been populating the `colors` vector in `Ogre2DynamicRenderable`, but not using the colors for anything.

For most dynamic renderables, we fill the vertex buffer with position and normal. But points don't have normals, so we've been leaving half the buffer empty. The implementation on this PR takes advantage of those empty slots to place per-point colors (only RGB, no A).

Thanks @WilliamLewww for helping me out with this!

## Test it

The marker example on https://github.com/ignitionrobotics/ign-gui/pull/318 uses this:

pts_color

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [x] Updated documentation (as needed)
- [x] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [ ] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [x] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸🔸

PR #519, opened 2021-12-21
Fix color of last / extra marker points

chapulina:
# 🦟 Bug fix

* Follow up to #494 

## Summary

On #494 I had skipped filling the rest of the buffer for points because that was adding multiple points to the end of the cloud with the color blue. What I didn't notice is that I swapped the "last blue points" for "extra black point at the origin" :upside_down_face: 

* "last blue points" issue: the end of the vertex buffer is filled with points that have a Z normal (0,0,1 : blue) and coincide with the last point
* "extra black point at the origin": if the vertex buffer isn't filled completely with points, it's filled with zeroes (origin position, black color)

The  solution here is to first fill the buffer with blue normals, then override that with the color of the last point. I did this to match the implementation for other markers and reduce the amount of code changes, but some drawbacks to this approach are:

* We have multiple overlapping points in one place, which could be a bit wasteful depending on the size of the buffer. Ideally it would be good to shrink the buffer, but I don't know if that's possible.
* We're assigning the colors to the buffer twice, but we should be able to do it only once
* I think other markers may be suffering from the same problem, but ending up with Z normals instead of blue colors. That's because `GenerateNormals` is not using the whole `vertexBufferCapacity`

## Checklist
- [ ] Added tests - I don't know how to

### DIFF
--- a/ogre2/src/Ogre2DynamicRenderable.cc
+++ b/ogre2/src/Ogre2DynamicRenderable.cc
@@ -311,8 +311,7 @@ void Ogre2DynamicRenderable::UpdateBuffer()
 
   // fill the rest of the buffer with the position of the last vertex to avoid
   // the geometry connecting back to 0, 0, 0
-  if (vertexCount > 0 && vertexCount < this->dataPtr->vertexBufferCapacity &&
-      this->dataPtr->operationType != Ogre::OperationType::OT_POINT_LIST)
+  if (vertexCount > 0 && vertexCount < this->dataPtr->vertexBufferCapacity)
   {
     math::Vector3d lastVertex = this->dataPtr->vertices[vertexCount-1];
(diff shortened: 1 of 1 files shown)


iche033:
> We have multiple overlapping points in one place, which could be a bit wasteful depending on the size of the buffer. Ideally it would be good to shrink the buffer, but I don't know if that's possible.

There is some logic to update the vertex buffer size so that it can dynamically increase / shrink if needed, though it tries not do it often as it requires destroying and recreating the buffers and ogre resources. For use case like lidar visualization in which the number of points are always fixed, I think one idea to improve our logic is to provide an API that allow users to specify the exact initial vertex buffer size so we don't need to duplicate points at the end of the vertex buffer.

iche033 (approved):
changes look good to me.

PR #520, opened 2021-12-22
Support using custom shader materials and updating uniform variables (ogre2)

iche033:

Implemented code to supported loading custom shader material and updating their uniform variables.

Extended the `custom_shaders_uniforms` to run with ogre2 render engine. Note I made a slight tweak to the shader code in the demo to make the ogre 1.x and ogre 2.x behavior more consistent. When running with `ogre2`, the clip space z pos returned by `gl_Position.z` is different from `ogre` so it was rendering slightly different visual effect.

To test:

build and run the [custom_shaders_uniforms](https://github.com/ignitionrobotics/ign-rendering/tree/dc4e4c1f480e109bd9b15920300d2c9e49796f36/examples/custom_shaders_uniforms) demo:

```
./custom_shaders_uniforms ogre2
```

custom_shader_uniforms_ogre2

### DIFF
--- a/examples/custom_shaders_uniforms/Main.cc
+++ b/examples/custom_shaders_uniforms/Main.cc
@@ -68,7 +68,7 @@ void buildScene(ScenePtr _scene, const std::string &_engineName)
   light0->SetSpecularColor(0.5, 0.5, 0.5);
   root->AddChild(light0);
 
-  t std::string vertexShaderFile;
+  std::string vertexShaderFile;
   std::string fragmentShaderFile;
   if (_engineName == "ogre2")
   {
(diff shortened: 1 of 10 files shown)


examples/custom_shaders_uniforms/Main.cc
```diff
@@ -64,6 +68,19 @@ void buildScene(ScenePtr _scene)
   light0->SetSpecularColor(0.5, 0.5, 0.5);
   root->AddChild(light0);
 
+  t std::string vertexShaderFile;
```
WilliamLewww, line 71:
```suggestion
  std::string vertexShaderFile;
```

WilliamLewww (commented):
Just 1 small syntax error, but everything looks great!

examples/custom_shaders_uniforms/Main.cc
iche033: oops, fixed. 00d56a1

WilliamLewww (approved):

Issue #521, opened 2021-12-25
Potential bug: Wrong value stuck in Ogre2GpuRays


darksylinc:
## Environment
* OS Version: Ubuntu 18.04 LTS
* Source or binary build?
  main 5e190619fddd15560ef4ef94558ae5e7fd46316d but I think it affects all of them

## Description

I noticed `Ogre2LaserRetroMaterialSwitcher` performs the following:

```cpp
if (!subItem->hasCustomParameter(this->customParamIdx))
{
  // limit laser retro value to 2000 (as in gazebo)
  if (retroValue > 2000.0)
  {
    retroValue = 2000.0;
  }
  float color = retroValue / 2000.0;
  subItem->setCustomParameter(this->customParamIdx,
      Ogre::Vector4(color, color, color, 1.0));
}
```

Basically: "If we haven't set the retroValue yet, set it".

However I see two potential issues with this:

 - If the retro value can be changed at runtime (can it?) the change won't register unless Gazebo is recreating the associated Item
 - `customParamIdx = 10u;` **_which is also shared with the thermal camera (same value)._** As a result, if an object is both in the Thermal Camera **and** in GpuRays, then the Thermal Camera will override the retroValue (because ThermalCamera never checks `hasCustomParameter( ... )`) and later `GpuRays` will not try to override anything.

**I see two possible solutions:**

 1. Change `customParamIdx = 10u;` to one of the camera implementations (not recommended, this problem will reappear)
 2. Get rid of the `if (!subItem->hasCustomParameter(this->customParamIdx))` check. It's no performance improvement, and this fix will get rid of the problem and any future clash.

## Steps to reproduce

This was found through manual examination, but I suppose it should be possible to repro by using GpuRays and Thermal Camera at the same time on the same object

## Other

It's possible both Ogre2 and Ogre1 backends are affected. I think the fix should be trivial to propagate to all branches without breaking ABI.

darksylinc:
While searching on what `laserRetro` does, it appears [someone stumbled on this bug](https://answers.gazebosim.org//question/24933/no-impact-of-laser_retro-when-using-ray-sensor/).

darksylinc:
This bug is fixed in Fortress & Garden on Ogre2.
It should be backported to Citadel.

I don't know if the bug exists in Ogre1.

PR #522, opened 2021-12-26
Fix various issues with Ogre2GpuRays

darksylinc:
There were multiple bugs that had to be fixed:

- Objects which had no laser_retro set (or an invalid one) would be
rendered using regular lighting PBS shaders; causing the retro value to
contain garbage instead of the intended default value of 0
- Performance regression in 993bb0bff6792627fe1f0cc19108d7280f369da1:
When ported to Ogre 2.2, ignition would render the scene 4 times instead
of 2 (twice for objects, twice for particles). Depth buffer data was
created through a 2nd redundant pass. This is inefficient as the depth
buffer data is already available from the 1st pass
- Bug regression in 993bb0bff6792627fe1f0cc19108d7280f369da1: In the
pass for normal objects, ignition would first use IGN_VISIBILITY_ALL
flags for colour, but it would use 0x01000000 for depth. This creates
the wrong mix of data because an occluder that had visibility set to
0x01000000 would use the correct depth but the retro value of the object
behind it. This bug is implicitly fixed by the previous item (colour and
depth data are now obtained from the same pass instead of using multi
pass)
- PFG_R8_UNORM is nowhere enough to contain all possible laser retro
values and would quantize too much. Increased precision to PFG_R16_UNORM
- Affects #521. The problem is fixed for Fortress by always overriding
setCustomParameter, but likely the fix should be backported to previous
versions.

# 🦟 Bug fix

Affects #521 
Fixes other bugs without ticket

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -238,6 +238,9 @@ void Ogre2LaserRetroMaterialSwitcher::cameraPreRenderScene(
     Ogre::Item *item = static_cast<Ogre::Item *>(object);
 
     std::string laserRetroKey = "laser_retro";
+
+    float retroValue = 0.0f;
+
     // get visual
     Ogre::Any userAny = item->getUserObjectBindings().getUserAny();
     if (!userAny.isEmpty() && userAny.getType() == typeid(unsigned int))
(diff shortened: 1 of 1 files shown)


iche033 (approved):
ah now this refreshes my memory of the retro + visibility mask issue.

I think with these changes, we should be able to remove this todo comment about using an even more complicated workaround to fix retro values for objects that do not have one set. 
https://github.com/ignitionrobotics/ign-rendering/blob/fa4b52cbe513282b7c2921e28d631d601a6fc318/ogre2/src/Ogre2GpuRays.cc#L910-L919

This is a much better fix.

I'll create a follow-up PR to remove this todo item and add a small check in the integration test for zero retro values.

osrf-triage:

PR #524, opened 2021-12-26
Performance optimization by avoiding unnecessary passes

darksylinc:
# 🦟 Bug fix

No ticket created

## Summary

Ignition is pointlessly performing two passes when one is enough, as the
depth buffer data is already filled.

In the case of `Ogre2DepthCamera` we cannot do the same with
colorTexture & depthTexture though; because the way they are filled is different (colorTexture contains particles
and potentially other stuff that might write to depth; depthTexture doesn't)

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2DepthCamera.cc
+++ b/ogre2/src/Ogre2DepthCamera.cc
@@ -670,11 +670,6 @@ void Ogre2DepthCamera::CreateDepthTexture()
     particleTexDef->preferDepthTexture = false;
     particleTexDef->fsaa = "0";
 
-    Ogre::RenderTargetViewDef *rtvParticleTexture =
-      baseNodeDef->addRenderTextureView("particleTexture");
-    rtvParticleTexture->setForTextureDefinition(
-      "particleTexture", particleTexDef);
-
     Ogre::TextureDefinitionBase::TextureDefinition *particleDepthTexDef =
         baseNodeDef->addTextureDefinition("particleDepthTexture");
     particleDepthTexDef->textureType = Ogre::TextureTypes::Type2D;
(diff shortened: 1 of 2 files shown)


ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -690,12 +681,15 @@ void Ogre2DepthCamera::CreateDepthTexture()
     particleDepthTexDef->fsaa = "0";
     particleDepthTexDef->textureFlags &= ~Ogre::TextureFlags::Uav;
 
-    Ogre::RenderTargetViewDef *rtvparticleDepthTex =
-      baseNodeDef->addRenderTextureView("particleDepthTexture");
-    rtvparticleDepthTex->setForTextureDefinition(
-      "particleDepthTexture", particleDepthTexDef);
+    // Auto setup the RTV then manually override the depth buffer so
+    // it uses the one we created (and thus we can sample from it later)
```
iche033, line 685: do we need to change the `depthBufferId` to something other than `Ogre::DepthBuffer::POOL_NON_SHAREABLE;`?

iche033 (approved):
nice, works for me. Verified that the particles in depth camera look good and also mouse camera control is also working as expected. Just have one minor question.

ogre2/src/Ogre2DepthCamera.cc
darksylinc: Hi!

This change from 2.1 to 2.2+

Now depthBufferId should be getting ignored when the texture is `PFG_D32_FLOAT` (or any other depth format) and also ignored when the RTV is specifically linked to a depth buffer.

IIRC `POOL_NON_SHAREABLE` should be avoided by user code now. It is still used internally, but makes no sense in user code.

In 2.1 when you created a depth texture, you were not actually creating a depth texture. You set the depth pool ID to access the contents of an already created depth buffer associated with the render target with the same parameters and pool ID.

With POOL_NON_SHAREABLE you would prevent its contents from later being overridden by a later pass.

In 2.2 when you create a depth texture, you're actually creating a unique depth texture. And use RTVs to associate that depth texture with a colour target for rendering in a specific pass (and thus now depth pools are ignored).

The pool ID system only remains for throwable depth buffers for which you don't want to access/sample its depth contents and are automatically created and reused.

iche033: got it, thanks for the explanation!

osrf-triage:

PR #558, opened 2022-02-08
Make sure shader param exists before setting its value

iche033:

# 🦟 Bug fix
## Summary

Fixes crash when setting a shader parameter that does not exist. Now it prints out a warning msg to let users know what param will not be set.

## Checklist
- [x] Added tests

### DIFF
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -646,6 +646,13 @@ void Ogre2Material::UpdateShaderParams(ConstShaderParamsPtr _params,
       continue;
     }
 
+    if (!_ogreParams->_findNamedConstantDefinition(name_param.first))
+    {
+      ignwarn << "Unable to find GPU program parameter: "
+              << name_param.first << std::endl;
+      continue;
+    }
+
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

osrf-triage:

srmainwaring:
@iche033 I've realised that this breaks the custom shaders for Metal, which can be seen on the waves example

waves-textures-issue

The issue is that the Metal shaders do not declare the textures as parameters and they are not found as named constants. As a result a warning is issued and registration of the texture unit state is skipped. For textures the named constant is only set for OpenGL:

https://github.com/ignitionrobotics/ign-rendering/blob/5e67f4159761151b09560d447c73d3123e954963/ogre2/src/Ogre2Material.cc#L786-L791

A possible fix would be to skip this check if the shader param type is a texture and not OpenGL?

iche033:
Thanks for the patch! yeah that looks good to me. I don't know if vulkan will have the same issue or not (being added in https://github.com/ignitionrobotics/ign-rendering/pull/553). I guess we'll find out. I created a PR for the patch: https://github.com/ignitionrobotics/ign-rendering/pull/574

PR #563, opened 2022-02-18
Fixed segfault caused by OgreRTShaderSystem

WilliamLewww:
# 🦟 Fixed segfault caused by OgreRTShaderSystem

Fixes Segfault in 
* https://github.com/ignitionrobotics/ign-gazebo/pull/1215

## Summary
Fixes the segfault that occurs after running any ign-gazebo test that uses OGRE1.

The removal of the `this->Fini()` is safe because OgreRTShaderSystem's `Fini()` method is already being ran in OgreRenderEngine:
https://github.com/ignitionrobotics/ign-rendering/blob/e0ed1b1732e654dab8084fc7821df05b4da7c759/ogre/src/OgreRenderEngine.cc#L108

## Checklist
- [ ] Added tests


iche033:
I'm testing this branch with the integration test in https://github.com/ignitionrobotics/ign-gazebo/pull/1215. This fixes the segfault for me. However I think the `Fini` function is also not being called by `OgreRenderEngine` in the test because when I tried adding a print statement, I don't see it printed.

This leads me to think that it's a threading issue. The OgreRTShaderSystem is created in the rendering thread but destroyed when it goes out of scope in the test main thread. For now, we can do something like the [Ogre2RayQuery](https://github.com/ignitionrobotics/ign-rendering/blob/ign-rendering6/ogre2/src/Ogre2RayQuery.cc#L103) class and add a check in the `OgreRTShaderSystem` destructor to make sure that this function is only called in the thread it's created in. Something like:

```
if (std::this_thread::get_id() == this->dataPtr->threadId)
  this->Fini();
```

A more involved fix is to make sure we destroy the render engine in ign-gazebo's sensors system properly in the [RenderThread](https://github.com/ignitionrobotics/ign-gazebo/blob/ign-gazebo6/src/systems/sensors/Sensors.cc#L301) but that's needs a bit more work and testing.

iche033 (approved):
looks good to me

Issue #543, opened 2022-01-22
Unit test coverage: low level materials


darksylinc:
I noticed [Fortress or Garden is gaining the ability](https://github.com/ignitionrobotics/ign-rendering/pull/537/files) to use custom shaders in fact [PR 541](https://github.com/ignitionrobotics/ign-rendering/pull/541) is a reflection of this.

There were adjustments to `Ogre2MaterialSwitcher` [to account this](https://github.com/ignitionrobotics/ign-rendering/pull/537/commits/73b9fa135883aa287290a32c171326ec7ad56a11#diff-8201a5d2b22efadecb321b97561fe1a203c6be22594cd7f8fe9f06fd3abb42ceL110) and this is covered by INTEGRATION_camera however the following switchers aren't covered:

1. `Ogre2LaserRetroMaterialSwitcher`
2. `Ogre2SegmentationMaterialSwitcher`
3. `Ogre2ThermalCameraMaterialSwitcher`

These switchers aren't currently handling objects with low level materials and there are no unit tests covering objects seen by those cameras.

## Desired behavior

Unit tests should cover all 4 material material switchers when low level materials are present

darksylinc:
I'm closing this because it seems #579 added coverage for most of these issues.

While 100% coverage of these material switchers hasn't been achieved; it would seem that such problem should be tracked by a different ticket tracking coverage of each specific switcher.

PR #578, opened 2022-03-06
Respect vertex shader from low level materials in sensor passes (#544)

### LINKED ISSUE #544: Incorrect sensor data: Low level materials and other cameras
This is a similar / sister ticket of #543:

After gaining the ability of custom low level materials on objects, it raises an issue: If the material implements custom vertex deformation (i.e. like the [waves sample](https://github.com/ignitionrobotics/ign-rendering/pull/541/files#diff-7b4359635ac188a7a76bf79a0b65cd871f9276635dd2acebafd2842d27d81ba4) from PR #541) then this deformation won't happen in the other cameras (Thermal, LaserRetro, Segmentation, SelectionBuffer).

The solutions I see are (some of them are not mutually exclusive):

1. Attempt to clone the material while overriding the pixel shader while keeping the vertex shader (should work 95% of time, as a fallback)
2. Let the interface define alternate materials for each of the 4 cameras
3. Patch the pixel shader ourselves. Ask the user to write a special string pattern e.g.
```cpp
#version 330

// IGN CUSTOM VARIABLES

void main()
{
   gl_FragColor = ...;
// IGN CUSTOM POST PS EXECUTION
}
```
and then when the switchers run, we look at the shader source for string matching `// IGN CUSTOM POST PS EXECUTION` and `// IGN CUSTOM VARIABLES` and we replace those strings with our custom strings, clone the pixel shader and compile that.

## Update

So, talking with @iche033 , solution 1 is likely the simplest and most viable.

I also remembered what was that "5% incompatibility" I mentioned. Shaders have signatures. What the VS sends to the PS must be compatible; and declaration order is important.

For example if Vertex Shader declares:

```hlsl
struct ToPixelShader
{
    float2 uv;
    float3 otherData;
}
```

Then our pixel shader can use:

```hlsl
struct ToPixelShader
{
    float2 uv;
}
```

However if the VS had declared:

```hlsl
struct ToPixelShader
{
    float3 otherData;
    float2 uv;
}
```

In other words, the declaration order has been reversed, then linkage will fail because the first float3 does not match with the float2 that the pixel shader will be expecting.

As I told Ian, it would be wise to make a "tester tool"; i.e. a program or routine that will attempt to run that low level material through all the cameras. If it passes, the materials is good. If it fails, the user can check the logs to see what's causing trouble (which is almost always going to be an incompatible VS to PS signature).

When it comes to HLSL (ie. Windows / D3D11) position is part of the signature, thus to maximize compatibility the signature should always look like this:

```hlsl
struct ToPixelShader
{
    float2 uv : TEXCOORD0;
    float3 position : SV_Position;

    // The rest
    float3 otherData : TEXCOORD1;
}
```

## Desired behavior

The 4 cameras should consider vertex deformation and pixel shader's discard functionality.

darksylinc:
# 🦟 Bug fix

Fixes #544 

## Summary

It implements the first option (clone the material with same VS but different PS).

However I could NOT test this PR because I had no test nor sample to test it.

That is the reason this PR is marked as draft. I would appreciate a way to play with this (i.e. a low level material entity being drawn in all sensors).

The lack of tests is tracked by #543

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -309,11 +309,27 @@ void Ogre2LaserRetroMaterialSwitcher::passPreExecute(
 
       if (!subItem->getMaterial().isNull())
       {
-        // TODO(anyone): We need to keep the material's vertex shader
-        // to keep vertex deformation consistent. See
-        // https://github.com/ignitionrobotics/ign-rendering/issues/544
         this->materialMap.push_back({ subItem, subItem->getMaterial() });
-        subItem->setDatablock(defaultPbs);
+
+        // We need to keep the material's vertex shader
+        // to keep vertex deformation consistent; so we use
+        // a cloned material with a different pixel shader
+        // https://github.com/ignitionrobotics/ign-rendering/issues/544
+        auto material = Ogre::MaterialManager::getSingleton().getByName(
+          subItem->getMaterial()->getName() + "_solid",
+          Ogre::ResourceGroupManager::DEFAULT_RESOURCE_GROUP_NAME);
+        if (material->getNumSupportedTechniques() > 0u)
+        {
+          subItem->setMaterial(material);
+        }
+        else
+        {
+          // The supplied vertex shader could not pair with the
+          // pixel shader we provide. Try to salvage the situation
+          // using PBS shader. Custom deformation won't work but
+          // if we're lucky that won't matter
+          subItem->setDatablock(defaultPbs);
+        }
       }
       else
       {
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -68,6 +68,22 @@ class ignition::rendering::Ogre2MaterialPrivate
   /// \brief Parameters to be bound to the fragment shader
   public: ShaderParamsPtr fragmentShaderParams;
 
+  /// \brief Material to be used when rendering to special
+  /// cameras (e.g. sensors) like Ogre2GpuRays,
+  /// Ogre2LaserRetroMaterialSwitcher, etc
+  ///
+  /// It shares the same Vertex Shader, but uses a different
+  /// Pixel Shader
+  public: Ogre::MaterialPtr ogreSolidColorMat;
+
+  /// \brief A clone of plaincolor_fs. We need a clone
+  /// because some Metal needs to pair with the vertex shader they're
+  /// going to be used and Vulkan needs to have the same or a compatible
+  /// Root Layout profile.
+  ///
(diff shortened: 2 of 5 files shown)


iche033:
I am testing this with this [integration test](https://github.com/ignitionrobotics/ign-rendering/blob/b999d0a2a66455c52793e39faa007b4d9b783725/test/integration/camera.cc) (`ShaderSelection`). Just copy and  paste to `test/integration/camera.cc`, build and run it: `./<install_path>/bin/INTEGRATION_camera`

I'm getting a segfault pointing to this [line](https://github.com/ignitionrobotics/ign-rendering/blob/6d96fa25637349081cb8f306e274504acdbbade3/ogre2/src/Ogre2SegmentationMaterialSwitcher.cc#L311) in `Ogre2SegmentationMaterialSwitcher.cc`.  Looks like it's not able to find the material file with `_solid` name suffix. Interestingly if I comment out the the thermal and gpu lidar cameras, segmentation camera works by itself. I confirmed this with ign-gazebo running this [world file](https://gist.github.com/iche033/01ef166ed150d0b0046daa8176de5640). If I only have the segmentation camera in the world I see that the sphere deforms in the camera view (yay!) but if I add a thermal camera or gpu lidar, it crashes with similar backtrace.

darksylinc:
Thanks for the test! I accidentally removed two lines I shouldn't have, thus Thermal camera wasn't restoring the original material once it was done.

The code failed because the last camera was trying to look for "material_name_solid_solid", since Thermal didn't restore ""material_name_solid" ->"material_name"

darksylinc:
It works!!

Screenshot_2022-03-12_13-10-39
!

darksylinc:
The code definitely handles the reflection hint, but perhaps it doesn't run or runs too late?

srmainwaring:
> The code definitely handles the reflection hint, but perhaps it doesn't run or runs too late?

It may be running too late. The error is emitted here:

https://github.com/ignitionrobotics/ign-rendering/blob/23955a98b0b5082b6f6a7ea1ed3a3e84e1c39cfd/ogre2/src/Ogre2Material.cc#L869-L870

before the reflection_pair_hint is set. I'll see if I can either move the reflection hint earlier or defer setting the constant `inColor`.

darksylinc:
Ahhh!! Of course!

That call can only be done in `Ogre2Material::SetVertexShader`, when shader_reflection_pair_hint is set for ogreSolidColorShader

srmainwaring:
Yep - just verified for the mesh_viewer example. Moved that block to `Ogre2Material::SetVertexShader` and seems ok. I'll check the other examples and the main gazebo calls now.

PR #618, opened 2022-05-05
ign -> gz Header Migration : ign-rendering

methylDragon:
See https://github.com/ignition-tooling/release-tools/issues/711


include/ignition/rendering/detail/Export.hh
```diff
@@ -0,0 +1,18 @@
+/*
```
chapulina, line 1: This file is not needed

ogre/include/ignition/rendering/ogre/detail/Export.hh
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+#include <gz/rendering/ogre/Export.hh>
```
chapulina, line 18: This is missing `detail` 🧐 

But in any case we shouldn't need any `ignition/**/detail` files

ogre2/src/terrain/Terra/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ file( GLOB_RECURSE TERRA_SOURCES
 
 add_library(${PROJECT_NAME} STATIC ${TERRA_SOURCES})
 
-if(IGN_ADD_fPIC_TO_LIBRARIES AND NOT _ign_add_library_INTERFACE)
+if(IGN_ADD_fPIC_TO_LIBRARIES AND NOT _gz_add_library_INTERFACE)
```
chapulina, line 11: Humm using CMake variables that were meant to be private 😕 

Good that this is migrated here though 🙂 Upstream already changed in 

* https://github.com/gazebosim/gz-cmake/pull/245

chapulina (approved):
🦾 

image

ogre2/src/terrain/Terra/CMakeLists.txt
methylDragon: Yepp

ogre/include/ignition/rendering/ogre/detail/Export.hh
methylDragon: Now no longer created in https://github.com/gazebosim/gz-rendering/pull/618/commits/a4191f722debd816f291558460a23751699cda8e

include/ignition/rendering/detail/Export.hh
methylDragon: Now no longer created in https://github.com/gazebosim/gz-rendering/pull/618/commits/a4191f722debd816f291558460a23751699cda8e

chapulina (approved):
🚢 it!

PR #635, opened 2022-05-26
Change aspect to AspectRatio()

AzulRadio:
# 🦟 Bug fix

Fixes ignition crash bug introduced in [gz-sim#1499](https://github.com/gazebosim/gz-sim/pull/1499)

## Summary

gz-sim#1499 adds a feature to change the FOV of the user camera at runtime. But if we scale the window after changing the FOV, Gazebo has a chance to crash. (Always crash on Ogre, sometimes crash on Ogre2)

Changing `aspect` to `AspectRatio()` in `OgreCamera::SetHFOV()` solves the bug. 

Also add comment to warn users not to use `aspect` in `BaseCamera` but to use `OrgeCamera::AspectRatio()` instead.

#### Why this bug happened: 
[Resizing the window](https://github.com/gazebosim/gz-gui/blob/5c096882eb2e7a3bd120983a627cf80da66064b6/src/plugins/minimal_scene/MinimalScene.cc#L292) needs the correct aspect ratio value. There are 2 aspect ratio values available: one in Ogre library (`AspectRatio()`), and the other is in `BaseCamera` (`this->aspect`). 

When we change the FOV and resize the window, the aspect ratio and fov value in the Ogre library will both change but `aspect` in the `BaseCamera` only update its fov value but not aspect ratio. However, in `Ogre2Camera::SetHFOV()`, it is still using the deprecated aspect ratio value from the `BaseCamera`. 

If you check with gdb, the value read by `AspectRatio()` (directly from Ogre) and `aspect` from `BaseCamera` no longer agree with each other after changing FOV and resizing the window. That is why there is a crash.

## Checklist
- [ ] Added tests

### DIFF
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -217,6 +217,7 @@ namespace ignition
       protected: double farClip = 1000.0;
 
       /// \brief Aspect ratio
+      /// Do not use this variable. Use ApsectRatio() instead.
       protected: double aspect = 1.3333333;
 
       /// \brief Horizontal camera field of view
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -286,7 +286,8 @@ namespace ignition
     void BaseCamera<T>::SetImageWidth(const unsigned int _width)
     {
       this->RenderTarget()->SetWidth(_width);
-      this->SetAspectRatio(1.0 * _width / this->ImageHeight());
+      this->SetAspectRatio(
+        static_cast<double>(_width) / static_cast<double>(this->ImageHeight()));
     }
 
     //////////////////////////////////////////////////
@@ -301,7 +302,8 @@ namespace ignition
     void BaseCamera<T>::SetImageHeight(const unsigned int _height)
     {
       this->RenderTarget()->SetHeight(_height);
-      this->SetAspectRatio(1.0 * this->ImageWidth() / _height);
+      this->SetAspectRatio(
+        static_cast<double>(this->ImageWidth()) / static_cast<double>(_height));
     }
 
     //////////////////////////////////////////////////
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -286,7 +286,7 @@ namespace ignition
     void BaseCamera<T>::SetImageWidth(const unsigned int _width)
     {
       this->RenderTarget()->SetWidth(_width);
-      this->aspect = this->AspectRatio();
+      this->SetAspectRatio(1.0 * _width / this->ImageHeight());
     }
 
     //////////////////////////////////////////////////
@@ -301,7 +301,7 @@ namespace ignition
     void BaseCamera<T>::SetImageHeight(const unsigned int _height)
     {
       this->RenderTarget()->SetHeight(_height);
-      this->aspect = this->AspectRatio();
+      this->SetAspectRatio(1.0 * this->ImageWidth() / _height);
     }
 
     //////////////////////////////////////////////////
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -217,7 +217,6 @@ namespace ignition
       protected: double farClip = 1000.0;
 
       /// \brief Aspect ratio
-      /// Do not use this variable. Use ApsectRatio() instead.
       protected: double aspect = 1.3333333;
 
       /// \brief Horizontal camera field of view
--- a/include/ignition/rendering/base/BaseCamera.hh
+++ b/include/ignition/rendering/base/BaseCamera.hh
@@ -286,6 +286,7 @@ namespace ignition
     void BaseCamera<T>::SetImageWidth(const unsigned int _width)
     {
       this->RenderTarget()->SetWidth(_width);
+      this->aspect = this->AspectRatio();
     }
 
     //////////////////////////////////////////////////
@@ -300,6 +301,7 @@ namespace ignition
     void BaseCamera<T>::SetImageHeight(const unsigned int _height)
     {
       this->RenderTarget()->SetHeight(_height);
+      this->aspect = this->AspectRatio();
     }
 
     //////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


ahcorde (dismissed):
linters https://github.com/gazebosim/gz-rendering/runs/6615769097?check_suite_focus=true

jennuine (commented):
Other than linters this LGTM, will need @iche033 's stamp of approval

iche033:
Ideally we also update the value of `this->aspect` after setting FOV and/or image size, potentially in BaseCamera's `SetImageWidth` / `SetImageHeight` functions so that it's not outdated. Can you also add a quick test in [Camera_TEST.cc](https://github.com/gazebosim/gz-rendering/blob/ign-rendering6/src/Camera_TEST.cc)?

AzulRadio:
@iche033 Just to clarify, do you mean we should update `this->aspect` in `BaseCamera::SetImageWidth`, `BaseCamera::SetImageHeight`, and `BaseCamera::SetHFOV`, and add a test to check if `aspect` agrees with `AspectRatio`?

iche033:
> `BaseCamera::SetImageWidth`, `BaseCamera::SetImageHeight`

yes I would expect the aspect ratio to change when user calls these 2 functions. I think the HFOV should not affect aspect ratio though but we should double check that.

include/ignition/rendering/base/BaseCamera.hh
```diff
@@ -217,6 +217,7 @@ namespace ignition
       protected: double farClip = 1000.0;
 
       /// \brief Aspect ratio
+      /// Do not use this variable. Use ApsectRatio() instead.
```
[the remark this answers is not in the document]
AzulRadio: fixed in 2286b60

AzulRadio:
@iche033 Aspect ratio change added in fbbbc4b. I'm not sure how to add the test to `Camera_TEST` because `aspect` is a protected variable (not accessible in Camera_TEST). I verified with GDB and it's now changing when resizing the window.

ogre2/src/Ogre2Camera.cc
```diff
@@ -76,7 +76,7 @@ void Ogre2Camera::SetHFOV(const math::Angle &_angle)
 {
   BaseCamera::SetHFOV(_angle);
   double angle = _angle.Radian();
-  double vfov = 2.0 * atan(tan(angle / 2.0) / this->aspect);
+  double vfov = 2.0 * atan(tan(angle / 2.0) / this->AspectRatio());
```
jennuine, line 79: If `this->aspect = this->AspectRatio();` is required in `OgreCamera::SetHFOV`, then we should add it to `Ogre2` camera's as well

jennuine (commented):
For a `Camera_TEST` you can check that after `SetImageWidth` `AspectRatio()` agrees.

e.g., 
```c++
double width = 100
camera.SetImageWidth(width)
double aspectRatio = width / camera.ImageHeight()
EXPECT_NEAR(aspectRatio, camera.AspectRatio(), 1e-6)
```

Similarly for `SetImageHeight`

src/Camera_TEST.cc
```diff
@@ -178,6 +178,11 @@ void CameraTest::RenderTexture(const std::string &_renderEngine)
   camera->SetImageHeight(80u);
   EXPECT_EQ(80u, camera->ImageHeight());
 
+  double width = 100
+  camera.SetImageWidth(width)
+  double aspectRatio = width / camera.ImageHeight()
+  EXPECT_NEAR(aspectRatio, camera.AspectRatio(), 1e-6)
```
jennuine, line 184: This is missing `;` at the end of each line and `camera` is a pointer. This test fails though, you will need to try to investigate why and how to correct this

Since you are using `ogre` on your system, update `ogre2` to `ogre` here: https://github.com/gazebosim/gz-rendering/blob/a3dc1aeb8102c1e96616dfee06310c226c1b995c/test/test_config.h.in#L16

To compile the test, run  `colcon build --cmake-args -DBUILD_TESTING=ON --merge-install --packages-select ignition-rendering6` in the root of your workspace. To run the test you can do `./build/ignition-rendering6/bin/UNIT_Camera_Test`

jennuine: You should also add a similar test with `SetImageHeight`

include/ignition/rendering/base/BaseCamera.hh
```diff
@@ -286,6 +286,7 @@ namespace ignition
     void BaseCamera<T>::SetImageWidth(const unsigned int _width)
     {
       this->RenderTarget()->SetWidth(_width);
+      this->SetAspectRatio(1.0 * _width / this->ImageHeight());
```
jennuine, line 289: I assume you are multiplying by `1.0` to cast the result to a `double`. It's better to use `static_cast<double>` to be explicit with the cast making the purpose clear. Same for `SetImageHeight`
```suggestion
      this->SetAspectRatio(
        static_cast<double>(_width) / static_cast<double>(this->ImageHeight()));
```

src/Camera_TEST.cc
```diff
@@ -178,6 +178,17 @@ void CameraTest::RenderTexture(const std::string &_renderEngine)
   camera->SetImageHeight(80u);
   EXPECT_EQ(80u, camera->ImageHeight());
 
+  double height = 80;
+  camera->SetImageHeight(height);
+  double aspectRatio = camera->ImageWidth() / height;
+  EXPECT_NEAR(aspectRatio, camera->AspectRatio(), 1e-6);
```
jennuine, line 184: Since your new tests are almost the same as the previous test, lets remove the new test and continue on the previous one. E.g.,
```suggestion
  unsigned int height = 80;
  camera->SetImageHeight(height);
  EXPECT_EQ(height, camera->ImageHeight());
  double aspectRatio =
    static_cast<double>(camera->ImageWidth()) / static_cast<double>(height));
  EXPECT_NEAR(aspectRatio, camera->AspectRatio(), 1e-6);
```

The same should be done for width.

Also, go ahead and update this expectation below to use the new variables: https://github.com/gazebosim/gz-rendering/blob/8d4e5425f86b6ff4885eb8362fc74c1ee21d993d/src/Camera_TEST.cc#L195

e.g., `EXPECT_EQ(width*height*3u, ...)`

jennuine (commented):
Nice work! Left some minor comments

jennuine (commented):
LGTM, @iche033 do you mind to take another look?

iche033 (approved):
looks good to me.

PR #639, opened 2022-06-04
ign -> gz Environment Variable Migration : gz-rendering

methylDragon:
See: https://github.com/gazebo-tooling/release-tools/issues/734

Actually no environment variables need to be migrated (though how it's been implemented means there isn't really a way to emit warnings...)

They were implemented like so in the namespace migration:
```
  const char *env = std::getenv("GZ_RENDERING_RESOURCE_PATH");
  env = (env) ? env : std::getenv("IGN_RENDERING_RESOURCE_PATH");
```

### DIFF
--- a/Migration.md
+++ b/Migration.md
@@ -15,8 +15,8 @@ release will remove the deprecated code.
 2. The `ignition` namespace is deprecated and will be removed in future versions.  Use `gz` instead.
 3. Header files under `ignition/...` are deprecated and will be removed in future versions.
    Use `gz/...` instead.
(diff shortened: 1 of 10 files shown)


chapulina (changes requested):
I just noticed that we're not printing warnings when the user uses the `IGN_` vars. Would you mind adding warnings as part of this PR?

methylDragon:
> I just noticed that we're not printing warnings when the user uses the `IGN_` vars. Would you mind adding warnings as part of this PR?

https://github.com/gazebosim/gz-rendering/pull/639/commits/5192651901a098b0ba6278a01660141f5e30ec9b !

chapulina (approved):
Thanks! LGTM

PR #642, opened 2022-06-06
Support for textures as binary blobs

luca-della-vedova:
# 🎉 New feature

Edit: Ready for review!

## Summary
This PR is an early request-for-feedback on having support for in-memory textures. It reached a "working" state with the gz-common [luca/inmemory_textures branch](https://github.com/gazebosim/gz-common/tree/luca/inmemory_textures) where GLB assets with textures stored as binary blobs (array of bytes containing the file) are correctly parsed and rendered. I.e. below a world with a nurse looking at an ambulance stretcher, both defined as single file GLBs with embedded textures:

image

Before I refine too much the implementation I'd like to know whether the API makes sense or I should define a different one, specifically:

Right now, to set a texture we currently use the [SetTexture](https://github.com/gazebosim/gz-rendering/blob/main/ogre2/include/gz/rendering/ogre2/Ogre2Material.hh#L111) method which takes as an input a filename, however we would need to support arbitrary binary blobs in memory for formats with embedded textures to work.

I have a few ideas and currently implemented one but happy to scrap it completely:

**Parse images into a gz::common::Image and change SetTexture (or add an overloaded method) to accept a `gz::common::Image` as an input.**

This would have worked and sounded the cleanest but I was concerned about doing additional processing that might not be needed. For example, we could parse an image and change it into RGBA8 but it is unclear what format might be demanded by downstream users of gz-rendering. For example OGRE2 can parse PNGs straightaway hence if we preprocessed the image we would have unnecessarily inflated the data before copying it around. Also `gz::Image` doesn't expose raw data yet so we would have needed to add an API to set (from compressed) and get raw data, which is what we would have ended up using anyway.

**Change SetTexture to accept a vector of raw data parameter, set to default empty**

 I.e. going from:
 
```
void SetTexture(const std::string &_texture);
```
To
```
void SetTexture(const std::string &_texture, std::vector<unsigned char> _data = {});
```

In this case `_texture` would either be a URI or a unique name that can be used to identify the binary texture (to be generated to make sure that if we load multiple instances of the same model through binary blobs we don't duplicate the texture).
This feels generally a lot simpler but the ambiguity of the `_texture` argument (URI to a file or unique name) for the two use cases felt a bit unclear.

This approach would probably reduce the diff / complexity quite a bit since we can leverage all the existing code and just add a check to the `SetTexture` function and load the texture from memory instead of a file if the `_buf` argument is not empty.

**Brand new overloaded function**

This is the approach I went for when prototyping for simplicity but I'm happy to scrap, it currently looks like this:

```
void SetTexture(const std::vector<unsigned char> &_buf, const std::string& _name)
```

The tricky part here is that we have to save the texture buffer and name to an internal member to make sure the data is retained when the class is copied around (which seems to happen quite a bit!)
Now it gets ambiguous, if a user calls the normal `SetTexture` and tells the rendering engine "load this file", then calls the overloaded method and tells the engine "load this texture from a binary file" what should the appropriate behavior be? I'm guessing clearing the previous texture, which is fine but again the API is slightly unclear, if you call a `SetTexture(URI)`, then somewhere else a `SetTexture(binary data)`, then next time you call the getter `Texture()` to get the URI the information is gone, or it's not gone but it's not rendered, just a bit tricky.
I found this especially tricky to figure out in the custom [copy constructor](https://github.com/gazebosim/gz-rendering/blob/luca/inmemory_textures/include/gz/rendering/base/BaseMaterial.hh#L982-L984) since its result will now depend on the order of operations.

### Other considerations

* std::vector<> vs pointers: I am currently using `std::vector<>` for the data but noticed while debugging that there are quite a few copies / clones happening behind the scenes and the data gets copied around at least twice (once in the gz-common -> gz-rendering transition and once through a [Clone()](https://github.com/gazebosim/gz-rendering/blob/main/include/gz/rendering/base/BaseMaterial.hh#L929) call). I feel this is a non trivial amount for data that is multiple megabytes per object so happy to move to a `shared_ptr` approach if that would help with performance.
* Passing vs not passing image formats: The asset parsing library has knowledge of what format the data is in (i.e. png, jpg). Right now the image is literally a binary file that includes the signature that can be used to decode the format hence it "just works" with OGRE2. We could add a parameter to pass the format specifically but I couldn't find a valid case for it yet.
* We would need a similar set of `SetTexture` overloads for all the PBR maps like metalness, roughness, normal, and probably lightmaps, hence it would be good to know the best API to make the diff as simple / manageable as possible

@iche033 @chapulina for API feedback!

## Test it
Stay tuned!

## Checklist
- [x] Added tests
- [x] Added example and/or tutorial


luca-della-vedova:
OK sounds good! Then I'll change the approach to use shared memory and the API to be:

```
SetTexture(std::string& _name, std::shared_ptr<rendering::Image> _image = nullptr);
```

Then current calls will go through the usual code path of using the texture as a URI and loading it from a file, if the image data is set it will be used as a unique texture name.
My instinct would be to use RGBA8 for a default format though, so information for textures that use alpha clamping is also present and users can still decide to ignore it if they want to. If we don't and send RGB8 from `gz::common` the alpha information will be silently dropped and not retrievable by `gz::rendering` users

iche033:
> My instinct would be to use RGBA8 for a default format though

ok yeah that's sounds good to me

luca-della-vedova:
Ok done! The API is [here](https://github.com/gazebosim/gz-rendering/pull/642/files#diff-c935009af544dd4595666b1ad9072165ae6bb5e9af5e52bcbf6ccdc73fbb7328R223-R232).

It not super ideal since I believe we can't change the current Texture getter since that would break downstream users (maybe could do a tick/tock cycle?):
```
       public: virtual std::string Texture() const = 0;
```
As an alternative I added another getter to get the texture data
```
       public: virtual std::shared_ptr<common::Image> TextureData() const = 0;
```

And changed the `SetTexture` call to take as an input a pointer to a `gz::common::Image`, set to `nullptr` by default to make sure current behavior is unaffected.

If this looks good I'll implement a similar update for PBR textures, as well as OGRE1

iche033:
that looks good to me! I think the `TextureData` getter function is fine actually. 

for OGRE1, you should be able to create a manual texture and load raw data to it. Here is an example: https://github.com/gazebosim/gz-rendering/blob/ign-rendering6/ogre/src/OgreDistortionPass.cc#L286-L405

luca-della-vedova:
Thanks for the pointer! It made things a lot easier.

Still, OGRE1 doesn't really work that well, specifically I had to add a workaround to swap red and blue [here](https://github.com/gazebosim/gz-rendering/pull/642/commits/8ee91a98fe92598b6fe9c02f8334dcecbe16ce0e#diff-1381d5bf7fc1839a3105873dc6f6158cbcce87fbfd4c2ec69a24f3f76a815124R606-R614). It seems that regardless of whether I specify the pixel format as `PF_R8G8B8A8` or `PF_B8G8R8A8` the behavior on the OGRE1 side is unchanged and the model above looks "very blue".

image

After the workaround the colors look better but some models (i.e. the nurse here) seem to have issues, probably due to the fact that it has an RGBA texture (so with an alpha channel)

image

I guess I'll focus on getting the implementation as good / clean as possible with OGRE2

iche033:
> Still, OGRE1 doesn't really work that well, specifically I had to add a workaround to swap red and blue here. It seems that regardless of whether I specify the pixel format as PF_R8G8B8A8 or PF_B8G8R8A8 the behavior on the OGRE1 side is unchanged and the model above looks "very blue".

hmm that's weird. Yeah getting OGRE2 to work first sounds good. You can ticket an issue for the OGRE1 implementation afterwards

ogre/src/OgreMaterial.cc
```diff
@@ -323,7 +329,8 @@ std::string OgreMaterial::NormalMap() const
 }
 
 //////////////////////////////////////////////////
-void OgreMaterial::SetNormalMap(const std::string &_name)
+void OgreMaterial::SetNormalMap(const std::string &_name,
+  const std::shared_ptr<common::Image>& /*_img*/)
```
luca-della-vedova, line 333: I didn't add the NormalMapData getter / copy the member variable since it seems normal maps are not implemented in OGRE1 anyway

include/gz/rendering/base/BaseMaterial.hh
```diff
@@ -960,14 +1022,17 @@ namespace gz
...
+      this->SetRoughnessMap(_material->RoughnessMap(),
+          _material->RoughnessMapData());
+      this->SetMetalnessMap(_material->MetalnessMap(),
+          _material->MetalnessMapData());
       this->SetRoughness(_material->Roughness());
       this->SetMetalness(_material->Metalness());
       this->SetEnvironmentMap(_material->EnvironmentMap());
-      this->SetEmissiveMap(_material->EmissiveMap());
+      this->SetEmissiveMap(_material->EmissiveMap(),
+          _material->EmissiveMapData());
```
luca-della-vedova, line 1035: Not all the maps are implemented, specifically:

* Environment and Lightmap don't have a setter / getter for using gz::common::Image, this is intended (for now) since the primary use for this is using the GLTF format which can embed textures, however the [material specification](https://github.com/KhronosGroup/glTF/blob/main/specification/2.0/schema/material.schema.json) shows no mention of environment or light maps
* AmbientOcclusion _is_ in the specification but seems not to be supported at the `gz-rendering` level. Since from what I understand adding virtual member variables to Material / BaseMaterial breaks ABI we could add them and set them to noop if there is support in any rendering engine, to avoid having to postpone that to a next release.

luca-della-vedova (commented):
OK I think this is pretty much ready for review!

To test it you will need the dependency PRs in [gz-cmake](https://github.com/gazebosim/gz-cmake/pull/278) and [gz-common](https://github.com/gazebosim/gz-common/pull/372), as well as a few assets in GLB, personally I tested with the attached ones:

[GLBModels.zip](https://github.com/gazebosim/gz-rendering/files/9165587/GLBModels.zip)

* `AmbulanceStretcher` is a PBR GLTF asset, under Ubuntu 20.04 and the older version of assimp 5.0.x it will only have a diffuse and normal texture, under Ubuntu 22.04 and `assimp` gltf PBR support it will also have metalness and roughness maps.
* `StaffFemale_CheckClipboard` An animated human with PBR textures, same as above for PBR support but also include animations. Also provided as a `dae` to make sure animations in Collada and GLB look the same.

Other assets can be found quite easily, the [Khronos repo](https://github.com/KhronosGroup/glTF-Sample-Models) has quite a variety of them.

All assets are provided as both `gltf` and `glb`.
`gltf` is made of a readable json that includes the main fields that describe the mesh and points to a binary file for the binary data (i.e. vertices, normals, texture cooridnates) and image files for textures.
`glb` includes all the data bundled in a single binary. `gltf` doesn't use the API in this branch since textures are external images but `glb` does, if all works well they should look exactly the same.

include/gz/rendering/Material.hh
```diff
@@ -220,8 +221,16 @@ namespace gz
       public: virtual std::string Texture() const = 0;
 
       /// \brief Set the material texture
-      /// \param[in] _texture URI of the new texture file
-      public: virtual void SetTexture(const std::string &_texture) = 0;
+      /// \param[in] _texture URI of the new texture file or identifying name
+      /// if _img is set
+      /// \param[in] _img Image data
+      public: virtual void SetTexture(const std::string &_texture,
+        const std::shared_ptr<common::Image> &_img = nullptr) = 0;
```
chapulina, line 228: I think this change needs a tick-tock, otherwise downstream projects who currently implement `SetTexture` with the old signature on Fortress won't compile against Garden without making changes to the code.

How about keeping the old function and making this one like this:

```cpp
      public: virtual void SetTexture(const std::string &_texture,
        const std::shared_ptr<common::Image> &_img)
     {
       this->SetTexture(_texture);
     }
```


i.e. not pure virtual and without a default value.

include/gz/rendering/Material.hh
```diff
@@ -220,8 +221,16 @@ namespace gz
...
+
+      /// \brief Get the texture data
+      /// \return Pointer to the common::Image with the data if the texture
+      /// was loaded from memory
+      public: virtual std::shared_ptr<common::Image> TextureData() const = 0;
```
chapulina, line 233: Same question as the `gz-common` PR about why this is a shared pointer, and whether the image should be const.

include/gz/rendering/base/BaseMaterial.hh
chapulina: > Since from what I understand adding virtual member variables to Material / BaseMaterial breaks ABI we could add them and set them to noop if there is support in any rendering engine, to avoid having to postpone that to a next release.

+1, I think it's worth it adding them as noop for now to keep the API consistent and future-proof.

chapulina (changes requested):
> To test it you will need the dependency PRs in https://github.com/gazebosim/gz-cmake/pull/278 and https://github.com/gazebosim/gz-common/pull/372, as well as a few assets in GLB, personally I tested with the attached ones:

Could you add an example to the `examples` folder that loads that for easier testing? Bonus for an accompanying tutorial under `tutorials` 😄

include/gz/rendering/Material.hh
luca-della-vedova: I thought that adding a default argument would be OK, the function should be able to be called with either:

```
// Old behavior, uses default parameter of _img = nullptr;
SetTexture(name);
// New behavior, assigns an image
SetTexture(name, image);
```

chapulina: It's ok for whoever calls the function, but classes inheriting from `Material` will need to update the function signature

luca-della-vedova: I implemented a similar suggestion for light maps in https://github.com/gazebosim/gz-rendering/pull/642/commits/a9be29c93ea27115551fa02711cc8a8e4d7b0f8c, if it looks good I'll make the same change for all the other functions as well.
Should I add deprecation warnings to the pure virtual methods so we can tock in Gazebo H?

chapulina: Cool! I'm ok adding a deprecation warning. Let's just make that not-pure-virtual anymore, so subclasses on Garden can opt out of it.

luca-della-vedova: Done, updated all of them! https://github.com/gazebosim/gz-rendering/pull/642/commits/19c542727ec168cdd1e8521bf05415f8b3c34f1b I had to change the implementation to a noop instead of calling the deprecated function otherwise there would be a lot of deprecation warnings printed everytime someone imported `Material.hh`

luca-della-vedova:
> Could you add an example to the `examples` folder that loads that for easier testing? Bonus for an accompanying tutorial under `tutorials` smile

Done! Since this is all handled by the MeshManager I just added another mesh to `mesh_viewer`, if that makes sense.
The mesh will render differently depending on whether you are on assimp 5.0.x (it will only have a diffuse component) or assimp 5.2.x (it will have full PBR with normal, metalness and roughness maps).
Is this sufficient or were you thinking about having an example with the full API of manually setting materials? It seems for the other mesh we "just" use the output of the MeshManager without going through the trouble of creating materials manually.
It is different for `ogre2_demo` where there is PBR and we create the materials manually but again this is because the formats used don't support PBR natively hence we either populate it internally or cannot have it, in this case all the PBR population is handled by MeshManager and users of the library "just" have to spawn the returned mesh without changing anything.

include/gz/rendering/base/BaseMaterial.hh
luca-della-vedova: Added Environment and Light maps https://github.com/gazebosim/gz-rendering/pull/642/commits/1fa4e88cde41700aed13bfe9f2346e4abf309e47, left is only a single TODO for binary environment maps but it's at the gz-common level since there is no use case for it (I couldn't find any format that supports it, and it can be added later without breaking API / ABI since `gz-common::Pbr` object doesn't rely on virtual functions).

From what I understand Ambient Occlusion and Light maps are quite similar and indeed `assimp` puts the glTF `ambientOcclusion` map inside the `LIGHTMAP` tag, hence that's where they are currently loaded.
There seems to be really no support anywhere in the stack for Ambient Occlusion maps in any of the parsers, sdformat, or rendering functions, so I didn't add those.

luca-della-vedova:
Good news! It seems the OGRE1 issue was a missing data copy caught by unit tests, now this works as intended in OGRE1 as well

image

include/gz/rendering/Material.hh
```diff
@@ -220,8 +221,31 @@ namespace gz
       public: virtual std::string Texture() const = 0;
 
       /// \brief Set the material texture
-      /// \param[in] _texture URI of the new texture file
-      public: virtual void SetTexture(const std::string &_texture) = 0;
+      /// \param[in] _texture URI of the new texture file or identifying name
+      /// if _img is set
+      public: virtual void GZ_DEPRECATED(7) SetTexture(
+        const std::string &_texture)
```
iche033, line 227: just thinking about this, we could offer an API that defaults the `_img` arg to `nullptr` then we don't need to do the deprecation? We'll just have one `SetTexture` function:

```
virtual void SetTexture(const std::string &_texture, const std::shared_ptr<const common::Image> &_img = nullptr)
```

User's old code should still work. This breaks ABI but that should be fine since it's a new version.

examples/mesh_viewer/Main.cc
```diff
@@ -66,7 +66,16 @@ void buildScene(ScenePtr _scene)
   MeshPtr meshGeom = _scene->CreateMesh(descriptor);
   mesh->AddGeometry(meshGeom);
   root->AddChild(mesh);
-//! [create a mesh]
```
iche033, line 69: I think this is needed as we need to wrap the code in the same tag?

examples/mesh_viewer/Main.cc
```diff
@@ -66,7 +66,16 @@ void buildScene(ScenePtr _scene)
...
+
+//! [create a glb pbr mesh]
+  mesh = _scene->CreateVisual();
+  mesh->SetLocalPosition(3, 2, 0);
+  mesh->SetLocalRotation(0, 0, 0);
+  descriptor.meshName = common::joinPaths(RESOURCE_PATH, "AmbulanceStretcher.glb");
+  descriptor.mesh = meshManager->Load(descriptor.meshName);
+  meshGeom = _scene->CreateMesh(descriptor);
+  mesh->AddGeometry(meshGeom);
+  root->AddChild(mesh);
```
iche033, line 78: add another `//! [create a glb pbr mesh]` here?

Alternatively remove the one above since it does not look like we are referring to it in the tutorial.

include/gz/rendering/Material.hh
luca-della-vedova: From what I understand @chapulina 's proposal for the current API [here](https://github.com/gazebosim/gz-rendering/pull/642#discussion_r930175716) was motivated by the fact that if the signature of the function was changed, users who inherit from `Material` and have a 

```
virtual void SetTexture(const std::string &_texture) override;
```

would end up having a compile error without a formal "tick" since their `override` function does not override anymore.
I can see arguments both ways so not sure what's better

iche033: oh I see, I missed that. Ok looks like we'll need to keep the existing one then. Can its implementation just call 

```
this->SetTexture(_texture, nullptr)
```

examples/mesh_viewer/Main.cc
luca-della-vedova: Done! https://github.com/gazebosim/gz-rendering/pull/642/commits/1fe7958360432922e3faecdf4fe7565d3cd28a8c

include/gz/rendering/Material.hh
luca-della-vedova: Good point! I gave it a quick spin in the beginning but I had some trouble with undefined references that made me just create empty functions for all of them.
Done https://github.com/gazebosim/gz-rendering/pull/642/commits/334b28a250ac75f32d03e71586849fcdf3314d94

iche033:
example mesh works for me on Focal with `ogre` and `ogre2`

mesh_viewer_glb

iche033:
Changes look good as they are now. Just waiting for upstream PRs and green CI.

One minor note that shouldn't block this PR from getting in is if we could keep the original `SetTexture` / `Set*Map` calls and not deprecate them since this version has been the common use case till now. Now we require existing downstream users to change the call to `SetTexture("tex", nullptr)` where the second arg is just `nullptr`.

chapulina (approved):

chapulina:
61 Windows failures 😩 Merging 😬

onurtore:
Tested locally, LGTM.

PR #670, opened 2022-07-13
Enable ogre2 heightmap test

iche033:

# 🦟 Bug fix

## Summary

Run `Heightmap_test` with `ogre2` now that it is supported

This unfortunately reduces the overall test coverage % because various ogre1.x code is not run any more.

## Checklist
- [x] Added tests

### DIFF
--- a/ogre2/include/ignition/rendering/ogre2/Ogre2Heightmap.hh
+++ b/ogre2/include/ignition/rendering/ogre2/Ogre2Heightmap.hh
@@ -64,6 +64,9 @@ namespace ignition
       // Documentation inherited.
       public: virtual void PreRender() override;
 
+      // Documentation inherited.
+      public: virtual void Destroy() override;
+
       /// \brief Returns the Terra pointer as it is a movable object that
       /// must be attached to a regular SceneNode
       /// \remarks This behavior is different from ogre1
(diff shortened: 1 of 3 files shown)


chapulina:
> This unfortunately reduces the overall test coverage % because various ogre1.x code is not run any more.

If we're not planning to test both engines, then we should exclude the other one from test coverage. I think we should test all engines that we officially support though. What do you think?

iche033:
> I think we should test all engines that we officially support though.

The problem we ran into before was that we were not able to run ogre and ogre2 tests in the same process, i.e. one after another since the plugins can't be loaded at the same time. We could look into generating tests for each render engine so that they are run in separate processes, e.g. UNIT_ogre_Scene_TEST and UNIT_ogre2_Scene_TEST.

chapulina:
How about we ticket an issue for that, and for now we exclude Ogre1 and Optix from the coverage reports?

iche033:
> How about we ticket an issue for that, and for now we exclude Ogre1 and Optix from the coverage reports?

yep ticketed #683  and created https://github.com/gazebosim/gz-rendering/pull/684 to ignore certain folders in text coverage

chapulina (approved):

PR #676, opened 2022-07-19
Fix double to float conversion beyond float max

chapulina:
* Follow up to https://github.com/gazebosim/gz-rendering/pull/655
* Related to
    * https://github.com/gazebosim/gz-sim/pull/1499#pullrequestreview-1009122496
    * https://github.com/gazebosim/gz-sim/issues/1072

---

Gazebo is using double precision, while Ogre is using floats. When we go beyond the limits of what a `float` can hold, some compilers will generate `inf`s when converting those `double`s to `float`s. But internally, Ogre doesn't handle inf values well, which results in crashes and exceptions (like https://github.com/gazebosim/gz-sim/issues/1072).

The solution I found was to guard some of our Gazebo ➡️ Ogre conversions against infs, but I'm open to other ideas.

This solved the issue in https://github.com/gazebosim/gz-sim/pull/1499#pullrequestreview-1009122496 for me, but since that's tricky to reproduce, I'd appreciate confirmation from @jennuine or @AzulRadio that it works for them too.

---

Opening for visibility so others can try it out. If the idea is accepted, I'll open a separate PR to Citadel and we can forward-port that one.

Do not merge.

### DIFF
--- a/ogre2/src/CMakeLists.txt
+++ b/ogre2/src/CMakeLists.txt
@@ -69,6 +69,12 @@ if (USE_GLX)
 endif()
 
 # Build the unit tests
-ign_build_tests(TYPE UNIT SOURCES ${gtest_sources} LIB_DEPS ${ogre2_target})
+ign_build_tests(TYPE UNIT
+    SOURCES
+      ${gtest_sources}
+    LIB_DEPS
+      ${ogre2_target}
+      IgnOGRE2::IgnOGRE2
+)
 
 install(DIRECTORY "media"  DESTINATION ${IGN_RENDERING_RESOURCE_PATH}/ogre2)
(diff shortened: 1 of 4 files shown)


iche033:
We have a few places in code that depend on values being +inf / -inf, e.g. Ogre2LidarVisual and various shaders. Looks like they may not be affected by the current changes but silent conversion to non-inf values may introduce issues that are hard to debug in the future. 

Is it possible to limit the `infToMax` calls in the Ogre2Node class after doing the conversion? 

Alternatively, we can check to see if the input double value is larger than max float value, and if it is, convert to max float value, otherwise leave it as inf. Something like:

```
float out_value
if !isinf(double_value) && double_value > max_float
    out_value = max_float
else
    out_value = static_cast<float>(double_value)
```

chapulina:
I'll just close this since it didn't fix the original issue 😕

PR #694, opened 2022-07-29
Port remaining tests to common framework

mjcarroll:
Follow up to #685.

There were two remaining tests that didn't get ported over to utilize the common framework.  These have been updated and we can remove the test_config.hh

Signed-off-by: Michael Carroll <michael@openrobotics.org>


iche033:
one more `RENDER_ENGINE_VALUES` here:
https://github.com/gazebosim/gz-rendering/blob/ign-rendering6/.github/ci/after_make.sh#L9

iche033:
I think something changed. The tests are failing: https://build.osrfoundation.org/job/ignition_rendering-ci-pr_any-ubuntu_auto-amd64/2580/

mjcarroll:
I have disabled the remaining 3 tests that fail on macOS metal and tracked them in #695

iche033 (commented):
I was testing this with `ogre` and `ogre2` on Focal and they pass locally. Changes look good to me.

PR #698, opened 2022-08-03
Destroy GpuRay camera on exit

iche033:

# 🦟 Bug fix

## Summary
Destroy camera on exit in Ogre2GpuRay class, which is also done in gz-rendering6 [here](https://github.com/gazebosim/gz-rendering/blob/ign-rendering6/ogre2/src/Ogre2GpuRays.cc#L498). This also ensures the `Destroy` function can be called multiple times.

This should fix the gpu lidar test in gz-sensors.

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -665,12 +665,7 @@ void Ogre2GpuRays::Destroy()
   if (this->scene)
   {
     Ogre::SceneManager *ogreSceneManager = this->scene->OgreSceneManager();
-    if (ogreSceneManager == nullptr)
-    {
-      ignerr << "Scene manager not available. "
-             << "Unable to remove cameras and listeners" << std::endl;
-    }
-    else
+    if (ogreSceneManager)
     {
       ogreSceneManager->destroyCamera(this->dataPtr->ogreCamera);
       this->dataPtr->ogreCamera = nullptr;
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -661,7 +661,22 @@ void Ogre2GpuRays::Destroy()
     ogreCompMgr->removeWorkspace(this->dataPtr->ogreCompositorWorkspace2nd);
     this->dataPtr->ogreCompositorWorkspace2nd = nullptr;
   }
+
+  if (this->scene)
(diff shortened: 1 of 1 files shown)


ogre2/src/Ogre2GpuRays.cc
```diff
@@ -661,7 +661,22 @@ void Ogre2GpuRays::Destroy()
...
     this->dataPtr->ogreCompositorWorkspace2nd = nullptr;
   }
+
+  if (this->scene)
+  {
+    Ogre::SceneManager *ogreSceneManager = this->scene->OgreSceneManager();
+    if (ogreSceneManager == nullptr)
+    {
+      ignerr << "Scene manager not available. "
+             << "Unable to remove cameras and listeners" << std::endl;
```
chapulina, line 671: We don't print errors for other missing things in the `Destroy` function. Just checking that there's a reason to treat the scene manager differently.

ogre2/src/Ogre2GpuRays.cc
```diff
@@ -661,7 +661,22 @@ void Ogre2GpuRays::Destroy()
     ogreCompMgr->removeWorkspace(this->dataPtr->ogreCompositorWorkspace2nd);
     this->dataPtr->ogreCompositorWorkspace2nd = nullptr;
   }
+
+  if (this->scene)
```
chapulina, line 665: Should we also check this?

```suggestion
  if (this->scene && this->dataPtr->ogreCamera)
```

chapulina (approved):
LGTM with 🟢 , I just have a couple of questions that are not blockers.

ogre2/src/Ogre2GpuRays.cc
iche033: there is a [check](https://github.com/gazebosim/gz-rendering/blob/f2f2ced9fe322f9897f752b731e7508665651025/ogre2/src/Ogre2GpuRays.cc#L588) earlier in the function that should cover this

ogre2/src/Ogre2GpuRays.cc
iche033: not really. I removed the error msg print line. [734dfed](https://github.com/gazebosim/gz-rendering/pull/698/commits/734dfed35d00b4dbac853c0d26fbe9cdd3e4d0d0)

PR #700, opened 2022-08-08
Expose camera intrinsic matrix API

deepanshubansal01:
Signed-off-by: deepanshu <deepanshubansal01@gmail.com>

# 🦟 Bug fix

Fixes https://github.com/gazebosim/gz-sim/issues/705

## Summary
This PR solves the camera intrinsics not being reflected properly in `/camera_info` topic when the `<intrinsics>` tag is not provided. The PR exposes the `CameraIntrinsicMatrix()` which can be used by downstream packages `gz-sensors` to update the DOM object and also the `/camera_info` topic with correct intrinsics. 

## Depends on
https://github.com/gazebosim/gz-sensors/pull/256

## Checklist
- [x] Added tests


adityapande-1995 (commented):
Could you add some tests for this one ?

deepanshubansal01:
> Could you add some tests for this one ?

Sounds good.

include/gz/rendering/base/BaseCamera.hh
```diff
@@ -498,6 +501,32 @@ namespace gz
...
+
+      double inverseWidth = 1.0 / (right - left);
+      double inverseHeight = 1.0 / (top - bottom);
+
+      double fX = projectionMat(0, 0)/ (2*inverseWidth);
+      double fY = projectionMat(1, 1) / (2*inverseHeight);
+      double cX = -((projectionMat(0, 2) -
+          ((right + left) / (right - left))) * (right - left)) / 2;
+      double cY = this->ImageHeight() + ((projectionMat(1, 2) -
+          ((top + bottom) / (top - bottom))) * (top - bottom)) / 2;
```
jacobperron, line 523: Do you have an online reference for the math? It looks like we can simplify, considering we have the image width and height directly, e.g.

```suggestion
      const double width = static_cast<double>(this->ImageWidth());
      const double height = static_cast<double>(this->ImageHeight());

      double fX = (projectionMat(0, 0) * width) / 2.0;
      double fY = (projectionMat(1, 1) * height) / 2.0;
      double cX = (-1.0 * width * (projectionMat(0, 2) - 1.0) / 2.0;
      double cY = height + (height * (projectionMat(1, 2) - 1)) / 2.0;
```

deepanshubansal01: I agree we can simplify but the only I had it this way so that it is easier to relate to the original `ProjectionMatrix()` calculation since that's how ogre and gazebo is using the term names. Here are some links for reference, though there is no specific link I think to calculate the intrinsics from `ProjectionMatrix()`, I just decoupled it. 

I can simplify the calculation as well not sure. 

https://ogrecave.github.io/ogre/api/13/class_ogre_1_1_math.html#a23c7798ef5a11dea62318b5be2983c2d

adityapande-1995: It would be nice to simplify the math here, and add the link as reference in a comment.

adityapande-1995:
Looks good to me overall, with the math simplified and the calculation documented !

test/common_test/Camera_TEST.cc
```diff
@@ -298,3 +298,60 @@ TEST_F(CameraTest, VisibilityMask)
...
+  EXPECT_EQ(camera->ImageHeight(), height);
+  EXPECT_EQ(camera->ImageWidth(), width);
+  EXPECT_NEAR(camera->HFOV().Radian(), hfov, error);
+
+  // Verify if intrinsics have changed
+  cameraIntrinsics = camera->CameraIntrinsicMatrix();
+  EXPECT_NEAR(cameraIntrinsics(0, 0), 866.223, error);
+  EXPECT_NEAR(cameraIntrinsics(1, 1), 866.223, error);
+  EXPECT_DOUBLE_EQ(cameraIntrinsics(0, 2), 500);
+  EXPECT_DOUBLE_EQ(cameraIntrinsics(1, 2), 500);
```
adityapande-1995, line 353: Can we have a comparison between the "fake" intrinsic matrix used by OpenGL and the actual intrinsic matrix ? We can explain the difference as a comment.

include/gz/rendering/base/BaseCamera.hh
adityapande-1995: Done https://github.com/gazebosim/gz-rendering/pull/700/commits/083883f1a90ca02c2892cfb52824f9b5121cf49d

jacobperron (approved):

adityapande-1995:
This is to be reverted in favour of https://github.com/gazebosim/gz-rendering/pull/755

PR #831, opened 2023-03-09
Use std::vector instead of std::map in BaseStorage

nkoenig:
# 🎉 New feature

## Summary

Most of the time the `store` std::map inside BaseStorage is iterated over, which is not as efficient as a std::vector.
 
## Checklist
- [ ] Added tests

### DIFF
--- a/include/gz/rendering/base/BaseStorage.hh
+++ b/include/gz/rendering/base/BaseStorage.hh
@@ -165,21 +165,61 @@ namespace gz
       /// \returns Iterator to end
       public: virtual UIter End();
 
+      /// \brief Get a const iterator to the object by pointer.
+      /// Note that the iterator may change or become invalid when objects are
+      /// added to / removed from the store.
+      /// \param[in] _object Object pointer
+      /// \return Const iterator to the object
       protected: virtual ConstUIter ConstIter(ConstTPtr _object) const;
 
+      /// \brief Get a const iterator to the object by id.
+      /// Note that the iterator may change or become invalid when objects are
+      /// added to / removed from the store.
+      /// \param[in] _id Object id
+      /// \return Const iterator to the object
       protected: virtual ConstUIter ConstIterById(unsigned int _id) const;
 
+      /// \brief Get a const iterator to the object by name.
+      /// Note that the iterator may change or become invalid when objects are
+      /// added to / removed from the store.
+      /// \param[in] _name Object name
+      /// \return Const iterator to the object
       protected: virtual ConstUIter ConstIterByName(
                      const std::string &_name) const;
 
+      /// \brief Get a const iterator to the object by index.
+      /// Note that the iterator may change or become invalid when objects are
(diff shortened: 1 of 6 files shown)


include/gz/rendering/base/BaseStorage.hh
```diff
@@ -854,7 +857,12 @@ namespace gz
     typename BaseStore<T, U>::ConstUIter
     BaseStore<T, U>::ConstIterByName(const std::string &_name) const
     {
-      return this->store.find(_name);
+      auto idx = this->storeMap.find(_name);
+      if (idx == this->storeMap.end())
+      {
+        return this->store.end();
```
azeey, line 863:
```suggestion
        return this->storeMap.end();
```

or `return idx;`

include/gz/rendering/base/BaseStorage.hh
```diff
@@ -854,7 +857,12 @@ namespace gz
...
+      }
+      return ConstIterByIndex(idx->second);
```
azeey, line 865: I see we're returning iterators. If the iterators are being stored somewhere, we need to be careful since insertion to the `std::vector` might invalidate it. This was not the case for `std::map`, so we might be introducing subtle bugs here.

include/gz/rendering/base/BaseStorage.hh
nkoenig: I think it's suppose to be `this->store.end();` because the return type is `ConstUIter` which is typedef'd to `std::vector`.

include/gz/rendering/base/BaseStorage.hh
iche033: Since the original implementation was based on a map, it makes sense to keep iters and pass them around. Now that we're changing to vector, I think we don't need to be doing this as much. The only place that use these `IterByName` functions internally is `ContainsName`, which can by re-implemented easily.

So how about we deprecate `ConstIterByName` / `IterByName` functions, update doxygen comment for these 2 APIs to warn users not to keep a handle to the iterator as it may change, and add a note to migration guide?

iche033:
`UNIT_Scene_TEST` is failing, could be related

include/gz/rendering/base/BaseStorage.hh
iche033: ok I created a PR: https://github.com/gazebosim/gz-rendering/pull/832

I didn't deprecate the functions. Instead I added doc with a note that iterators can change / become invalid. I also updated migration guide.

iche033:
> `UNIT_Scene_TEST` is failing, could be related

fixed in https://github.com/gazebosim/gz-rendering/pull/832

iche033 (approved):

nkoenig:
@azeey , are you okay with this PR?

azeey:
If Ian has approved it, it's good with me. I was just quickly looking through and noticed iterators being returned and alarm bells went off in my head.

Issue #880, opened 2023-08-15
Gizmo uses a transform identifier not included in `TransformAxis`


azeey:
## Environment
* OS Version: all
* Source or binary build? all

## Description
* Expected behavior:  `GizmoVisual` uses a transform identifier defined in `TransformAxis`.
* Actual behavior: As seen in https://github.com/gazebosim/gz-rendering/blob/e58870ce3ea82711fc42777709e2170341b1eb3c/include/gz/rendering/base/BaseGizmoVisual.hh#L231, `GizmoVisual` uses `TransformAxis::TA_TRANSLATION_Z << 1` which is not documented.

iche033:
hmm the logic there is to find a unique key that's not already used by `this->visuals`. I'll add some comments to it

iche033:
Addressed by #881, and #882 (see related comment: https://github.com/gazebosim/gz-sim/pull/2051#discussion_r1296506004)

PR #881, opened 2023-08-15
Add minor comments to BaseGizmoVisual

iche033:
# 🦟 Bug fix

Related issue: https://github.com/gazebosim/gz-rendering/issues/880

## Summary
Add some minor comments to explain what the TransformAxis enum << 1 means. 

We could also define the values as static const variables, e.g. `this->visuals[kTranslationOriginKey] = transOrigin;`

## Checklist
- [ ] Added tests


include/gz/rendering/base/BaseGizmoVisual.hh
```diff
@@ -495,6 +495,8 @@ namespace ignition
       this->visuals[TransformAxis::TA_TRANSLATION_X] = transXVis;
       this->visuals[TransformAxis::TA_TRANSLATION_Y] = transYVis;
       this->visuals[TransformAxis::TA_TRANSLATION_Z] = transZVis;
+      // Store the translation origin visual in this->visuals using a key
```
azeey, line 498: As a user of `GizmoVisual`, it's difficult to know how to access the origin without reading the code. I would suggest adding something in `ChildByAxis` or `TransformAxis` itself.

iche033: There is not a enum for it in `TransformAxis` or accessible by `ChildByAxis` because it's currently not meant to be exposed to the public. So checking to see what the use case is for accessing the GizmoVisual's origin visual? Do you need to change the origin's visual properties like color or size? If we want to expose this, we'll need to target `main`.

Henrique-BO: Actually, that gz-sim PR accesses the visual in `TA_ROTATION_Z << 1`, which is a gray circle that has to be rotated just like the other rotation circles, and so might need to be public. 

This is also done in `TransformController`: https://github.com/gazebosim/gz-rendering/blob/e58870ce3ea82711fc42777709e2170341b1eb3c/src/TransformController.cc#L161-L162

The origin visual referenced by `TA_TRANSLATION_Z << 1` isn't used in those cases.

iche033: as per https://github.com/gazebosim/gz-sim/pull/2051#discussion_r1296364794, I'll do some refactoring of code so we can avoid exposing the confusing `TransformAxis::TA_ROTATION_Z << 1` value

azeey (approved):

PR #950, opened 2024-01-04
Correctly set position-independent code

mjcarroll:
This was broken by https://github.com/gazebosim/gz-cmake/pull/399, since we dropped the gazebo-specific flag, we can use native CMake mechanism for marking this library position-independent.


ahcorde (approved):

mjcarroll:
Window failures are existing here.

PR #955, opened 2024-01-13
Improve Ogre2GpuRays performance

iche033:
# 🦟 Bug fix

## Summary

Tuned Ogre2GpusRays' (lidar) ogre2 compositor to bypass computationally expensive phases involving lights. Since we are just generating depth data, we do not need to care about the lights in the scene so we can skip these computations.

The compositor parameter description can be found in this [ogre doc](https://ogrecave.github.io/ogre-next/api/2.3/compositor.html). 
* `enable_forwardplus` - affects performance when there are many lights in the scene. However we only care about depth data for lidars (and not the rgb color images that are affected by lights)  so set this to be disabled. Specifically, a big chunk of time was found to be spent in ogre scene manager's [`_cullPhase01`](https://github.com/OGRECave/ogre-next/blob/62fa784b5b7fcd2449b81e90ee67d666cf94e985/OgreMain/src/OgreSceneManager.cpp#L1258) function. Disabling this flag gave the most perf boost. 
    * **Note** the ogre doc lists `enable_plus` which is a typo. 
* `light_visibility_mask` - tells ogre that we don't want to see any lights

I used this [lidar_profile.sdf](https://gist.github.com/iche033/cdbe559b747a152f73dec901ff9b54a6) world for profiling. It consists of a single lidar with 160 degrees FOV and 640 samples. 
  * Note: To simulate this lidar, internally `Ogre2GpuRays` uses **3** environment cameras for collecting depth data - the number of env cameras is important as it directly affects to total render time. If a lidar has a smaller FOV (<90), only 1 env camera is needed and the update time would be close to ~1/3 of the time shown below.

Here're before and after plots of time taken to do a [Render](https://github.com/gazebosim/gz-rendering/blob/eeaf6e814d83a9e7fa0332747ce3b2a40660633e/ogre2/src/Ogre2GpuRays.cc#L1246-L1267) operation
lidar_profile

The time taken is now roughly 2-4ms for each render update. The performance is still behind ogre 1.x in which the render operation for the same lidar takes ~1ms

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2GpuRays.cc
+++ b/ogre2/src/Ogre2GpuRays.cc
@@ -19,7 +19,6 @@
 #include <gz/math/Vector3.hh>
 
 #include <gz/common/Console.hh>
-#include <gz/common/Timer.hh>
 #include <gz/math/Helpers.hh>
 
 #include "gz/rendering/ogre2/Ogre2Camera.hh"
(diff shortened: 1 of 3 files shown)


ahcorde (approved):

ahcorde:
@iche033 can we backport this to fortress ?

iche033:
> @iche033 can we backport this to fortress ?

yep I'll do that

PR #965, opened 2024-01-25
Optimize Ogre2DepthCamera performance by skipping color target passes when there are no connections

iche033:
# 🦟 Bug fix

## Summary

Performance optimization - Ogre2DepthCamera is capable of generating color images in addition to depth and the color data are packed into a point cloud buffer. So if there are no point cloud connections, we can skip computations needed to generate the color rgb images.

Also added changes to skip light-related computations in the same way as lidars in https://github.com/gazebosim/gz-rendering/pull/955

More info:

This is done by using ogre2's "execution mask". If the render passes execution mask matches the workspace's execution mask, the passes get executed. So by default, the color target's render passes' execution mask is set to something that does not match the workspace execution mask so we can skip computation for the color passes. When point cloud connections is detected, the passes' execution mask is updated to match the workspace's mask.

## To Test

This is to be used with https://github.com/gazebosim/gz-sensors/pull/413. You can see graphs in that pull request showing the performance improvements for a depth camera and a RGBD camera in gz sim.

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2DepthCamera.cc
+++ b/ogre2/src/Ogre2DepthCamera.cc
@@ -167,7 +167,7 @@ class gz::rendering::Ogre2DepthCameraPrivate
   /// target will be updated to match the workspace's execution mask so that
   /// these passes are executed, otherwise they will be skipped for performance
   /// improvement.
-  public: const uint8_t kDepthExecutionMask = 0xFE;
+  public: const uint8_t kDepthExecutionMask = 0xEF;
 
   /// \brief Pointer to the color target in the workspace
   public: Ogre::CompositorTargetDef *colorTarget{nullptr};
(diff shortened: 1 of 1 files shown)


ahcorde (approved):

ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -147,6 +161,17 @@ class gz::rendering::Ogre2DepthCameraPrivate
 
   /// \brief Name of shadow compositor node
   public: const std::string kShadowNodeName = "PbsMaterialsShadowNode";
+
+  /// \brief Execution mask for this workspace
+  /// If RGB point color data are requested, the execution mask of the color
+  /// target will be updated to match the workspace's execution mask so that
+  /// these passes are executed, otherwise they will be skipped for performance
+  /// improvement.
+  public: const uint8_t kDepthExecutionMask = 0xEF;
```
ahcorde, line 170: include `<cstdint>`

iche033: added [fd1db40](https://github.com/gazebosim/gz-rendering/pull/965/commits/fd1db40e63a212532db0b5794199c6094d3f0556)

PR #976, opened 2024-03-01
Fix setting normal map when not all submeshes have texcoords (ogre2)

iche033:
# 🦟 Bug fix

## Summary

Setting normal map to a a submesh could fail in ogre2 if the mesh contains other submeshes that do not have texture coordinates set. Ogre2 will fail to generate tangents (needed for normal maps) and throw an exception:

```
03:59:05: OGRE EXCEPTION(5:ItemIdentityException): Cannot locate an appropriate 2D texture coordinate set for all the vertex data in this mesh to create tangents from.  in Mesh::suggestTangentVectorBuildParams at ./OgreMain/src/OgreMesh.cpp (line 1913)
```

The fix is to generate texcoords for all submeshes when importing the `common::SubMesh` into ogre even if the `common::SubMesh` does not have any texcoords. 

The alternative (ideal) fix is to update ogre to just skip generating tangents for those submeshes that do not have texcoords instead of throwing an exception.

Tested with the [AnisotropyBarnLamp.glb](https://github.com/KhronosGroup/glTF-Sample-Assets/tree/main/Models/AnisotropyBarnLamp) model.

Before the fix, the whole metal part of the lamp mesh appears white. After the fix, the PBR textures are applied correctly:

Screenshot 2024-02-29 at 7 54 07 PM

The light bulb should be transparent so there's another bug somewhere else and that still needs to be fixed. For comparison, see the [result rendered by threejs](https://threejs.org/examples/webgl_loader_gltf_anisotropy.html)

## Checklist
- [x] Added tests

### DIFF
--- a/ogre2/src/Ogre2MeshFactory.cc
+++ b/ogre2/src/Ogre2MeshFactory.cc
@@ -382,13 +382,22 @@ bool Ogre2MeshFactory::LoadImpl(const MeshDescriptor &_desc)
 
       // two dimensional texture coordinates
       // add all texture coordinate sets
-      for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
+      if (subMesh.TexCoordSetCount() == 0u)
       {
-        if (subMesh.TexCoordCountBySet(k) > 0)
-        {
           vertexDecl->addElement(0, currOffset, Ogre::VET_FLOAT2,
-              Ogre::VES_TEXTURE_COORDINATES, k);
+              Ogre::VES_TEXTURE_COORDINATES, 0);
           currOffset += Ogre::v1::VertexElement::getTypeSize(Ogre::VET_FLOAT2);
+      }
+      else
+      {
+        for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
+        {
+          if (subMesh.TexCoordCountBySet(k) > 0)
+          {
+            vertexDecl->addElement(0, currOffset, Ogre::VET_FLOAT2,
+                Ogre::VES_TEXTURE_COORDINATES, k);
+            currOffset += Ogre::v1::VertexElement::getTypeSize(Ogre::VET_FLOAT2);
+          }
         }
       }
 
@@ -436,12 +445,20 @@ bool Ogre2MeshFactory::LoadImpl(const MeshDescriptor &_desc)
         }
 
         // Add all texture coordinate sets
-        for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
+        if (subMesh.TexCoordSetCount() == 0u)
         {
-          if (subMesh.TexCoordCountBySet(k) > 0u)
+          *vertices++ = 0;
+          *vertices++ = 0;
+        }
+        else
+        {
+          for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
           {
-            *vertices++ = subMesh.TexCoordBySet(j, k).X();
-            *vertices++ = subMesh.TexCoordBySet(j, k).Y();
+            if (subMesh.TexCoordCountBySet(k) > 0u)
+            {
+              *vertices++ = subMesh.TexCoordBySet(j, k).X();
+              *vertices++ = subMesh.TexCoordBySet(j, k).Y();
+            }
           }
         }
       }
--- a/ogre2/src/Ogre2MeshFactory.cc
+++ b/ogre2/src/Ogre2MeshFactory.cc
@@ -381,13 +381,18 @@ bool Ogre2MeshFactory::LoadImpl(const MeshDescriptor &_desc)
       // TODO(anyone): specular colors
 
       // two dimensional texture coordinates
-      // add all texture coordinate sets
+      // If submesh does not have texcoord sets, add one default set.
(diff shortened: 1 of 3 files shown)


ogre2/src/Ogre2MeshFactory.cc
```diff
@@ -382,13 +382,22 @@ bool Ogre2MeshFactory::LoadImpl(const MeshDescriptor &_desc)
 
       // two dimensional texture coordinates
       // add all texture coordinate sets
-      for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
+      if (subMesh.TexCoordSetCount() == 0u)
```
shameekganguly, line 385: Consider adding a comment explaining why this is needed?

ogre2/src/Ogre2MeshFactory.cc
```diff
@@ -436,12 +445,20 @@ bool Ogre2MeshFactory::LoadImpl(const MeshDescriptor &_desc)
         }
 
         // Add all texture coordinate sets
-        for (unsigned int k = 0u; k < subMesh.TexCoordSetCount(); ++k)
+        if (subMesh.TexCoordSetCount() == 0u)
+        {
+          *vertices++ = 0;
```
shameekganguly, line 450: Do we need to check if the mesh has an associated texture before assigning (0,0) texture coordinate? Will there ever be a situation where this can cause bad access?

shameekganguly (approved):

test/integration/mesh.cc
```diff
@@ -0,0 +1,177 @@
...
+  textureMapData0->SetFromData(red.get(), 1, 1, common::Image::RGB_INT8);
+  material0->SetTextureImage(textureMapName0, textureMapData0);
+
+  common::Pbr pbr;
+  std::string normalMapName = "normal_map";
+  auto normalMapData = std::make_shared<common::Image>();
+  auto normal = std::make_unique<unsigned char[]>(3);
+  normal.get()[0] = 127;
+  normal.get()[1] = 127;
+  normal.get()[2] = 255;
```
ahcorde, line 97:
```suggestion
  normal.get()[0] = 127u;
  normal.get()[1] = 127u;
  normal.get()[2] = 255u;
```

test/integration/mesh.cc
```diff
@@ -0,0 +1,177 @@
...
+  subMesh0.AddTexCoordBySet(math::Vector2d(0, 1), 0);
+  subMesh0.AddTexCoordBySet(math::Vector2d(0, 0), 0);
+  common::MaterialPtr material0;
+  material0 = std::make_shared<common::Material>();
+  std::string textureMapName0 = "red_diffuse_map";
+  auto textureMapData0 = std::make_shared<common::Image>();
+  auto red = std::make_unique<unsigned char[]>(3);
+  red.get()[0] = 255;
+  red.get()[1] = 0;
+  red.get()[2] = 0;
```
ahcorde, line 87:
```suggestion
  red.get()[0] = 255u;
  red.get()[1] = 0;
  red.get()[2] = 0;
```

test/integration/mesh.cc
```diff
@@ -0,0 +1,177 @@
...
+  subMesh1.AddIndex(2);
+
+  common::MaterialPtr material1;
+  material1 = std::make_shared<common::Material>();
+  std::string textureMapName1 = "green_diffuse_map";
+  auto textureMapData1 = std::make_shared<common::Image>();
+  auto green = std::make_unique<unsigned char[]>(3);
+  green.get()[0] = 0;
+  green.get()[1] = 255;
+  green.get()[2] = 0;
```
ahcorde, line 125:
```suggestion
  green.get()[0] = 0;
  green.get()[1] = 255u;
  green.get()[2] = 0;
```

ahcorde (approved):
LGTM, just a minor comment about unsigned varirables

test/integration/mesh.cc
iche033: [4985dbe](https://github.com/gazebosim/gz-rendering/pull/976/commits/4985dbeac5634b1014a623180a6ee2d8178ccfe4)

test/integration/mesh.cc
iche033: [4985dbe](https://github.com/gazebosim/gz-rendering/pull/976/commits/4985dbeac5634b1014a623180a6ee2d8178ccfe4)

test/integration/mesh.cc
iche033: [4985dbe](https://github.com/gazebosim/gz-rendering/pull/976/commits/4985dbeac5634b1014a623180a6ee2d8178ccfe4)

ogre2/src/Ogre2MeshFactory.cc
iche033: added comment in [4985dbe](https://github.com/gazebosim/gz-rendering/pull/976/commits/4985dbeac5634b1014a623180a6ee2d8178ccfe4)

ogre2/src/Ogre2MeshFactory.cc
iche033: this should be fine. Given that we declare that the mesh will now always have texcoords, we have to add these texcoords regardless of whether the mesh has a texture or not.

PR #1019, opened 2024-07-24
Add gamma correction to simple_demo_qml example

athenaz2:
# 🦟 Bug fix

## Summary
When run with Ogre2 and OpenGL, `simple_demo_qml` has darker shadows and the scene is a bit darker overall when compared to `simple_demo`. This is because the texture from the render engine is in sRGB color space, but Qt's textures are created in linear color space. As per https://github.com/gazebosim/gz-gui/pull/629, this fix for `simple_demo_qml` ensures that the render engine texture, originally in sRGB format, is correctly passed to Qt in linear format. Thus, now the `simple_demo_qml` scene is lighter and matches `simple_demo`'s scene.

Before
image

After
simple demo no qml vs yes qml fix gl_gamma

## Checklist
- [ ] Added tests

### DIFF
--- a/examples/simple_demo_qml/ThreadRenderer.cpp
+++ b/examples/simple_demo_qml/ThreadRenderer.cpp
@@ -334,10 +334,6 @@ void TextureNode::PrepareNode()
         // note: include QQuickWindow::TextureHasAlphaChannel if the rendered content
         // has alpha.
 
-        QOpenGLFunctions *glFuncs = QOpenGLContext::currentContext()->functions();
-        glFuncs->glBindTexture(GL_TEXTURE_2D, newId);
-        glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_SRGB_DECODE_EXT, GL_SKIP_DECODE_EXT);
-
 #if QT_VERSION < QT_VERSION_CHECK(5, 14, 0)
 # ifndef _WIN32
 #   pragma GCC diagnostic push
--- a/examples/simple_demo_qml/ThreadRenderer.cpp
+++ b/examples/simple_demo_qml/ThreadRenderer.cpp
@@ -326,6 +326,7 @@ void TextureNode::PrepareNode()
     QSize newSize = this->size;
     this->id = 0;
     this->mutex.unlock();
+
     if (newId)
     {
         delete this->texture;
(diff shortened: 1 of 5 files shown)


examples/simple_demo_qml/ThreadRenderer.cpp
```diff
@@ -321,32 +321,84 @@ void TextureNode::NewTexture(int _id, const QSize &_size)
...
+
+        GLuint texForQt = newId;
+
+        if (this->graphicsAPI == gz::rendering::GraphicsAPI::OPENGL)
+        {
+            // intermediate FBO for copying texture data
+            GLuint rhiFbo = 0;
+            // internal texture ID to hold texture data from render engine texture,
+            // in linear GL_RGB format
+            GLuint rhiId = 0;
```
iche033, line 345: can you make these ids member variables so that we can delete the fbo and texture in the destructor when the app exits? e.g. see code [here](https://github.com/gazebosim/gz-gui/blob/2496041b72e6cc29b2b520ef5aca0c622618cb0a/src/plugins/minimal_scene/MinimalSceneRhiOpenGL.cc#L209-L214)

examples/simple_demo_qml/ThreadRenderer.cpp
```diff
@@ -321,32 +321,84 @@ void TextureNode::NewTexture(int _id, const QSize &_size)
...
+
+            if (newTexFormat == GL_SRGB || newTexFormat == GL_SRGB8 ||
+                    newTexFormat == GL_SRGB_ALPHA || newTexFormat == GL_SRGB8_ALPHA8)
+            {
+                // get currently bound FBO so it can be restored later
+                GLint currentFbo;
+                f->glGetIntegerv(GL_FRAMEBUFFER_BINDING, &currentFbo);
+
+                // bind render engine texture to RHI FBO
+                f->glGenFramebuffers(1, &rhiFbo);
```
iche033, line 364: this will create a new fbo every frame. So related to previous comment, we need to store this id and as member var, and here only create the frame buffer if the fbo does not exist yet, i.e. `rhiFbo` is 0. Same goes for the `glGenTextures` line below

iche033 (approved):

Issue #1007, opened 2024-05-30
INTEGRATION_load_unload_ogre_gl3plus test fails on gz_rendering-ci-pr_any-noble-amd64


iche033:
## Environment
* OS Version: Ubuntu Noble 24.04
* Source or binary build? source build, `main` branch (gz-rendering9)

    - Rendering plugin: ogre. 

## Description
`INTEGRATION_load_unload_ogre_gl3plus` test is failing on `gz_rendering-ci-pr_any-noble-amd64` CI

The test passes on Ubuntu Jammy on github actions and jenkins

## Output

CI console output

```
39:  GZ_ENGINE_TO_TEST=ogre
39:  GZ_ENGINE_BACKEND=gl3plus
39: Test timeout computed to be: 240
39: Running main() from /home/jenkins/workspace/gz_rendering-ci-pr_any-noble-amd64/gz-rendering/test/gtest_vendor/src/gtest_main.cc
39: [==========] Running 1 test from 1 test suite.
39: [----------] Global test environment set-up.
39: [----------] 1 test from LoadUnloadTest
39: [ RUN      ] LoadUnloadTest.Thread
 39/132 Test  #39: INTEGRATION_load_unload_ogre_gl3plus .................***Exception: SegFault  0.63 sec
```

Blast545:
:farmer: This is also affecting main noble jobs, see:
https://build.osrfoundation.org/job/gz_rendering-ci-main-noble-amd64/6/

iche033:
I was able to reproduce the issue on Ubuntu Noble with ogre 1.9.0 debs but I haven't found a fix for it yet. Some observations:
* Test is designed to verify rendering works in a separate thread. Crash happens when the rendering (ogre) thread exists. No errors when deleting and cleaning up gz-rendering / ogre resources.
* `gdb` did not give any useful backtraces.
* Commenting out the call to [delete ogre root](https://github.com/gazebosim/gz-rendering/blob/a7cf4f0dba349f9f6fa9a70a8f082862e387f2e7/ogre/src/OgreRenderEngine.cc#L116) made the test pass but that's not ideal.
* When testing with ogre 1.9.0 built from source (with the same cmake args as ones used to build the debs), the test passes.

iche033:
> When testing with ogre 1.9.0 built from source (with the same cmake args as ones used to build the debs), the test passes.

[...]

Issue #745, opened 2022-10-18
:farmer: `SEH Exceptions` in gz-rendering-win


Crola1702:
## Environment
* OS Version:
  * Windows
* Source build
  * gz-rendering7, gz-rendering8, gz-rendering9, gz-rendering-main

## Description
* Expected behavior: All test passing
* Actual behavior: 71 tests failing

## Steps to reproduce
1. Run a build on [gz-rendering7-win](https://build.osrfoundation.org/job/ign_rendering-gz-7-win/), [gz-rendering8-win](https://build.osrfoundation.org/job/ign_rendering-gz-8-win/) or [ign_rendering-ci-win](https://build.osrfoundation.org/job/ign_rendering-ci-win/)
2. See it fail

## Output

We have 71 test regressions in gz-rendering windows (7, 8, 9 and main)

First time we saw it: [Sept 1](https://build.osrfoundation.org/job/ign_rendering-gz-7-win/4/)
It stopped between [Sept 28 and Oct 11](https://build.osrfoundation.org/job/ign_rendering-gz-7-win/12/)

And now it [came back](https://build.osrfoundation.org/job/ign_rendering-gz-7-win/14/)

We think it's probably related to this issue: https://stackoverflow.com/questions/13157671/seh-exception-with-code-0xc0000005-thrown-in-the-test-body

---

#### Update (17/08/2023):

Reference build: https://build.osrfoundation.org/job/ign_rendering-ci-win/254/#showFailuresLink

<details>
<summary>
Test regressions (71 last updated in 13-09-2024)
</summary>

* [CameraTest.ViewProjectionMatrix](https://build.osrfoundation.org/job/gz_rendering-9-win/9/testReport/junit/(root)/CameraTest/ViewProjectionMatrix/)
* [CameraTest.RenderTexture](https://build.osrfoundation.org/job/gz_rendering-9-win/9/testReport/junit/(root)/CameraTest/RenderTexture/)
* [CameraTest.TrackFollow](https://build.osrfoundation.org/job/gz_rendering-9-win/9/testReport/junit/(root)/CameraTest/TrackFollow/)
* [CameraTest.AddRemoveRenderPass](https://build.osrfoundation.org/job/gz_rendering-9-win/9/testReport/junit/(root)/CameraTest/AddRemoveRenderPass/)
[...]

PR #1033, opened 2024-08-09
Workaround for ogre crash on shutdown

iche033:
# 🦟 Bug fix

Workaround for https://github.com/gazebosim/gz-rendering/issues/1007

## Summary

The `INTEGRATION_load_unload` test crashes with ogre 1.x. The tests checks to see if the render engine can be loaded and unloaded in a thread (which is how `gz sim` runs the render engine). The crash happens on ubuntu 24.04 with the system debs, see https://github.com/gazebosim/gz-rendering/issues/1007 for more info.

The workaround implemented in this PR is to prevent the `RenderSystem_GL` plugin library from being unloaded on shutdown to avoid the crash. The plugin will be still be stopped and deleted but we do not call `dlclose` on it. I verified that the destructors of various objects / singletons in the plugin are called.

This is not ideal and we should find a proper solution in the future. 

## To test

The `INTEGRATION_load_unload` test should now pass with ogre 1.x on Ubuntu 24.04 with the system ogre debs:

```
GZ_ENGINE_TO_TEST=ogre  ./build/gz-rendering9/bin/INTEGRATION_load_unload
```

gz-sim should no longer crash on exit:

```
gz sim -v 4 -r --iterations 5 -s sensors_demo.sdf --render-engine ogre
```

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre/src/OgreRenderEngine.cc
+++ b/ogre/src/OgreRenderEngine.cc
@@ -180,13 +180,20 @@ void OgreRenderEngine::Destroy()
     {
       // TODO(anyone): do we need to catch segfault on delete?
 
-      // manually shutdown render system and stop the RenderSystem_GL library
+      // Manually shutdown ogre and stop the RenderSystem_GL library
+      // without unloading it.
+      // \todo(iche033) replace this whole block with:
+      //   delete this->ogreRoot
+      // once we are able to unload the RenderSystem_GL library without
+      // crashing.
       this->ogreRoot->shutdown();
-      this->ogreRoot->getRenderSystem()->shutdown();
+      this->ogreRoot->destroySceneManager(
+          this->ogreRoot->_getCurrentSceneManager());
+      Ogre::TextureManager::getSingletonPtr()->unloadAll();
+      Ogre::ShadowTextureManager::getSingletonPtr()->clear();
       this->ogreRoot->setRenderSystem(nullptr);
       for (auto &lib : mPluginLibs)
         unloadPlugin(lib);
-
       delete this->ogreRoot;
     }
     catch (...)
--- a/ogre/src/OgreRenderEngine.cc
+++ b/ogre/src/OgreRenderEngine.cc
@@ -81,12 +81,12 @@ using namespace rendering;
 // \todo(iche033) Find a proper way to unload the library without causing a
 // crash.
 typedef void (*DLL_START_PLUGIN)(void);
-DYNLIB_HANDLE glPluginHandle;
+void *glPluginHandle;
 
 void loadGLPlugin(const std::string &_pluginName)
 {
   glPluginHandle = dlopen(_pluginName.c_str(), RTLD_LAZY | RTLD_LOCAL);
-  DLL_START_PLUGIN pFunc = (DLL_START_PLUGIN)DYNLIB_GETSYM(
+  DLL_START_PLUGIN pFunc = (DLL_START_PLUGIN)dlsym(
       glPluginHandle, "dllStartPlugin");
   pFunc();
   return;
--- a/ogre/src/OgreRenderEngine.cc
+++ b/ogre/src/OgreRenderEngine.cc
@@ -70,83 +70,27 @@ class gz::rendering::OgreRenderEnginePrivate
 using namespace gz;
 using namespace rendering;
 
-// Plugin loading and unloading code adapted from Ogre.
 // Unloading the RenderSystem_GL plugin was found to cause a crash when the
 // rendering thread exits. This only happens on Ubuntu 24.04 when running
 // ogre using system debs, see
 // https://github.com/gazebosim/gz-rendering/issues/1007
-// To workaournd ths issue, we adapth the code for loadng and unloading
-// ogre plugin here. We load the RenderSystem_GL library manually and store it
-// in gz-rendering and on shutdown we stop the plugin and delete the lib without
-// unloading it.
+// To workaround ths issue, we adapt the code for loading the ogre plugin here.
+// We load the RenderSystem_GL library manually and store it in gz-rendering.
+// On shutdown, ogre root will uninstall this plugin and delete the GL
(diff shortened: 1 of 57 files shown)


scpeters:
potential fix for the remaining cmake warning in https://github.com/gazebo-tooling/release-tools/pull/1161

scpeters:
> potential fix for the remaining cmake warning in [gazebo-tooling/release-tools#1161](https://github.com/gazebo-tooling/release-tools/pull/1161)

testing shows that https://github.com/gazebo-tooling/release-tools/pull/1161 should fix the warning

ogre/src/OgreRenderEngine.cc
```diff
@@ -66,6 +70,84 @@ class gz::rendering::OgreRenderEnginePrivate
 using namespace gz;
 using namespace rendering;
 
+// Plugin loading and unloading code adapted from Ogre.
+// Unloading the RenderSystem_GL plugin was found to cause a crash when the
+// rendering thread exits. This only happens on Ubuntu 24.04 when running
+// ogre using system debs, see
+// https://github.com/gazebosim/gz-rendering/issues/1007
+// To workaournd ths issue, we adapth the code for loadng and unloading
```
scpeters, line 78:
```suggestion
// To workaournd ths issue, we adapt the code for loadng and unloading
```

scpeters (commented):
minor nit; I'm still trying to understand the code

azeey:
We now have API to load plugins with `RTLD_NODELETE`. This would ensure that the plugin is not actually deleted even if `dlclose` is called. Would changing https://github.com/gazebosim/gz-rendering/blob/6a2217fce93cbc9c5db6495c711e6cd9b1c7991d/src/RenderEngineManager.cc#L498 to  `LoadLib(pathToLib, true)` fix this issue without any of the other changes?

ogre/src/OgreRenderEngine.cc
```diff
@@ -118,9 +200,13 @@ void OgreRenderEngine::Destroy()
 
   if (ogreRoot)
   {
+    // TODO(anyone): do we need to catch segfault on delete?
     try
     {
-      // TODO(anyone): do we need to catch segfault on delete?
+      // Install a dummy plugin that allows use to manually shutdown ogre
```
scpeters, line 206:
```suggestion
      // Install a dummy plugin that allows us to manually shutdown ogre
```

ogre/src/OgreRenderEngine.cc
```diff
@@ -66,6 +70,84 @@ class gz::rendering::OgreRenderEnginePrivate
...
+// rendering thread exits. This only happens on Ubuntu 24.04 when running
+// ogre using system debs, see
+// https://github.com/gazebosim/gz-rendering/issues/1007
+// To workaournd ths issue, we adapth the code for loadng and unloading
+// ogre plugin here. We load the RenderSystem_GL library manually and store it
+// in gz-rendering and on shutdown we stop the plugin and delete the lib without
+// unloading it.
+// \todo(iche033) Find a proper way to unload the library without causing a
+// crash.
+typedef void (*DLL_START_PLUGIN)(void);
```
scpeters, line 84: `DLL_START_PLUGIN` has a `const NameValuePairList *` argument in [OgreRoot.cpp:132](https://github.com/OGRECave/ogre-next/blob/v2.3.3/OgreMain/src/OgreRoot.cpp#L132), but is `void` here. Does that matter?

ogre/src/OgreRenderEngine.cc
```diff
@@ -66,6 +70,84 @@ class gz::rendering::OgreRenderEnginePrivate
...
+  // on thread exit so commented out for now.
+  // _pluginLib->unload();
+  delete _pluginLib;
+}
+
+/// A dummy plugin class that is installed to the ogre root before shutdown.
+/// We use this as a hook to get a callback from ogre root when all plugins
+/// are being unloaded. This is so that we can unload the RenderSystem_GL
+/// at the right time.
+class DummyPlugin : public Ogre::Plugin
```
scpeters, line 131: nit: rename to `DummyUnloadPlugin`

scpeters:
> We now have API to load plugins with `RTLD_NODELETE`. This would ensure that the plugin is not actually deleted even if `dlclose` is called. Would changing
> 
> https://github.com/gazebosim/gz-rendering/blob/6a2217fce93cbc9c5db6495c711e6cd9b1c7991d/src/RenderEngineManager.cc#L498
> 
> to `LoadLib(pathToLib, true)` fix this issue without any of the other changes?

testing in a draft PR: https://github.com/gazebosim/gz-rendering/pull/1038

scpeters:
> > We now have API to load plugins with `RTLD_NODELETE`. This would ensure that the plugin is not actually deleted even if `dlclose` is called. Would changing
> > https://github.com/gazebosim/gz-rendering/blob/6a2217fce93cbc9c5db6495c711e6cd9b1c7991d/src/RenderEngineManager.cc#L498
> > 
> > to `LoadLib(pathToLib, true)` fix this issue without any of the other changes?
> 
> testing in a draft PR: #1038

the noble workflow passes; I've marked https://github.com/gazebosim/gz-rendering/pull/1038 as ready for review

ogre/src/OgreRenderEngine.cc
scpeters: Oops thanks!

scpeters (approved):
this has fixed CI and looks fine to me; I defer to @azeey about whether to follow the suggestion from @iche033 to use this instead of #1038

azeey (approved):
I was going to check if windows CI was okay, but it looks like the Ogre1 tests have been failing for a while. I thought it was just Ogre2 that was failing :(. Anyway, I'm good with this approach.

iche033:
yeah we should try setting up ogre 2.3 on windows, https://github.com/gazebosim/gz-rendering/issues/745

PR #1068, opened 2024-09-18
Update code for loading normal and albedo maps from image data

iche033:
# 🦟 Bug fix

## Summary
Currently when we load a normal map from an image data (e.g. from a glb file that has an embedded normal map texture), we upload the image data as RGBA8 with 1 mip map level. This is different from what ogre's texture manager does when it loads a normal map from file. To make the process consistent, we follow the same steps: convert to an RG signed 8 bit format then generate multiple levels of mip maps. I found that the mip map generation process made a difference to the final appearance. It helps to make the visual look smoother when viewed from a distance.

Before: The visuals appears to have some artifacts / noise, e.g more evident on the cushions behind the tables

ionic_demo_normal_map_before

After: visuals appear smoother

ionic_demo_normal_map

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/src/Ogre2Material.cc
+++ b/ogre2/src/Ogre2Material.cc
@@ -1274,7 +1274,7 @@ void Ogre2Material::SetTextureMapDataImpl(const std::string& _name,
     // \todo(iche033) See if there is a way to reuse these functions
     // from ogre-next without copying the code here.
 
-    // Step 1: convert to two component signed 8 bit format:
+    // Normal maps - convert to two component signed 8 bit format:
     // Ogre::PFG_RG8_SNORM format
     if (_type == Ogre::PBSM_NORMAL)
       this->dataPtr->PrepareForNormalMapping(texture, img);


ogre2/src/Ogre2Material.cc
```diff
@@ -1558,3 +1576,102 @@ ShaderParamsPtr Ogre2Material::FragmentShaderParams()
...
   return this->dataPtr->fragmentShaderParams;
 }
+
+//////////////////////////////////////////////////
+void Ogre2MaterialPrivate::PrepareAndUploadNormalMap(Ogre::TextureGpu *_texture,
+    Ogre::Image2 &_image)
+{
+  // The code below is adapted from ogre-next v2-3. It replicates the steps that
+  // ogre does when it loads a normal map from file.
+  // Here we adapte the code to load a normal map texture from an image in
```
ahcorde, line 1586:
```suggestion
  // Here we adapt the code to load a normal map texture from an image in
```

ogre2/src/Ogre2Material.cc
```diff
@@ -1558,3 +1576,102 @@ ShaderParamsPtr Ogre2Material::FragmentShaderParams()
...
+    Ogre::PixelFormatGpuUtils::convertForNormalMapping(
+        srcBox, _image.getPixelFormat(),
+        dstBox, dstFormat);
+  }
+  _image.loadDynamicImage(imgData, _image.getWidth(), _image.getHeight(),
+      _image.getDepthOrSlices(), _image.getTextureType(), dstFormat, false,
+      numMipmaps);
+  _texture->setPixelFormat(dstFormat);
+
+  // _image.save(_name + ".png", 0, 1);
```
ahcorde, line 1633: remove ?

ogre2/src/Ogre2Material.cc
```diff
@@ -1558,3 +1576,102 @@ ShaderParamsPtr Ogre2Material::FragmentShaderParams()
...
+      Ogre::TextureFlags::DiscardableContent,
+      _texture->getTextureType());
+  tempTexture->copyParametersFrom(_texture);
+  tempTexture->unsafeScheduleTransitionTo(Ogre::GpuResidency::Resident);
+  Ogre::TextureBox box = _texture->getEmptyBox(0);
+  _texture->copyTo(tempTexture, box, 0, box, 0);
+  tempTexture->_autogenerateMipmaps();
+
+  Ogre::uint8 mipmaps = _texture->getNumMipmaps();
+  for (size_t i = 1u; i < mipmaps; ++i)
```
ahcorde, line 1670: include `<cstddef>`

ahcorde (changes requested):

azeey:
I just also noticed that the some of the objects in the scene appear to have a metallic look? Or at least a black band that I don't think should be there.
image

For reference, with all lights turned off and using https://github.com/gazebosim/gz-rendering/pull/1067
image


And for additional reference, here's the render using Blender Cycles 
cycles_reference

iche033:
> This is certainly better, but it looks like there still some differences

While testing with ogre built from source, I noticed that it also generates mip maps for the albedo textures (and maybe others) which we're not doing. It could contribute to the differences, and there could also be others.

iche033:
> While testing with ogre built from source, I noticed that it also generates mip maps for the albedo textures (and maybe others) which we're not doing. It could contribute to the differences, and there could also be others.

Updated to also generate mip maps for albedo textures. It made subtle difference. Surfaces appear slightly smoother

ionic_demo_normal_and_albedo

ogre2/src/Ogre2Material.cc
iche033: [615dfd5](https://github.com/gazebosim/gz-rendering/pull/1068/commits/615dfd5f52551720bdc42970a63330be43066a9f)

ogre2/src/Ogre2Material.cc
iche033: [615dfd5](https://github.com/gazebosim/gz-rendering/pull/1068/commits/615dfd5f52551720bdc42970a63330be43066a9f)

ogre2/src/Ogre2Material.cc
iche033: removed. [615dfd5](https://github.com/gazebosim/gz-rendering/pull/1068/commits/615dfd5f52551720bdc42970a63330be43066a9f)

iche033:
> I just also noticed that the some of the objects in the scene appear to have a metallic look? Or at least a black band that I don't think should be there.

For that particular black band, I think at least partially it's because GI is low res and not working well. If I turn off GI, that part of the ceiling is dark since it's not lit by the spot lights. With GI enabled, it lights up part of that surface but not all, leaving the black band. If I change GI resolution to [64, 64, 64], that band is gone. However, there are still small artifacts in other places.

iche033:
I think the appearance is now closer to the result from loading textures files. Can we get this in first?

azeey (approved):

ahcorde (approved):

PR #324, opened 2021-05-18
Destroy material when a mesh is deleted

ahcorde:
# 🎉 Enhancement 

## Summary

This PR allows to remove materials when a mesh is removed.

There are two cases:
 - When the material is loaded, the method in charge to load the meshes will create a cache of material. I added a method to remove this cache
 - When a mesh is removed the material/s are not removed from the memory. This PR also includes this step.

I had to break the ABI, I'm open to suggestions to be able to backport this PR

## Checklist
- [x] Signed all commits for DCO
- [ ] Added tests
- [ ] Added example and/or tutorial
- [ ] Updated documentation (as needed)
- [ ] Updated migration guide (as needed)
- [x] `codecheck` passed (See [contributing](https://ignitionrobotics.org/docs/all/contributing#contributing-code))
- [x] All tests passed (See [test coverage](https://ignitionrobotics.org/docs/all/contributing#test-coverage))
- [ ] While waiting for a review on your PR, please help review [another open pull request](https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aignitionrobotics+repo%3Aosrf%2Fsdformat+archived%3Afalse+) to support the maintainers

**Note to maintainers**: Remember to use **Squash-Merge**

PR #1110, opened 2025-02-27
Remove ClearMaterialsCache call when a material is destroyed in ogre2

iche033:
# 🦟 Bug fix

## Summary

The material cache in `Ogre2MeshFactory` was introduced in #324 to clear material objects when a mesh is removed. Looking at the current state, this call should no longer be needed as we now have logic in `Ogre2Material::Destroy` function to destroy ogre internal material / texture objects (and also unregistering gz-rendering material objects) when they [are not in use](https://github.com/gazebosim/gz-rendering/blob/a48f757d9448e05bdc3e6e73c61ddf4869febac6/ogre2/src/Ogre2Material.cc#L254-L268). So this PR removes the material cache and the the call to `ClearMaterialCache` function.

Note that we ran into an issue with the `ClearMaterialsCache` function. It clears materials by the name of the texture it contains. However, it's possible for multiple materials to use the same texture, and the current logic may incorrectly destroy the wrong material. Removing the call to `ClearMaterialsCache` resolves that issue.

## Checklist
- [ ] Added tests

### DIFF
--- a/ogre2/include/gz/rendering/ogre2/Ogre2MeshFactory.hh
+++ b/ogre2/include/gz/rendering/ogre2/Ogre2MeshFactory.hh
@@ -95,6 +95,7 @@ namespace gz
       protected: Ogre2ScenePtr scene;
 
       /// \brief Remove internal material cache for a specific material
+      /// \todo(iche033) Deprecate this function in gz-rendering10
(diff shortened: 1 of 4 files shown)


ahcorde (approved):

iche033:
@mergifyio backport gz-rendering8

PR #1131, opened 2025-05-01
Bump gz-cmake and others in jetty

scpeters:
Part of https://github.com/gazebo-tooling/release-tools/issues/1309.


.github/workflows/windows.yaml
iche033, line 1: I think we need to update line 34 below to `${env:PACKAGE}10.yaml` so it can download the right file

scpeters: I renamed the variable to `PACKAGE_WITH_VERSION` and restored the number, then separately computed a `PACKAGE` variable by removing digits in https://github.com/gazebosim/gz-rendering/pull/1131/commits/e31ef2a0642b56d690a0e1331c7dc5f2827a97b2

scpeters: the windows workflow doesn't check for `ci_matching_branch/`, so I think it's now failing due to that

iche033 (approved):

.github/workflows/windows.yaml
scpeters: I reported this in https://github.com/gazebosim/gz-rendering/issues/1132 and will try merging the gazebodistro PR and rerunning the windows workflow

scpeters:
windows github workflow is fixed

two jenkins windows test failures are preexisting, so I'll go ahead and merge

* https://build.osrfoundation.org/job/gz_rendering-main-cnlwin/6/testReport/

PR #1138, opened 2025-05-11
Fix user settings of thermal camera being ignored

### LINKED ISSUE #1134: Thermal camera does not react to updated settings once thermal texture has been created
## Environment
* OS Version: Ubuntu 22.04
* Binary and source build 8.2.2

## Description
* Expected behavior: Changing the thermal camera resolution to very low values through a `.sdf` file will clearly influence the output
* Actual behavior: Thermal camera output is unaffected, no matter what resolution is set through the `.sdf` file

## Steps to reproduce

1. Open `examples/worlds/thermal_camera.sdf` in a text editor
2. Change the resolution of the 8-bit camera through the `gz-sim-thermal-sensor-system` plugin to `0.1`
3. Run `gz sim -r examples/worlds/thermal_camera.sdf`
4. Notice that the output of the 8-bit camera is unchanged

## Debugging efforts

I have noticed this when trying to use the `gz-sim-thermal-sensor-system` plugin to change the properties of thermal cameras, but I believe the root cause is in the rendering code itself. That is unless it is actually intended behaviour that the thermal camera does not react to setting changes until it is reset in some way which I was unable to find.

By tracing the calls for setting the thermal camera resolution through `gz-sim`, `gz-sensors` and `gz-rendering` I have seen that the callchain does indeed reach `BaseThermalCamera<T>::SetLinearResolution`. However, as far as I can tell `CreateThermalTexture` is only ever called once for the lifetime of the camera. That means that `this->resolution` is only read once and set for the shader: https://github.com/gazebosim/gz-rendering/blob/99031b74bd8bf63116eeaa8585025f0fe6b3edd3/ogre2/src/Ogre2ThermalCamera.cc#L887-L888
Any further calls to `BaseThermalCamera<T>::SetLinearResolution`, for example through the `gz-sim-thermal-sensor-system` plugin only set `this->resolution`, but never the constant in the shader, meaning that the thermal camera output remains unaffected. For Ogre2, the new resolution is also never reported to the `OgreThermalCameraMaterialSwitcher` that uses it for creating normalized materials. 

Note that `min_temp` and `max_temp` of the `gz-sim-thermal-sensor-system` plugin are also affected by this. It is not limited to use with this plugin though. Changing any of the settings that are only read once in `CreateThermalTexture` after the first call to `thermalCamera->Update()` in the thermal camera integration tests will not affect the result of any future calls to `thermalCamera->Update()` either.

As a simple workaround, I tried adding something like
```c++
  Ogre::Pass *pass =
    this->dataPtr->thermalMaterial->getTechnique(0)->getPass(0);
  Ogre::GpuProgramParametersSharedPtr psParams =
    pass->getFragmentProgramParameters();
  psParams->setNamedConstant("max",
      static_cast<float>(this->maxTemp));
  psParams->setNamedConstant("min",
      static_cast<float>(this->minTemp));
  psParams->setNamedConstant("resolution",
      static_cast<float>(this->resolution));

  this->dataPtr->thermalMaterialSwitcher->SetLinearResolution(this->resolution);
```
to the end of `Ogre2ThermalCamera::PreRender`, which then indeed allows me to successfuly change the settings through the `gz-sim-thermal-sensor-system` plugin. However, I am very unsure about whether `PreRender` is the correct place for this, perhaps overriding the setter functions to also update the shader constants would be better? And I'm not sure whether other settings like AmbientTemperature should also be covered by this or if it is intended to always be fixed.

But I do hope that I'm not wrong in expecting that at least `resolution`, `maxTemp` and `minTemp` are supposed to cause direct change in the output, otherwise I apologize. In that case I suppose I should file a bug with `gz-sim-thermal-sensor-system` instead to add some kind of call that resets the thermal camera whenever it updates its settings?

mjacobse:
# 🦟 Bug fix

Fixes #1134

## Summary
Changes to thermal camera settings that can be made through the plugin gz-sim-thermal-sensor-system would not show up in the result. While the value of the data members was updated, the named constants in the material shader used for rendering were not. We fix that by always setting the shader constants in `PreRender` to the current values.

This might be relevant for other thermal camera settings as well, but this focuses on min and max temperature and resolution, as those can be modified through the gz-sim-thermal-sensor-system plugin.

Unfortunately this only seems to work for Ogre2 right now, updating the shader constants does not seem to work for Ogre1. Keeping this as draft in hope to find a fix for Ogre1 before merging.

## Checklist
- [x] Added tests

### DIFF
--- a/ogre/src/OgreThermalCamera.cc
+++ b/ogre/src/OgreThermalCamera.cc
@@ -527,6 +527,15 @@ void OgreThermalCamera::PreRender()
   BaseCamera::PreRender();
   if (!this->dataPtr->ogreThermalTexture)
     this->CreateThermalTexture();
+
+  // ensure that certain shader constants are up-to-date so that changes that
+  // users can make to the settings show up in the thermal result immediately
+  Ogre::Pass *pass =
+      this->dataPtr->thermalMaterial->getTechnique(0)->getPass(0);
+  auto params = pass->getFragmentProgramParameters();
+  params->setNamedConstant("max", this->maxTemp);
+  params->setNamedConstant("min", this->minTemp);
+  params->setNamedConstant("resolution", this->resolution);
 }
 
 //////////////////////////////////////////////////
--- a/ogre2/src/Ogre2ThermalCamera.cc
+++ b/ogre2/src/Ogre2ThermalCamera.cc
@@ -1120,6 +1120,21 @@ void Ogre2ThermalCamera::PreRender()
   if (!this->dataPtr->ogreThermalTexture)
     this->CreateThermalTexture();
 
+  // ensure that certain shader constants are up-to-date so that changes that
+  // users can make to the settings show up in the thermal result immediately
+  Ogre::Pass *pass =
+    this->dataPtr->thermalMaterial->getTechnique(0)->getPass(0);
+  Ogre::GpuProgramParametersSharedPtr psParams =
+    pass->getFragmentProgramParameters();
+  psParams->setNamedConstant("max",
+      static_cast<float>(this->maxTemp));
+  psParams->setNamedConstant("min",
+      static_cast<float>(this->minTemp));
+  psParams->setNamedConstant("resolution",
+      static_cast<float>(this->resolution));
+
+  this->dataPtr->thermalMaterialSwitcher->SetLinearResolution(this->resolution);
+
   // todo(iche033) Override BaseCamera::SetProjectionMatrix() function in
   // main / gz-rendering9 instead of checking and setting the custom
   // projection matrix here
--- a/test/integration/thermal_camera.cc
+++ b/test/integration/thermal_camera.cc
@@ -706,6 +706,71 @@ TEST_F(ThermalCameraTest, ThermalCameraParticles)
 }
 
 //////////////////////////////////////////////////
+TEST_F(ThermalCameraTest, ThermalCameraResolutionUpdate)
+{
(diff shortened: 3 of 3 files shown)


iche033 (commented):
ogre 1.x's thermal camera implementation is quite behind ogre2's. I just ticketed #1140. You can skip updating ogre1.x for now. For the test, can you add the [CHECK_SUPPORTED_ENGINE("ogre2")](https://github.com/gazebosim/gz-rendering/blob/2fa760be4026f86c5e988243a5051e4bd30cf38b/test/integration/thermal_camera.cc#L385C3-L385C34) macro to the top of the test cases so that they only run the test for `ogre2`?

mjacobse:
> ogre 1.x's thermal camera implementation is quite behind ogre2's. I just ticketed #1140. You can skip updating ogre1.x for now. For the test, can you add the [CHECK_SUPPORTED_ENGINE("ogre2")](https://github.com/gazebosim/gz-rendering/blob/2fa760be4026f86c5e988243a5051e4bd30cf38b/test/integration/thermal_camera.cc#L385C3-L385C34) macro to the top of the test cases so that they only run the test for `ogre2`?

ok thanks, done as suggested

iche033 (approved):

iche033:
looks good to me!

mjacobse:
Should this be backported to gz-rendering9 and main too?

iche033:
@mergifyio backport gz-rendering9

PR #1160, opened 2025-08-25
Update references of main to gz-rendering10

iche033:
Part of https://github.com/gazebosim/gz-jetty/issues/30


.github/workflows/bazel.yml
```diff
@@ -1,9 +1,14 @@
...
   push:
-    branches: [gz-rendering9, main]
+    branches:
+      - 'gz-rendering[1-9]?[0-9]'
+      - 'main'
   pull_request:
-    branches: [gz-rendering9, main]
+    branches:
+      - 'gz-rendering[1-9]?[0-9]'
+      - 'main'
```
azeey, line 10: I don't think we need the `branches` argument for the `pull_request` parameters because we want the job to run on every pull request. See https://github.com/gazebosim/gz-rendering/blob/23cdd5a0492a7df73c5c363e6da39ab9fe36ca19/.github/workflows/ci.yml#L3-L9

azeey (approved):

.github/workflows/bazel.yml
iche033: also updated gz-sensors: https://github.com/gazebosim/gz-sensors/pull/550

PR #1303, opened 2026-06-22
Persistent GPU->CPU readback to avoid duplicate copies

j-rivero:
# 🎉 New feature

## Summary

`Ogre2DepthCamera::PostRender()` and `Ogre2GpuRays::PostRender()` read the GPU render target back to CPU memory every frame through `Ogre::Image2::convertFromTexture`. That convenience call hides a wasteful per-frame pattern: it allocates a fresh ~4.9 MB CPU buffer, spins up a throwaway `AsyncTextureTicket`, does the blocking GPU→CPU download, `memcpy`s the staging buffer into the `Image2`, then frees the ticket — after which gz **re-copies** the whole buffer into its persistent member buffer, and the depth camera makes a **third** per-pixel pass to extract channel 0.

This PR introduces a small reusable helper, **`Ogre2GpuReadbackTicket`**, that keeps **one persistent `AsyncTextureTicket`** alive across frames, maps its staging memory, and lets the caller copy **once** in a single stride-aware pass (fusing the depth camera's channel-0 extraction into that same copy). Per camera per frame this goes from *(1 malloc/free + 2 full memcpys + 1 extraction pass)* to *(0 allocations + 1 fused pass)*.

### Interactive rich explanation

I've asked the AI agents to create an easy-to-understand explanation of the changes, I think worth it to give it a look https://j-rivero.github.io/gz-rendering-pipeline/gz_readback_dedup.html

## Rollout / opt-out

The new path is **on by default**. Setting `GZ_RENDERING_OGRE2_LEGACY_READBACK=1` restores the original `convertFromTexture` path verbatim — kept in-tree as a faithful A/B control and a safety hatch. (It can be removed in a later cleanup once the new path has soaked.)

## Performance

Measured on an RTX 4060 Ti / 16-core box, 640×480 RGBD cameras, `new` (default) vs `legacy` (`GZ_RENDERING_OGRE2_LEGACY_READBACK=1`). Single A/B run each — direction is consistent; treat the exact percentages as indicative.

**Default backend (OpenGL / GL3Plus — the gz-rendering ogre2 default):**

| Scenario | Legacy | New | Δ |
|---|---|---|---|
| 16 cams, saturated, headless | 11.55 Hz/cam | **12.73 Hz/cam** | **+10.2% throughput** |
| 8 cams, 15 Hz cap, GUI | 15.15 Hz/cam @ 148% core | 15.15 Hz/cam @ **140% core** | same FPS, **~5% less server CPU** |

**Both backends (8 cams, uncapped, GUI; active `RenderSystem` confirmed in the log):**

| Backend | Legacy | New | Δ (the fix) |
|---|---|---|---|
| OpenGL (GL3Plus) | 23.59 Hz/cam | **26.94 Hz/cam** | **+14.2%** |
| Vulkan | 26.02 Hz/cam | **27.26 Hz/cam** | **+4.8%** |

The optimization helps on **both** backends — it operates through OGRE-Next's `AsyncTextureTicket`/`convertFromTexture` abstraction over the active render system. Vulkan's *legacy* staging path is already ~10% faster than OpenGL's, so the redundant-copy removal has less headroom to recover there (smaller Δ); after the fix both backends converge to ~27 Hz/cam. Across all runs serverCPU stays ~190–200% and GPU ~33%, confirming the bottleneck removed is CPU-side memory bandwidth, not the GPU. When frame rate is capped the same saving appears as lower CPU for identical output instead of higher throughput.

## Follow-ups (out of scope)

Extend the helper to the other readback sites that still use `convertFromTexture`  (thermal, segmentation, bounding-box, wide-angle, selection buffer) and the regular camera's `copyContentsToMemory`.

## Test it

By manual look into gz-sim:

https://github.com/user-attachments/assets/6d3dc406-e572-49e8-806c-937af80e2d04

By an agent instructed to do the checks:

- **Byte-identical output** verified via header-stripped data-md5 oracles on a static
  scene, for **both** the new and legacy paths:
  - depth: image / depth / point-cloud md5s match the unmodified baseline;
  - gpu rays: scan md5 matches the unmodified baseline.
  - Oracle sensitivity confirmed (a deliberately perturbed copy → MISMATCH).
  - **Verified on both backends.** OpenGL: new/legacy match the baseline. Vulkan:  new and legacy match each other on depth/points/image (internal A/B). The Vulkan capture's depth/points differ from the OpenGL render — proving the Vulkan backend  was genuinely active, not a silent GL fallback — while new==legacy still holds within the backend.
- `INTEGRATION_depth_camera` (ogre2): 4/4 pass.
- `INTEGRATION_gpu_rays_ogre2`: 2/2 pass.
- **No leak:** RSS over a 25 s soak shows 0 kB growth (ticket is reused, not reallocated).

## Backport Policy

Changes seems to be API/ABI compatible to me so I don't see any reason not to backport the improvement to other releases. If we consider it risky, the feature could be shipped OFF by default.

## Checklist
- [ ] Added tests

Generated-by: Opus 4.8

### DIFF
--- a/ogre2/src/Ogre2DepthCamera.cc
+++ b/ogre2/src/Ogre2DepthCamera.cc
@@ -1204,7 +1204,7 @@ void Ogre2DepthCamera::PostRender()
   if (!this->dataPtr->depthBuffer)
     this->dataPtr->depthBuffer = new float[len * channelCount];
   if (!this->dataPtr->depthImage)
-    this->dataPtr->depthImage = new float[len];
+    this->dataPtr->depthImage = new float[len * channelCount];
 
   if (Ogre2UseLegacyReadback())
   {
--- a/ogre2/src/Ogre2DepthCamera.cc
+++ b/ogre2/src/Ogre2DepthCamera.cc
@@ -41,6 +41,7 @@
 #include "gz/rendering/ogre2/Ogre2Scene.hh"
 #include "gz/rendering/ogre2/Ogre2Sensor.hh"
 
+#include "Ogre2GpuReadbackTicket.hh"
 #include "Ogre2ParticleNoiseListener.hh"
 
 #ifdef _MSC_VER
@@ -100,6 +101,10 @@ class gz::rendering::Ogre2DepthCameraPrivate
 
   /// \brief Outgoing depth data, used by newDepthFrame event.
   public: float *depthImage = nullptr;
+
+  /// \brief Persistent GPU->CPU readback ticket used by the non-legacy
+  /// PostRender path.
+  public: Ogre2GpuReadbackTicket depthReadback;
 
   /// \brief maximum value used for data outside sensor range
   public: float dataMaxVal = gz::math::INF_D;
@@ -304,6 +309,7 @@ void Ogre2DepthCamera::Init()
 void Ogre2DepthCamera::Destroy()
 {
   this->RemoveAllRenderPasses();
+  this->dataPtr->depthReadback.Destroy();
 
   if (this->dataPtr->depthBuffer)
   {
@@ -1158,6 +1164,30 @@ void Ogre2DepthCamera::PreRender()
   this->dataPtr->renderPassDirty = false;
 }
 
+namespace
+{
+  /// \brief Copy one mapped RGBA32 readback box into the persistent depth
+  /// buffers in a single stride-aware pass: the full RGBA row into
+  /// _depthBuffer (point cloud + DepthData()) and channel 0 into _depthImage
+  /// (depth frame). Honors _box.bytesPerRow as the source stride.
+  void FillDepthBuffers(const Ogre::TextureBox &_box,
+      float *_depthBuffer, float *_depthImage,
+      unsigned int _width, unsigned int _height,
+      unsigned int _channelCount, unsigned int _bytesPerChannel)
+  {
+    const float *src = static_cast<const float *>(_box.data);
+    for (unsigned int i = 0; i < _height; ++i)
+    {
+      const unsigned int rawRowIdx = i * _box.bytesPerRow / _bytesPerChannel;
+      const unsigned int rowIdx = i * _width * _channelCount;
+      memcpy(&_depthBuffer[rowIdx], &src[rawRowIdx],
+          _width * _channelCount * _bytesPerChannel);
+      for (unsigned int j = 0; j < _width; ++j)
+        _depthImage[i * _width + j] = src[rawRowIdx + j * _channelCount];
+    }
+  }
+}  // namespace
+
 //////////////////////////////////////////////////
 void Ogre2DepthCamera::PostRender()
 {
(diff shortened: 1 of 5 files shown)


ogre2/src/Ogre2GpuReadbackTicket.hh
```diff
@@ -0,0 +1,98 @@
...
+      /// \brief Cached ticket pixel format; see width.
+      private: Ogre::PixelFormatGpu format{Ogre::PFG_UNKNOWN};
+
+      /// \brief True while the ticket is currently mapped.
+      private: bool mapped{false};
+    };
+
+    /// \brief Whether to use the legacy Ogre::Image2::convertFromTexture
+    /// readback path instead of the persistent-ticket path. True iff the env
+    /// var GZ_RENDERING_OGRE2_LEGACY_READBACK is set. Evaluated once.
```
iche033, line 92: My understanding is that this env var is temporary so we can fallback to legacy route in case we see behavior change. Can you add a todo note here reminding when to deprecate / remove this env variable? e.g. we can ship gz-rendering11 (gazebo M) with this env var available, deprecate in N, then remove in O? Alternatively, if this works well, just can just remove in N.

iche033 (approved):
looks good, just one minor comment

ogre2/src/Ogre2DepthCamera.cc
```diff
@@ -1171,40 +1201,59 @@ void Ogre2DepthCamera::PostRender()
...
-  Ogre::Image2 image;
-  image.convertFromTexture(this->dataPtr->ogreDepthTexture[1], 0u, 0u);
-  Ogre::TextureBox box = image.getData(0);
-  float *depthBufferTmp = static_cast<float *>(box.data);
   if (!this->dataPtr->depthBuffer)
-  {
     this->dataPtr->depthBuffer = new float[len * channelCount];
-  }
+  if (!this->dataPtr->depthImage)
+    this->dataPtr->depthImage = new float[len];
```
iche033: keep this consistent with `depthBuffer` allocation above? In the typical use case, `channelCount` should be always 1, but just in case we introduce a different format for depth images in the future

```suggestion
    this->dataPtr->depthImage = new float[len * channelCount];
```

ogre2/src/Ogre2GpuReadbackTicket.hh
j-rivero: [cf925c8](https://github.com/gazebosim/gz-rendering/pull/1303/commits/cf925c8fa3b38b2b268ba08905999c2ae6266cec)

j-rivero:
Include a note on migrations [c745d1e](https://github.com/gazebosim/gz-rendering/pull/1303/commits/c745d1ec6a9bb73bf11ef5ddf80a596d68ffa5e9)

ogre2/src/Ogre2GpuReadbackTicket.hh
```diff
@@ -0,0 +1,101 @@
...
+
+      /// \brief Cached ticket pixel format; see width.
+      private: Ogre::PixelFormatGpu format{Ogre::PFG_UNKNOWN};
+
+      /// \brief True while the ticket is currently mapped.
+      private: bool mapped{false};
+    };
+
+    /// \brief Whether to use the legacy Ogre::Image2::convertFromTexture
+    /// readback path instead of the persistent-ticket path. True iff the env
```
ahcorde:
```suggestion
    /// readback path instead of the persistent-ticket path. True if the env
```

ahcorde (approved):

PR #1309, opened 2026-07-08
Do not pre-render each visual's children twice per frame

caguero:
# 🦟 Bug fix

## Summary

`BaseVisual::PreRender()` pre-renders each visual's children twice every frame. It calls `T::PreRender()`, which resolves to `BaseNode::PreRender()` and already recurses into the children through the virtual `PreRenderChildren()`, and then calls `PreRenderChildren()` again explicitly. Because the duplication happens at every level of the scene graph, it compounds to O(2^depth) redundant traversals: the more deeply nested a scene's model → link → visual hierarchy, the worse it gets.

This patch removes the redundant call so each visual's children are pre-rendered exactly once. 

## Benchmark results

Measured on the Gazebo GUI process (`gz-sim-gui-client`, OgreNext + NVIDIA), matched `RelWithDebInfo` builds, before → after:

| World | FPS | Total GUI CPU | `BaseVisual::PreRenderChildren` (inclusive) |
|---|---|---|---|
| 3000 static shapes | 33 → 58 (**+73%**) | ~flat | 52% → 32% |
| 3000 moving shapes | 52 → 64 (**+24%**) | **−30% less CPU** | 53% → 35% |
| jetty (~320 models) | ~60 (vsync-capped) | **−15% less CPU** | 22% → 10% |

The saving shows up as a higher frame rate when the render thread is the bottleneck (the 3000-shape scene was render-bound at 33 FPS) and as lower CPU when the scene already reaches the vsync target (jetty). The relative reduction in `PreRenderChildren` grows with scene-graph depth, so deeply nested worlds (jetty) see the largest cut.

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist

- [ ] Added tests
- [ ] `codecheck` passed
- [ ] All tests passed

**Note to maintainers**: Remember to use **Squash-Merge** and edit the commit
message to match the pull request summary while retaining the `Signed-off-by`
lines.


iche033 (approved):

caguero:
@Mergifyio backport gz-rendering10 gz-rendering9 gz-rendering8

PR #1324, opened 2026-08-18
Fix Ogre2 CPU ray queries from inside AABBs

### LINKED ISSUE #1226: Ogre2 CPU implementation of RayQuery skips over geometry whose AABB contains the origin of the ray
## Environment
* OS Version: Ubuntu 22.04
* Source or binary build? Binary, from apt for my version of Ubuntu
    - Rendering plugin: ogre2.
      - [x] running in Docker/Singularity

Omitting some details since the bug is within the CPU code

## Description
This is specifically when using the CPU implementation of Ogre2RayQuery

In [Ogre2RayQuery](https://github.com/gazebosim/gz-rendering/blob/gz-rendering10/ogre2/src/Ogre2RayQuery.cc) , `ThreadedTriRay::execute` will skip over any object whose intersection distance is `<= 0`.  Unfortunately,  in [OgreMath.cpp](https://github.com/OGRECave/ogre/blob/master/OgreMain/src/OgreMath.cpp) which OGRE's ray-scene code uses under the hood, `Math::intersects(ray, box)` will return 0 when the origin of the ray is within the bounding box.  The interaction of these two systems leads to the described bug: if the object's AABB contains the origin of the ray, it is skipped for consideration within the CPU implementation of Ogre2RayQuery.

I think this check should just be `< 0` instead of `<= 0` on `ThreadedTriRay::execute`, but I'm not sure what considerations were at play when this code was written (and I have not been able to successfully build gazebo, so I only have access to the binary versions).

Best I can tell, this bug affects my version (7) all the way through the version on master.

## Steps to reproduce

Use the CPU implementation of RayQuery within OGRE2, command a query that _should_ intersect valid geometry, but starts within the AABB of that geometry, observe that it does not.

Dylan-Gallagher:
# 🦟 Bug fix

Fixes #1226.

## Summary

Ogre returns a broadphase distance of zero when a ray starts inside an object's axis-aligned bounding box. The Ogre2 CPU ray-query path discarded all candidates with a distance less than or equal to zero, so it never ran the triangle-level intersection pass for those objects.

This change keeps zero-distance candidates while continuing to reject negative distances. The narrowphase still determines whether the ray actually intersects the geometry.

The regression test uses a rotated box whose AABB contains the ray origin while the box geometry does not. Before the fix, the CPU ray query reports no result. After the fix, it returns the expected front-face intersection, object ID, distance, and point.

## Backport Policy

- [ ] This is safe to backport to the following versions:
    - [ ] Jetty
    - [ ] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [x] I am not sure

## Validation

- Final regression on Ogre and Ogre2 GL3Plus: 4/4 CTest entries passed.
- Full serial CTest suite under Xvfb: 257/257 passed.
- `codecheck`: cpplint and cppcheck targets passed.
- Negative control: changing the production condition back to `<= 0.0` makes `RayQueryTest.RayOriginInsideObjectBounds` fail because the result is empty.

## Checklist

- [ ] Added a screen capture or video to the PR description that demonstrates the fix (not needed for this CPU query regression)
- [x] Added tests
- [x] Updated documentation (not needed; no public API or documented behavior changed)
- [x] Updated migration guide (not needed; no migration impact)
- [x] Consider updating Python bindings (not needed; no API change)
- [x] Updated Bazel files (not needed; no files or targets were added)

Generated-by: OpenAI Codex (GPT-5; accessed 2026-08-18)

### DIFF
--- a/ogre2/src/Ogre2RayQuery.cc
+++ b/ogre2/src/Ogre2RayQuery.cc
@@ -350,7 +350,8 @@ void ThreadedTriRay::execute(size_t _threadId, size_t _numThreads)
   for (auto iter = this->ogreResult.begin(); iter != this->ogreResult.end();
        ++iter)
   {
-    if (iter->distance <= 0.0)
+    // A distance of zero is valid when the ray origin is inside the AABB.
+    if (iter->distance < 0.0)
       continue;
 
     if (!iter->movable || !iter->movable->getVisible())
--- a/test/common_test/RayQuery_TEST.cc
+++ b/test/common_test/RayQuery_TEST.cc
@@ -17,11 +17,14 @@
 
 #include <gtest/gtest.h>
 
+#include <cmath>
+
 #include "CommonRenderingTest.hh"
 
 #include "gz/rendering/Camera.hh"
 #include "gz/rendering/RayQuery.hh"
 #include "gz/rendering/Scene.hh"
+#include "gz/rendering/Visual.hh"
 
 using namespace gz;
 using namespace rendering;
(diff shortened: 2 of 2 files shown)


kvnloo:
I traced this through the Ogre broadphase behavior and the change looks well-scoped to me.

"Math::intersects(ray, box)" can legitimately produce a distance of "0" when the ray origin is inside the AABB, so filtering "<= 0" here prevents a valid broadphase candidate from ever reaching the triangle-level intersection. Changing the condition to "< 0" preserves rejection of negative distances while allowing that valid zero-distance case through.

I also like the regression geometry: the ray origin is inside the rotated box's AABB but outside the actual box, so the test still requires the narrowphase to find the real front-face intersection rather than merely accepting an AABB hit. The negative-control result in the PR description is a good check that the test is exercising the intended boundary condition.

I don't see a reason to broaden the fix beyond this.

Assisted-by: OpenAI ChatGPT GPT-5.6 Sol

PR #1325, opened 2026-08-31
Wayland integration

AbdulkadirEroglu:
Hello everybody,
First this is my second pr to a big project like this so if i made any mistakes please excuse me and tell me so that i can fix it in the feature. Let's get started.
Also i didn't see any enhancement request for this, i think most people will use the xwayland route. I just wanted it to work natively and seeing that i can help to the project that i used for years, i gave it a try. 
screenshot-2026-08-31_23-10-44

# 🎉 New feature

## Summary
Adds native Wayland EGL support to the `ogre2` rendering backend. Previously, running
under a Wayland session meant either falling back to XWayland or crashing outright when
`useCurrentGLContext` was requested (Gazebo's normal GUI rendering path always sets this).

`Ogre2RenderEngine::CreateRenderWindow()` now recognizes a `wayland`/`waylandDisplay`/
`waylandSurface` set of params and passes through the "Wayland EGL Window" Ogre
interface accordingly, plumbing through to OGRE-Next's Wayland EGL windowing backend.

This is one part of a three-repo change enabling native Wayland rendering in Gazebo
(gz-rendering + gz-gui + gz-sim), and depends on an OGRE-Next fix that isn't merged
upstream yet:
- OGRECave/ogre-next#593 (adopt Qt's current EGL context — required for this to work
  correctly, not just build)
- gazebosim/gz-gui#780
- gazebosim/gz-sim#3943

## Backport Policy
- [ ] I am not sure

## Test it
Requires the OGRE-Next PR above checked out locally (this repo alone will build and run
under XWayland/GLX exactly as before — no behavior change there).

1. Build OGRE-Next from `OGRE-Next/ogre-next#<TBD>` and install to a shared prefix.
2. Build this branch against that prefix.
3. `wayland_sanity_check.cpp` at the repo root reproduces the new path standalone
   (kept in this PR specifically as a manual-verification aid, since the Wayland path
   needs a live compositor that CI can't provide): it opens its own EGL context on a
   bare `wl_surface` (mimicking what Qt does), sets `useCurrentGLContext=1`, then
   creates an `ogre2` engine/scene/camera and runs 30 render iterations, confirming the
   caller's context survives engine teardown.
   g++ -std=gnu++17 $(pkg-config --cflags gz-rendering)
       $(pkg-config --cflags wayland-client wayland-egl egl)
       wayland_sanity_check.cpp -o wayland_sanity_check
       $(pkg-config --libs gz-rendering)
       $(pkg-config --libs wayland-client wayland-egl egl)
   ./wayland_sanity_check
4. End-to-end: verified via a real `gz sim` GUI session under
`GZ_GUI_WAYLAND=1`, rendering `examples/worlds/shapes.sdf` with correct pixel output
(screenshot attached).

## Checklist
- [x] Added a screen capture or video to the PR description that demonstrates the fix
- [ ] Added tests — no gtest coverage added; the Wayland path needs a live compositor
   CI doesn't have, so `wayland_sanity_check.cpp` above is kept as the manual
   verification path instead
- [x] `codecheck` passed (There are warnings but nothing related to the code i changed or this pr)
- [x] All tests passed (134/134 locally; existing suite has no Wayland-specific coverage)
- [ ] While waiting for a review on your PR, please help review another open pull request
- [x] Was GenAI used to generate this PR?

Assisted-by: Claude (Anthropic) — used for implementation and debugging assistance.


AbdulkadirEroglu:
Hi,
Here's a little (not so little) update.

The ogre-next maintainers want the Wayland changes to land on their master branch instead of v2.3. That means I'd need to update gz-rendering8's Ogre dependency to a newer target — ogre-next 4.0 (their master branch is versioned 4.0, not 3.0 as the README might suggest).

That leaves me with a few questions:

Which branch should I target? Continue on gz-rendering8, move to gz-rendering10 (the current newest series), or work directly against gz-rendering's own master/main?

What dependency strategy would you prefer? A clean cutover to the new ogre-next release, or a dual-build option supporting both v2.3 and the new version during a transition period?

For this PR — should I close it and open a new one against the target branch, or keep updating this one?

Thanks for all your work and support.

azeey:
@AbdulkadirEroglu, thanks for the contribution in this direction, but we currently don't plan to upgrade our version of Ogre-next in the short/medium term. If the wayland changes land in Ogre-next 4.0 and cannot be backported to the 2.3 line, I see that as a blocker. For rendering related work, we are looking into using an external renderer process as mentioned in https://gazebosim.org/docs/latest/roadmap/#primary-roadmap. Preliminary design is [here](https://github.com/gazebosim/design/pull/10) if you are interested.

PR #1330, opened 2026-09-02
Change Bazel to use Clang for layering checks, and drop Bazel 8.0.0 from CI

cheriehu:
# 🦟 Bug fix
Tracking issue: https://github.com/gazebo-tooling/gz-bazel-dev/issues/5

## Summary

Change bazel build to use clang

## Backport Policy

These versions have a MODULE.bazel file
- [X] This is safe to backport to the following versions:
    - [X] Jetty
    - [X] Ionic
    - [ ] Harmonic
    - [ ] Fortress
- [ ] I am not sure

## Checklist
- [X] Signed all commits for DCO
- [ ] Added tests
- [ ] Was GenAI used to generate this PR? If so, make sure to add "Assisted-by" to your commits. (See [this policy](https://osralliance.org/wp-content/uploads/2025/05/OSRF-Policy-on-the-Use-of-Generative-Tools-Generative-AI-in-Contributions.pdf) for more info.)

### DIFF
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -23,4 +23,4 @@ jobs:
           {"folder": ".", "bzlmodEnabled": false}
         ]
       exclude_windows: true
-      bazel_test_command: "bazel test --test_output=errors //..."
+      bazel_test_command: "bazel test --repo_env=CC=clang --repo_env=CXX=clang++ --test_output=errors //..."
--- a/.github/workflows/bazel.yml
+++ b/.github/workflows/bazel.yml
@@ -14,6 +14,8 @@ jobs:
   test:
     uses: bazel-contrib/.github/.github/workflows/bazel.yaml@v7.7.0
     with:
+      # Make sure to keep in sync with version in root's .bazelversion
+      bazel_versions: '["9.1.1"]'
       folders: |
         [
           "."


cheriehu:
The Bazel CI failure seems to be pre-existing and only for Bazel 8.0.0. Looks like it is because it is unable to download freetype.

Gemini suggested bumping the Bazel 8 version to 8.3.0 (if we still want to use Bazel 8). 8.3.0 supports the `mirror_urls` field so it can download freetype from the mirror URLs if the primary URL fails. This works for me on my machine but it means the bazel version for this repo will different from all the other gz repos (?)

luca-della-vedova (commented):
Imho I would skip Bazel 8.0.0, if I remember correctly from the workflow it is only ran because the bazel workflow by default runs 8.0.0, not really because we are explicitly targeting.
I would change the workflow so we only run it on the Bazel that we are targeting explicitly, I think I did it for one of the repos but I can't remember which (it's a flag in the workflow to disable the default version)

luca-della-vedova:
That being said if you address that make sure to update the PR title so the commit message is informative if it is merged

cheriehu:
> Imho I would skip Bazel 8.0.0, if I remember correctly from the workflow it is only ran because the bazel workflow by default runs 8.0.0, not really because we are explicitly targeting. I would change the workflow so we only run it on the Bazel that we are targeting explicitly, I think I did it for one of the repos but I can't remember which (it's a flag in the workflow to disable the default version)

Thanks! I copied the way you did it in gz-transport

iche033 (approved):

luca-della-vedova:
@mergifyio backport gz-rendering10 gz-rendering9

PR #1331, opened 2026-09-07
Add an image constructor that wraps a caller owned pixel buffer

caguero:
# 🎉 New feature

## Summary

Today an `Image` always allocates and owns its own pixel buffer. This patch adds one constructor that takes a buffer the caller owns and uses it as the image's storage. The image never frees it.

Here's the example that motivated this PR: a camera sensor in gz-sensors publishes every frame as a `msgs::Image`. The frame is read back from the GPU into the `Image` buffer, and then copied a second time into the gz-msg payload, because the message needs the bytes in memory it owns. With this constructor the sensor can wrap the message payload in an `Image`, so the GPU readback lands directly in the message and the second copy disappears. On a profile of the server with two 1920x1080 cameras that second copy costs about 425 microseconds per frame (6 MB per frame), and removing it made the server use 18% less CPU per step on that world. For small cameras (320x240) the saving is about 10 microseconds per frame, which does not show at world level. The full numbers are in the profiling report at https://caguero.github.io/gz-profiling/ (finding 9.6 of the 2026 09 01 run).

The gz-sensors side will come as a separate PR once this one is available.

Rules for callers, also documented on the constructor: the buffer must hold at least `MemorySize()` bytes and must outlive the image and any copy of it, since copies share the buffer.

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

Adding a constructor to a class that keeps its state behind a private implementation pointer does not change the class layout or any existing symbol, so it is ABI safe and can ship in a minor release. The `Image` class is identical on gz-rendering10, 9 and 8, so the patch applies as is there. On Fortress (ign-rendering6) `Image` predates the implementation pointer helper, so the backport needs a two line adaptation in the constructor body. Backporting matters for the gz-sensors follow up, which needs this constructor to remove the copy on the stable branches too.

## Checklist
- [x] Added tests
- [x] Updated Bazel files (if adding new files). Created an issue otherwise. (The Bazel build globs `src/*_TEST.cc`, no change needed.)

Assisted-by: Claude Fable 5.1

### DIFF
--- a/include/gz/rendering/Image.hh
+++ b/include/gz/rendering/Image.hh
@@ -45,6 +45,19 @@ namespace gz
       public: Image(unsigned int _width, unsigned int _height,
                   PixelFormat _format);
 
+      /// \brief Constructor. Creates an image that stores its pixels in a
+      /// caller owned buffer instead of allocating its own. The buffer must
+      /// hold at least MemorySize() bytes and outlive the image and any copy
+      /// of it, since copies share the same buffer. Copying a camera into
+      /// such an image writes straight into the caller's memory, for example
+      /// a message payload, and saves one copy per frame.
+      /// \param[in] _width Image width in pixels
+      /// \param[in] _height Image height in pixels
+      /// \param[in] _format Image pixel format
+      /// \param[in] _data Caller owned pixel buffer
+      public: Image(unsigned int _width, unsigned int _height,
+                  PixelFormat _format, void *_data);
+
       /// \brief Destructor
       public: virtual ~Image();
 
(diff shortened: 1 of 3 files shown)


ahcorde (approved):

caguero:
@Mergifyio backport gz-rendering10 gz-rendering9 gz-rendering8 ign-rendering6

azeey:
@caguero have you considered using a shared_ptr/weak_ptr for this? I would think that would be safer.

PR #1336, opened 2026-09-08
Image: take a shared pointer for the caller owned pixel buffer

caguero:
# 🎉 New feature

## Summary

Follow up to #1331, as suggested by @azeey in https://github.com/gazebosim/gz-rendering/pull/1331#issuecomment-5585974062: the constructor that wraps a caller owned pixel buffer now takes a shared pointer (`Image::DataPtr`) instead of a raw pointer. 

## Backport Policy
    - [x] Jetty
    - [x] Ionic
    - [x] Harmonic
    - [x] Fortress
- [ ] I am not sure

## Checklist
- [x] Added tests

Assisted-by: Claude Fable 5.1

### DIFF
--- a/src/Image.cc
+++ b/src/Image.cc
@@ -17,10 +17,11 @@
 
 #include <memory>
 
+#include <utility>
+
 #include "gz/rendering/Image.hh"
 
 /// \brief Shared pointer to raw image buffer
-typedef std::shared_ptr<unsigned char> DataPtr;
 
 /// \brief Private fields of Image
 class gz::rendering::Image::Implementation
@@ -35,7 +36,7 @@ class gz::rendering::Image::Implementation
   public: PixelFormat format = PF_UNKNOWN;
 
   /// \brief Pointer to the image data
-  public: DataPtr data = nullptr;
+  public: Image::DataPtr data = nullptr;
 };
 
 using namespace gz;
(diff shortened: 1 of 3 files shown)


src/Image.cc
```diff
@@ -67,20 +68,18 @@ Image::Image(unsigned int _width, unsigned int _height,
   this->dataPtr->format = PixelUtil::Sanitize(_format);
   unsigned int size = this->MemorySize();
   this->dataPtr->data =
-      DataPtr(new unsigned char[size], ArrayDeleter<unsigned char>());
+      Image::DataPtr(new unsigned char[size], ArrayDeleter<unsigned char>());
```
azeey, line 71: I think we can remove the deleter. C++17 supports `std::shared_ptr<T[]>` [1].


[1]: https://en.cppreference.com/cpp/memory/shared_ptr/shared_ptr

azeey (commented):
Looking at the usage at the callsite, I'm not sure if this is safer, if not, even more dangerous. If the most common use case is to pass along a buffer that is managed by a std::vector, maybe a raw pointer was a better choice. Passing a empty deleter or using aliasing shared_ptr seem very error prone to me. However, I'm still concerned that the change to an external owner can cause issues down the road. Since the class is small, maybe sticking with #1331 and adding extra documentation is sufficient. @iche033 any thoughts?

caguero:
Thanks @azeey, I agree, I think we should close this PR and stick with the raw pointer version.

For the use case that motivated this, the buffer belongs to a protobuf message, so there is no natural shared pointer to pass in. The caller would end up faking one with an empty deleter or an aliasing constructor, which is error prone, agreed. And a shared pointer only guards against the owner going away, not against the owner moving its memory, which is another risk in my opinion.

While looking at this I also found that the Bayer path in `Copy()` replaces the image it was handed with a new allocated Image, so a caller owned buffer would never receive the pixels. That fix, together with some extra documentation, makes more sense as a separate PR. I'll open it if you agree.

azeey:
Yes, let's close this. But the issue you mentioned with the Bayer path shows the kind of problems we will run into. It would be nice to prevent this kind of bug at compile time.

caguero:
> Yes, let's close this. But the issue you mentioned with the Bayer path shows the kind of problems we will run into. It would be nice to prevent this kind of bug at compile time.

I think there's a way to do it at compile time, let me move this discussion to a separate PR.

